PR #11, opened 2024-10-09
Add Dockerfile, Makefile and use `uv` instead of `poetry`

rnetser:
- Move to use [uv](https://docs.astral.sh/uv/) 
- Add Dockerfile and Makefile


adolfo-ab (approved):

Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+    && tar xvf /tmp/rosa-linux.tar.gz --no-same-owner \
+    && mv rosa /usr/bin/rosa \
+    && chmod +x /usr/bin/rosa \
+    && rosa version
+
+# Install the OpenShift CLI (OC)
+RUN curl -L https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/stable/openshift-client-linux.tar.gz --output /tmp/openshift-client-linux.tar.gz \
+    && tar xvf /tmp/openshift-client-linux.tar.gz --no-same-owner \
+    && mv oc /usr/bin/oc \
+    && mv kubectl /usr/bin/kubectl \
```
mwaykole, line 26: why kubectl ?

pyproject.toml
```diff
@@ -16,42 +16,34 @@ no_implicit_optional = true
...
+]
+
+[project]
+requires-python = ">=3.8"
 name = "opendatahub-tests"
 version = "0.1.0"
 description = "Tests repository for Open Data Hub (ODH)"
 authors = []
-license = "Apache License 2.0"
+license = "Apache-2.0"
```
mwaykole, line 31: are u sure aboout `License`

mwaykole (dismissed):
few comments else looks good

Makefile
```diff
@@ -0,0 +1,28 @@
+IMAGE_BUILD_CMD = $(shell which podman 2>/dev/null || which docker)
+IMAGE_REGISTRY ?= "quay.io"
+REGISTRY_NAMESPACE ?= "modh"
+IMAGE_NAME="opendatahub-tests"
+IMAGE_TAG ?= "latest"
+
+FULL_OPERATOR_IMAGE ?= "$(IMAGE_REGISTRY)/$(REGISTRY_NAMESPACE)/$(IMAGE_NAME):$(IMAGE_TAG)"
+POETRY_BIN = poetry
```
lugi0, line 8: I think this is not needed if we switch to uv?

pyproject.toml
lugi0: I vote for agpl 3

lugi0 (dismissed):
I have my concerns about switching to `uv` since it's a newcomer and written in Rust, but we can evaluate it and switch back to Poetry if needed

pyproject.toml
rnetser: why not apache? checking `kserve` and `opendatahub-operator` as examples of repos in the org - they use apache. we should use the same

Makefile
rnetser: leftover, removing

Dockerfile
rnetser: will remove

pyproject.toml
rnetser: > are u sure aboout `License`

updated the string

lugi0 (approved):

mwaykole (approved):

PR #14, opened 2024-10-10
Add log_listener and logging for pytest to support additional loggers

rnetser:
Required to allow multiple loggers (one with additional details, e.g date and a basic one).
This enables adding additional info to the logs - test name. setup and teardown sections


rnetser:
verified

conftest.py
```diff
@@ -1,14 +1,50 @@
...
 BASIC_LOGGER = logging.getLogger("basic")
-LOGGER = get_logger(__name__)
+
+
+def pytest_addoption(parser):
+    log_group = parser.getgroup(name="PytestLog")
+    log_group.addoption(
+        "--pytest-log-file",
+        help="Path to pytest log file",
+        default="pytest-tests.log",
```
mwaykole, line 18: may be we should give the default path to /tmp/odh-test.log

rnetser: i want to keep the same convention as pytest (default location in parent dir)

adolfo-ab (approved):

PR #21, opened 2024-10-23
[model server] Add authentication tests

rnetser:


tests/model_serving/model_server/authentication/conftest.py
```diff
@@ -0,0 +1,120 @@
...
+    with ServingRuntimeFromTemplate(
+        client=admin_client,
+        name=request.param["name"],
+        namespace=model_namespace.name,
+        template_name=request.param["template-name"],
+    ) as model_runtime:
+        yield model_runtime
+
+
+@pytest.fixture(scope="class")
```
mwaykole, line 86: I believe adding an inference_service fixture solely for the new storage isn't the best approach. As you've noted, there's repeated code for both the serving runtime and the inference service. We should aim to find a more optimal solution to address this.

tests/model_serving/model_server/authentication/test_kserve_token_authentication.py
```diff
@@ -0,0 +1,65 @@
...
+    # def test_block_cross_model_authentication(self):
+    #     pass
+    #
+    # def test_disabled_model_authentication(self):
+    #     pass
+    #
+    # def test_re_enabled_model_authentication(self):
+    #     pass
+    #
+    # def test_model_authentication_using_invalid_token(self):
```
mwaykole, line 64: for this Test we should try to create 2  model 
model 1 token1
model 2 token 2

try to access model 1 with token 2

tests/model_serving/model_server/authentication/conftest.py
rnetser: the  logic should be in `create_isvc`; it is ok to have more than one fixture calling the same function.
in this case, as different storages require diffrenet setup (.e pvc needs to have pvc with download data first) it makes sense to have different fixtures

tests/model_serving/model_server/authentication/test_kserve_token_authentication.py
rnetser: These are the basic tests; will add cross-auth once the above is working :)

tests/model_serving/model_server/authentication/conftest.py
```diff
@@ -0,0 +1,155 @@
...
+        namespace=model_namespace.name,
+        template_name=request.param["template-name"],
+        enable_grpc=True,
+        enable_http=True,
+    ) as model_runtime:
+        yield model_runtime
+
+
+@pytest.fixture(scope="class")
+def s3_inference_service(
```
mwaykole, line 92: move this to conftest under model_server

mwaykole:
where data connection is created ?

tests/model_serving/conftest.py
```diff
@@ -0,0 +1,32 @@
+from typing import Optional, Tuple
```
lugi0, line 1: You should also add fixtures for the buckets in this conftest

tests/model_serving/model_server/authentication/conftest.py
```diff
@@ -0,0 +1,302 @@
...
+
+
+@pytest.fixture(scope="class")
+def s3_models_storage_uri(request, models_s3_bucket_name) -> str:
+    return f"s3://{models_s3_bucket_name}/{request.param['model-dir']}/"
+
+
+@pytest.fixture(scope="session")
+def models_s3_endpoint(models_s3_bucket_region) -> str:
+    return f"https://{models_s3_bucket_region}.amazonaws.com/"
```
lugi0, line 75: These fixtures should be in a more general conftest file

tests/model_serving/model_server/authentication/conftest.py
```diff
@@ -0,0 +1,302 @@
...
+
+
+@pytest.fixture(scope="class")
+def endpoint_s3_secret(
```
lugi0, line 79: This is already defined here: https://github.com/opendatahub-io/opendatahub-tests/pull/22/files/38b1f728c8d82955805212d36cf8c47f03d17a10#diff-ba4079b730a33a115ca68082ee9cdafe3a1d4cbff3aba9d09d508deacafbb183R51

tests/model_serving/model_server/authentication/constants.py
```diff
@@ -0,0 +1,15 @@
+# Protocols
+HTTP_STR: str = "http"
+GRPC_STR: str = "grpc"
+
+# Deployment mode
+SERVERLESS_STR: str = "Serverless"
```
lugi0, line 6: Why would you need to define these as constants?

tests/model_serving/model_server/authentication/utils.py
```diff
@@ -0,0 +1,111 @@
...
+from ocp_resources.inference_service import InferenceService
+from ocp_resources.role import Role
+from simple_logger.logger import get_logger
+
+from utilities.inference_utils import Inference
+
+LOGGER = get_logger(name=__name__)
+
+
+def base64_encode_str(text: str) -> str:
```
lugi0, line 16: This is already defined in https://github.com/opendatahub-io/opendatahub-tests/pull/22/files/38b1f728c8d82955805212d36cf8c47f03d17a10#diff-5045407005153d8609d7df4be7a249a67b6eebd4c8b970597fcecedce95fd2c5R101

tests/model_serving/model_server/authentication/utils.py
```diff
@@ -0,0 +1,111 @@
...
+
+    elif output := re.findall(r"generated_text\": \"(.*)\"", res["output"], re.MULTILINE):
+        assert "".join(output) == expected_response_text
+
+    else:
+        LOGGER.error(f"Inference response text not found in response. Response: {res}")
+        raise
+
+
+def get_s3_secret_dict(
```
lugi0, line 73: Already defined in https://github.com/opendatahub-io/opendatahub-tests/pull/22/files/38b1f728c8d82955805212d36cf8c47f03d17a10#diff-ba4079b730a33a115ca68082ee9cdafe3a1d4cbff3aba9d09d508deacafbb183R51

tests/model_serving/model_server/authentication/utils.py
```diff
@@ -0,0 +1,111 @@
...
+
+        else:
+            LOGGER.error(f"Auth header {auth_header} not found in response. Response: {res['output']}")
+            raise
+
+    elif inference.inference_response_text_key_name:
+        assert res["output"][inference.inference_response_text_key_name] == expected_response_text
+
+    elif output := re.findall(r"generated_text\": \"(.*)\"", res["output"], re.MULTILINE):
+        assert "".join(output) == expected_response_text
```
lugi0, line 66: This is not a great way to define the validation of inference outputs, it is rare that you get the exact same response every single time

tests/model_serving/model_server/utils.py
```diff
@@ -0,0 +1,61 @@
...
+    external_route: bool = False,
+    model_service_account: Optional[str] = "",
+    min_replicas: Optional[int] = None,
+) -> InferenceService:
+    predictor_config: Dict[str, Any] = {
+        "model": {
+            "modelFormat": {"name": model_format},
+            "version": "1",
+            "runtime": runtime,
+            "storageUri": storage_uri,
```
lugi0, line 28: If you want to redefine this before #22 , please also make sure to include support for storage_key and storage_path, see: https://github.com/opendatahub-io/opendatahub-tests/pull/22/files/38b1f728c8d82955805212d36cf8c47f03d17a10#diff-afd99f559fc337b6071cc32a05efb2c1ad55e61e9d2bbc685545fc26a64cdd0fR17

utilities/inference_utils.py
```diff
@@ -0,0 +1,145 @@
...
+
+                else:
+                    raise ValueError(
+                        f"Protocol {self.protocol} not supported.\nSupported protocols are {inference_type}"
+                    )
+
+            else:
+                raise ValueError(
+                    f"Inference type {inference_type} not supported.\n"
+                    f"Supported inference types are {runtime_config['endpoints']}"
```
lugi0, line 62: What are these endpoints and why are you using them to specify what "inference types" are supported?

lugi0 (changes requested):

tests/model_serving/conftest.py
rnetser: code moved to tests.conftest

tests/model_serving/model_server/authentication/conftest.py
rnetser: once needed, we will move them

tests/model_serving/model_server/authentication/utils.py
rnetser: please check `INFERENCE_QUERIES` not sure how this can change?

tests/model_serving/model_server/utils.py
rnetser: addressed in #22

tests/model_serving/model_server/authentication/conftest.py
mwaykole: +1

tests/model_serving/model_server/authentication/utils.py
lugi0: Consider moving to `utilities/infra.py`

tests/model_serving/model_server/authentication/utils.py
lugi0: In ods-ci we have a few different approaches to handle potentially different responses from models. 
In the case of flan-t5-small I think we are ok with a `==` check, but there are other models where the response will be slightly different every time (still text based) and we use a similarity score between expected and received text. 
There are also models that do not respond with text but rather with images or arrays of floats (that can vary slightly due to imperfect rounding), and in both cases there we had to define helper functions to parse the results and handle validation. 
There is no one size fits all, but perhaps it could make sense to start setting up this function to call other helper functions mapped to the model being used, rather than having the check here directly.

conftest.py
```diff
@@ -33,7 +33,6 @@ def pytest_addoption(parser: Parser) -> None:
     buckets_group.addoption(
         "--ci-s3-bucket-name", default=os.environ.get("CI_S3_BUCKET_NAME"), help="Ci S3 bucket name"
     )
-
     buckets_group.addoption(
         "--ci-s3-bucket-region", default=os.environ.get("CI_S3_BUCKET_REGION"), help="Ci S3 bucket region"
```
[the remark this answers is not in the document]
rnetser: we have it https://github.com/opendatahub-io/opendatahub-tests/pull/21/files#diff-a31c7ed5d35f5ed8233994868c54d625b18e6bacb6794344c4531e62bd9dde59L51

conftest.py
```diff
@@ -43,14 +42,11 @@ def pytest_addoption(parser: Parser) -> None:
...
-        "--models-s3-bucket-name",
-        default=os.environ.get("MODELS_S3_BUCKET_NAME"),
-        help="Models S3 bucket name",
+        "--models-s3-bucket-name", default=os.environ.get("MODELS_S3_BUCKET_NAME"), help="Models S3 bucket name"
     )
-
     buckets_group.addoption(
         "--models-s3-bucket-region",
-        default=os.environ.get("MODELS_S3_BUCKET_REGION"),
+        default=os.environ.get("MODELS_S3_BUCKET_REGION", "us-east-1"),
```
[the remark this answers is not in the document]
rnetser: removing it

tests/model_serving/model_server/authentication/conftest.py
rnetser: because it requires the namespace

utilities/infra.py
```diff
@@ -22,3 +28,55 @@ def create_ns(
...
+        client=admin_client,
+        name=name,
+        namespace=namespace,
+        annotations={"opendatahub.io/connection-type": "s3"},
+        data_dict=get_s3_secret_dict(
+            aws_access_key=aws_access_key,
+            aws_secret_access_key=aws_secret_access_key,
+            aws_s3_bucket=aws_s3_bucket,
+            aws_s3_endpoint=aws_s3_endpoint,
+            aws_s3_region=aws_s3_region,
```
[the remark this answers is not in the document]
rnetser: no,  we should separate resource-related stuff and helper utils:
- file size
- easier to look for things - you know where each type of function is expected to be

tests/model_serving/model_server/authentication/utils.py
adolfo-ab: Do we even want to verify the actual content of the response? We are not testing the model itself, so if a given model answers nonsense, it's a model issue, not an ODH issue IMO. We could try to assert that the content of the answer makes sense, but I think that is a bit out of scope for us, and can potentially make the test flaky due to the reasons highlighted by @lugi0 . I think that in this case just verifying that the answer contains some non-empty string should be enough.

utilities/constants.py
```diff
@@ -10,8 +10,12 @@ class ModelFormat:
     CAIKIT: str = "caikit"
 
 
+class ModelName:
+    FLAN_T5_SMALL: str = f"flan-t5-small-{ModelFormat.CAIKIT}"
```
adolfo-ab, line 14: Maybe create a constant for `flan-t5-small` and reuse

utilities/inference_utils.py
```diff
@@ -0,0 +1,145 @@
...
+            inference_service: InferenceService object
+        """
+        self.inference_service = inference_service
+        self.runtime = runtime
+        self.protocol = protocol
+        self.inference_type = inference_type
+
+    @cache
+    def get_inference_url(self) -> str:
+        if url := self.inference_service.instance.status.components.predictor.url:
```
adolfo-ab, line 42: I think this is different if the inference service uses ModelMesh instead of KServe. @lugi0 can you confirm? If you are not gonna use ModelMesh for now, maybe we can address this on a different PR.

tests/model_serving/model_server/authentication/conftest.py
lugi0: Can it not be passed as an argument to the fixture?

rnetser: no

it is ok to have fixtures which do the same thing; this is why we have a function to do it in one place

the tests need to stay clean. separate resources for each test

tests/model_serving/model_server/authentication/utils.py
rnetser: we should separate util functions from resource-related functions. under "infra" we have resource stuff.

utilities/constants.py
rnetser: done

tests/model_serving/model_server/authentication/conftest.py
rnetser: deleted, we gat if from cli

tests/model_serving/model_server/authentication/utils.py
rnetser: moved to utlities/general.py

utilities/inference_utils.py
rnetser: added a `TODO` to support modelmesh

tests/model_serving/model_server/authentication/conftest.py
```diff
@@ -26,32 +26,27 @@
 
 @pytest.fixture(scope="session")
 def skip_if_no_authorino_operator(admin_client: DynamicClient):
+    name = "authorino"
```
lugi0, line 29: Wouldn't it be better to have this in a constants file? perhaps under something like `class RequiredOperators`

lugi0 (dismissed):
small comment but otherwise lgtm

tests/model_serving/model_server/authentication/conftest.py
rnetser: at the moment we only check for authorino in this module; it is not mandatory in other use cases so i think it's ok to leave it here for now. wdyt?

lugi0 (approved):

adolfo-ab (approved):

PR #22, opened 2024-10-23
Add kserve private endpoint tests

lugi0:
## Description

Adding two tests for kserve private endpoints. 

## How Has This Been Tested?

This has been tested multiple times from a local environment. It passes most of the time, but rarely fails on one of the assert calls in `test_curl_with_istio` because it receives an empty reply rather than an `OK` - might be a timing related problem, but seems weird because it is usually the second assert that fails.

If you want to run the tests yourself you will need to run `uv run pytest tests/model_serving/model_server/private_endpoint/`

## Merge criteria:

- [x ] The commits are squashed in a cohesive manner and have meaningful messages.
- [x ] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x ] The developer has manually tested the changes and verified that the changes work


tests/model_serving/model_server/private_endpoint/constants.py
```diff
@@ -0,0 +1,33 @@
+from typing import Dict, List, Any
+
+AWS_REGION: str = "us-east-2"
```
mwaykole, line 3: AWS_BUCKET_EAST_2

tests/model_serving/model_server/private_endpoint/constants.py
```diff
@@ -0,0 +1,33 @@
...
+AWS_BUCKET: str = "ods-ci-wisdom"
```
mwaykole, line 4: AWS_BUCKET_ODS_CI_WISDOM

tests/model_serving/model_server/private_endpoint/constants.py
```diff
@@ -0,0 +1,33 @@
...
+AWS_ENDPOINT: str = "https://s3.us-east-2.amazonaws.com/"
+
+SR_CONTAINERS_KSERVE_CAIKIT: List[Dict[Any, Any]] = [
+    {
+        "args": ["--model-name=/mnt/models/artifacts/"],
+        "command": ["text-generation-launcher"],
+        "env": [{"name": "HF_HOME", "value": "/tmp/hf_home"}],
+        "image": "quay.io/modh/text-generation-inference@sha256:294f07b2a94a223a18e559d497a79cac53bf7893f36cfc6c995475b6e431bcfe",
```
mwaykole, line 12: the image can be taken from process SR template

tests/model_serving/model_server/private_endpoint/constants.py
```diff
@@ -0,0 +1,33 @@
+from typing import Dict, List, Any
+
+AWS_REGION: str = "us-east-2"
+AWS_BUCKET: str = "ods-ci-wisdom"
+AWS_ENDPOINT: str = "https://s3.us-east-2.amazonaws.com/"
+
```
mwaykole, line 6: may be u can check  old code for pvc how its written

tests/model_serving/model_server/private_endpoint/test_kserve_private_endpoint.py
```diff
@@ -0,0 +1,75 @@
...
+        assert curl_stdout == "OK"
+
+    def test_curl_outside_istio(
+        self: Self,
+        endpoint_isvc: InferenceService,
+        endpoint_pod_without_istio_sidecar: Pod,
+        diff_pod_without_istio_sidecar: Pod,
+        service_mesh_member: ServiceMeshMember,
+    ) -> None:
+        LOGGER.info("Testing curl from the same namespace with a pod not part of the service mesh")
```
mwaykole, line 57: where are u comparing query output ?

uv.lock
```diff
@@ -695,11 +695,11 @@ dependencies = [
     { name = "semver" },
     { name = "timeout-sampler" },
 ]
-sdist = { url = "https://files.pythonhosted.org/packages/e4/73/e829a94696d4d2d3074affe97b2e3591a1776a9dd50a20dde10b93540819/openshift_python_utilities-5.0.68.tar.gz", hash = "sha256:afe16b18eec3b20ca71a41601904bb5d7416bf3ec27877faecc0f0c29b7c2f7c", size = 18225 }
+sdist = { url = "https://files.pythonhosted.org/packages/a4/72/fa3d90605f10621b2b3409c400ca7230c36ec4ca1b549f74829c196451e1/openshift_python_utilities-5.0.70.tar.gz", hash = "sha256:dcc8e31f263f04584cda6e327415e72ac00f5a563e7eecbd401b0cc9a6abc382", size = 18300 }
 
 [[package]]
 name = "openshift-python-wrapper"
-version = "10.0.93"
+version = "10.0.99"
```
mwaykole, line 702: 10.0.99 ---> change to 10.0.100

tests/model_serving/model_server/private_endpoint/test_kserve_private_endpoint.py
lugi0: in the assert

tests/model_serving/model_server/private_endpoint/constants.py
lugi0: Not sure I understand the suggestion

tests/model_serving/model_server/private_endpoint/constants.py
mwaykole: check the code for pvc test  for Serving runtime u can pull the same function here
we dont have to add the  runtime image

lugi0: If you mean https://github.com/opendatahub-io/opendatahub-tests/blob/main/tests/model_serving/model_server/storage/pvc/conftest.py#L132, I had looked at it but I'm not convinced it would work correctly. 
It is passing `kserve-ovms` as a template name (https://github.com/opendatahub-io/opendatahub-tests/blob/main/tests/model_serving/model_server/storage/constants.py#L10C23-L10C33) but I can't see where that template is actually defined, either in https://github.com/opendatahub-io/opendatahub-tests/blob/main/utilities/serving_runtime.py#L8 or in https://github.com/RedHatQE/openshift-python-wrapper/blob/main/ocp_resources/template.py#L1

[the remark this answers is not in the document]
tarukumar: This is what i was thinking.  I would suggest implementing an action that can download the `.env` file from the model-controller repository based on the branch, copy it to a specified folder in this repo, and make it available for testing using constants. Additionally, plan to add an option for users to provide an image URL as input for their tests. This would override the values in the constants file if not provided use the image present in the env file), allowing us to run tests on any image dynamically with input variables.

tarukumar: https://github.com/opendatahub-io/opendatahub-tests/blob/main/utilities/serving_runtime.py#L17

tarukumar: here is the template being fetch based on name

lugi0: Then I'm not sure I understand the logic here, is it only creating a `Template` instance with name `kserve-ovms` and namespace `redhat-ods-applications`? Where does it fetch the actual ServingRuntime details needed to instance the CR?

lugi0: Oh, that's cool! I didn't know about this.

README.md
```diff
@@ -1,2 +1,4 @@
 # opendatahub-tests
 # TODO: update file
+
+# Run `pre-commit` before committing. You will need npm installed to do so.
```
rnetser, line 4: Should be under `CONTRIBUTING.md` with:
## Code check
Install pre-commit for code check.
```bash
pre-commit install
```

mypy.ini
```diff
@@ -0,0 +1,8 @@
+[mypy]
```
rnetser, line 1: already exists under `pyproject.toml`

tests/model_serving/model_server/private_endpoint/conftest.py
```diff
@@ -0,0 +1,207 @@
...
+def diff_namespace(admin_client: DynamicClient) -> Generator[Namespace]:
+    yield from create_ns(admin_client=admin_client, name="diff-namespace")
+
+
+@pytest.fixture()
+def endpoint_sr(
+    admin_client: DynamicClient,
+    endpoint_namespace: Namespace,
+) -> Generator[ServingRuntime]:
+    with ServingRuntime(
```
rnetser, line 44: why not use `caikit-tgis-serving-template`  template?
see https://github.com/opendatahub-io/opendatahub-tests/pull/21

tests/model_serving/model_server/private_endpoint/conftest.py
```diff
@@ -0,0 +1,207 @@
...
+
+@pytest.fixture()
+def endpoint_isvc(
+    admin_client: DynamicClient,
+    endpoint_sr: ServingRuntime,
+    endpoint_s3_secret: Secret,
+    storage_config_secret: Secret,
+    endpoint_namespace: Namespace,
+) -> Generator[InferenceService]:
+    predictor = {
```
rnetser, line 86: please use `create_isvc`  and add what is missing there

tests/model_serving/model_server/private_endpoint/conftest.py
```diff
@@ -0,0 +1,207 @@
...
+        istio=True,
+        pod_name="test-with-istio",
+    )
+    yield pod
+    pod.clean_up()
+
+
+@pytest.fixture()
+def diff_pod_without_istio_sidecar(admin_client: DynamicClient, diff_namespace: Namespace) -> Generator[Pod]:
+    pod = create_sidecar_pod(
```
rnetser, line 189: please define `create_sidecar_pod` as contextmanager and have it yield the pod

tests/model_serving/model_server/private_endpoint/conftest.py
```diff
@@ -0,0 +1,207 @@
...
+        namespace=diff_namespace.name,
+        istio=False,
+        pod_name="test",
+    )
+    yield pod
+    pod.clean_up()
+
+
+@pytest.fixture()
+def running_flan_pod(admin_client: DynamicClient, endpoint_isvc: InferenceService) -> None:
```
rnetser, line 200: get the deployment and use `wait_for_replicas`

tests/model_serving/model_server/private_endpoint/constants.py
rnetser, line 4: please see https://github.com/opendatahub-io/opendatahub-tests/pull/21/files#diff-8df5ee07ff5e7eb9edcc5490df50302e439eb5f8e62fe669f159ce71b3407880R13-R15

this type of config should go there

tests/model_serving/model_server/private_endpoint/constants.py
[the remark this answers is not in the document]
rnetser: iiuc in this use case you can use the template.
we should have the info in the cluster and avoid setting this values in the tests

tests/model_serving/model_server/private_endpoint/infra.py
```diff
@@ -0,0 +1,40 @@
+from typing import Generator, Dict, Optional
+
+from kubernetes.dynamic import DynamicClient
+from ocp_resources.namespace import Namespace
+from ocp_resources.project_project_openshift_io import Project
+from ocp_resources.project_request import ProjectRequest
+
+TIMEOUT_2MIN = 2 * 10
+TIMEOUT_6MIN = 6 * 10
```
rnetser, line 9: used once, no need to save

tests/model_serving/model_server/private_endpoint/infra.py
```diff
@@ -0,0 +1,40 @@
...
+
+def create_ns(
+    name: str,
+    unprivileged_client: DynamicClient = None,
+    labels: Optional[Dict[str, str]] = None,
+    admin_client: DynamicClient = None,
+    teardown: bool = True,
+    delete_timeout: int = TIMEOUT_6MIN,
+) -> Generator[Namespace] | Generator[Project]:
+    if not unprivileged_client:
```
rnetser, line 20: we currently do not have `unprivileged_client`, please add the option in `tests/conftest.py`

tests/model_serving/model_server/private_endpoint/test_kserve_private_endpoint.py
```diff
@@ -0,0 +1,75 @@
...
+from ocp_resources.pod import Pod
+from ocp_resources.service_mesh_member import ServiceMeshMember
+from tests.model_serving.model_server.private_endpoint.utils import curl_from_pod
+
+
+LOGGER = get_logger(name=__name__)
+
+
+class TestKserveInternalEndpoint:
+    def test_deploy_model(
```
rnetser, line 15: i think it is better to split into 2 tests:
`test_deploy_model_state_loaded` and `test_deploy_model_url`

tests/model_serving/model_server/private_endpoint/test_kserve_private_endpoint.py
```diff
@@ -0,0 +1,75 @@
...
+        LOGGER.info("Testing curl from the same namespace with a pod part of the service mesh")
+
+        curl_stdout = curl_from_pod(
+            isvc=endpoint_isvc,
+            pod=endpoint_pod_with_istio_sidecar,
+            endpoint="health",
+        )
+        assert curl_stdout == "OK"
+
+        LOGGER.info("Testing curl from a different namespace with a pod part of the service mesh")
```
rnetser, line 40: should be in a separate test

tests/model_serving/model_server/private_endpoint/test_kserve_private_endpoint.py
```diff
@@ -0,0 +1,75 @@
...
+
+        curl_stdout = curl_from_pod(
+            isvc=endpoint_isvc,
+            pod=endpoint_pod_without_istio_sidecar,
+            endpoint="health",
+            protocol="https",
+        )
+        assert curl_stdout == "OK"
+
+        LOGGER.info("Testing curl from a different namespace with a pod not part of the service mesh")
```
rnetser, line 67: ditto

tests/model_serving/model_server/private_endpoint/utils.py
```diff
@@ -0,0 +1,77 @@
...
+class FlanPodNotFoundError(Exception):
+    pass
+
+
+class ProtocolNotSupported(Exception):
+    def __init__(self, protocol: str):
+        self.protocol = protocol
+        self.message = f"Protocol {protocol} is not supported"
+        super().__init__(self.message)
+
```
rnetser, line 23:
```suggestion
class ProtocolNotSupported(Exception):
    def __init__(self, protocol: str):
        self.protocol = protocol

    def __str__(self) -> str:
        return f"Protocol {protocol} is not supported"
```

tests/model_serving/model_server/private_endpoint/utils.py
```diff
@@ -0,0 +1,77 @@
...
+
+def get_flan_pod(client: DynamicClient, namespace: str, name_prefix: str) -> Pod:
+    for pod in Pod.get(dyn_client=client, namespace=namespace):
+        if name_prefix + "-predictor" in pod.name:
+            return pod
+
+    raise FlanPodNotFoundError(f"No flan predictor pod found in namespace {namespace}")
```
rnetser, line 30: please use k8s `ResourceNotFoundError` exception

tests/model_serving/model_server/private_endpoint/utils.py
```diff
@@ -0,0 +1,77 @@
...
+
+def curl_from_pod(
+    isvc: InferenceService,
+    pod: Pod,
+    endpoint: str,
+    protocol: str = "http",
+) -> str:
+    if protocol == "http":
+        tmp = isvc.instance.status.address.url
+        host = "http://" + tmp.split("://")[1]
```
rnetser, line 41: please use `urlparse`

tests/model_serving/model_server/private_endpoint/utils.py
```diff
@@ -0,0 +1,77 @@
...
+    cmd += " -- sleep infinity"
+
+    _, _, err = run_command(command=shlex.split(cmd), check=False)
+    if err:
+        # pytest.fail(f"Failed on {err}")
+        LOGGER.info(msg=err)
+
+    pod = Pod(name=pod_name, namespace=namespace, client=admin_client)
+    pod.wait_for_status(status="Running")
+    pod.wait_for_condition(condition="Ready", status="True")
```
rnetser, line 70: pod.wait_for_status(status="Running") should be enough

tests/model_serving/model_server/private_endpoint/utils.py
```diff
@@ -0,0 +1,77 @@
...
+    cmd = f"oc run {pod_name} -n {namespace} --image=registry.access.redhat.com/rhel7/rhel-tools"
+    if istio:
+        cmd = f'{cmd} --annotations=sidecar.istio.io/inject="true"'
+
+    cmd += " -- sleep infinity"
+
+    _, _, err = run_command(command=shlex.split(cmd), check=False)
+    if err:
+        # pytest.fail(f"Failed on {err}")
+        LOGGER.info(msg=err)
```
rnetser, line 66: any reason why not to create the pod using Pod()?

rnetser, line 66: if the command failed, there's no point in moving forward

tests/model_serving/model_server/private_endpoint/utils.py
```diff
@@ -0,0 +1,77 @@
...
+) -> Pod:
+    cmd = f"oc run {pod_name} -n {namespace} --image=registry.access.redhat.com/rhel7/rhel-tools"
+    if istio:
+        cmd = f'{cmd} --annotations=sidecar.istio.io/inject="true"'
+
+    cmd += " -- sleep infinity"
+
+    _, _, err = run_command(command=shlex.split(cmd), check=False)
+    if err:
+        # pytest.fail(f"Failed on {err}")
```
rnetser, line 65: leftover?

`pytest.fail` should be done in the test, if needed, not in util functions.
the function should raise, if needed.

rnetser (changes requested):

README.md
rnetser: https://github.com/opendatahub-io/opendatahub-tests/pull/25

mypy.ini
lugi0: It won't work with VSCode extension without the ini files

tests/model_serving/model_server/private_endpoint/utils.py
lugi0: OC returns RC 1 also for the warning about the security context of the pod, blocking the execution even though the pod exists

tests/model_serving/model_server/private_endpoint/conftest.py
lugi0: I was not aware of the possibility, see https://github.com/opendatahub-io/opendatahub-tests/pull/22#discussion_r1815032259

tests/model_serving/model_server/private_endpoint/constants.py
lugi0: I'm not sure I agree on this point, what if I want to define different variables for different test contexts without having to pass them in or define them as environment variables all the time? I think having some of these set as constants in the code works fine

tests/model_serving/model_server/private_endpoint/conftest.py
lugi0: why? I have a single pod and need to make sure that it's running

tests/model_serving/model_server/private_endpoint/utils.py
rnetser: ?

tests/model_serving/model_server/private_endpoint/conftest.py
rnetser: not addressed

tests/model_serving/model_server/private_endpoint/conftest.py
```diff
@@ -0,0 +1,195 @@
...
+    predictor_pod = get_flan_pod(
+        namespace=endpoint_isvc.namespace,
+        client=admin_client,
+        name_prefix=endpoint_isvc.name,
+    )
+    predictor_pod.wait_for_status(status="Running")
+    predictor_pod.wait_for_condition(condition="Ready", status="True")
```
rnetser, line 195: why wait for both?

tests/model_serving/model_server/private_endpoint/constants.py
```diff
@@ -0,0 +1,2 @@
+AWS_REGION_EAST_2: str = "us-east-2"
+AWS_ENDPOINT_EAST_2: str = "https://s3.us-east-2.amazonaws.com/"
```
rnetser, line 2:
```suggestion
AWS_ENDPOINT_EAST_2: str = f"https://s3.{AWS_REGION_EAST_2}.amazonaws.com/"
```

tests/model_serving/model_server/private_endpoint/infra.py
```diff
@@ -0,0 +1,22 @@
+from typing import Generator, Dict, Optional
+
+from kubernetes.dynamic import DynamicClient
+from ocp_resources.namespace import Namespace
+
+
+def create_ns(
+    name: str,
+    labels: Optional[Dict[str, str]] = None,
```
rnetser, line 9: 1. please place optional last
2. why is client default set to `None`?

tests/model_serving/model_server/private_endpoint/utils.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def __str__(self) -> str:
+        msg = f"""
+            You've passed the following parameters:
+            "storageUri": {self.storageUri}
+            "storage_key": {self.storage_key}
+            "storage_path: {self.storage_path}
+            In order to create a valid ISVC you need to specify either a storageUri value
+            or both a storage key and a storage path.
+            or both a storage key and a storage path.
```
rnetser, line 44: duplicate line

tests/model_serving/model_server/private_endpoint/utils.py
```diff
@@ -0,0 +1,104 @@
...
+        """
+        return msg
+
+
+def get_flan_pod(client: DynamicClient, namespace: str, name_prefix: str) -> Pod:
+    for pod in Pod.get(dyn_client=client, namespace=namespace):
+        if name_prefix + "-predictor" in pod.name:
```
rnetser, line 51: please get by `serving.kserve.io/inferenceservice` label

tests/model_serving/model_server/private_endpoint/utils.py
```diff
@@ -0,0 +1,104 @@
...
+    protocol: str = "http",
+) -> str:
+    if protocol == "http":
+        parsed = urlparse(isvc.instance.status.address.url)
+        host = parsed._replace(scheme="http").geturl()
+
+    elif protocol == "https":
+        host = isvc.instance.status.address.url
+    else:
+        raise ProtocolNotSupported(protocol)
```
rnetser, line 70:
```suggestion
    if protocol not in ("http"...):
        raise ...
    url = isvc.instance.status.address.url
    if protocol == "http":
        host = utl._replace(scheme="http").geturl()

```

tests/model_serving/model_server/private_endpoint/utils.py
```diff
@@ -0,0 +1,104 @@
...
+
+    _, _, err = run_command(command=shlex.split(cmd), check=False)
+    if err:
+        LOGGER.info(msg=err)
+
+    containers = [pod_name]
+    if istio:
+        containers.append("istio-proxy")
+
+    pod = Pod(name=pod_name, namespace=namespace, client=admin_client)
```
rnetser, line 95: why not as contextmanager?

tests/model_serving/model_server/private_endpoint/utils.py
```diff
@@ -0,0 +1,104 @@
...
+    pod.wait_for_status(status="Running")
+    pod.wait_for_condition(condition="Ready", status="True")
+    return pod
+
+
+def b64_encoded_string(string_to_encode: str) -> str:
+    # encodes the string to bytes-like, encodes the bytes-like to base 64, decodes the b64 to a string and returns it
```
rnetser, line 102: please use proper docstring format, ie.:
1st line  - a oneliner describing the function.
(if needed, newline and the more info)
then args, return etc

Please see https://google.github.io/styleguide/pyguide.html#383-functions-and-methods

tests/model_serving/model_server/storage/pvc/utils.py
```diff
@@ -10,12 +12,29 @@ def create_isvc(
...
+    predictor_storage = {
+        "minReplicas": min_replicas,
+        "model": {
+            "modelFormat": {"name": model_format},
+            "version": "1",
+            "runtime": runtime,
+        },
+    }
+    if storage_uri:
+        predictor_storage["model"]["storageUri"] = storage_uri  # type: ignore
```
rnetser, line 32: why is the `type:ignore` needed?
will it be solved if typing is added to `predictor_storage`?

tests/model_serving/model_server/storage/pvc/utils.py
```diff
@@ -10,12 +12,29 @@ def create_isvc(
...
+) -> Generator[InferenceService, Any, Any]:
+    predictor_storage = {
+        "minReplicas": min_replicas,
+        "model": {
+            "modelFormat": {"name": model_format},
+            "version": "1",
+            "runtime": runtime,
+        },
+    }
+    if storage_uri:
```
rnetser, line 31: what if all 3 are passed?

tests/model_serving/model_server/storage/pvc/utils.py
```diff
@@ -10,12 +12,29 @@ def create_isvc(
...
     model_format: str,
     runtime: str,
+    storage_uri: Optional[str] = None,
+    storage_key: Optional[str] = None,
+    storage_path: Optional[str] = None,
     min_replicas: int = 1,
     wait: bool = True,
-) -> InferenceService:
+) -> Generator[InferenceService, Any, Any]:
+    predictor_storage = {
```
rnetser, line 23: i think the name is misleading as the dict does not contain only storage info

rnetser (changes requested):

tests/model_serving/model_server/private_endpoint/conftest.py
```diff
@@ -0,0 +1,195 @@
...
+def endpoint_s3_secret(
+    admin_client: DynamicClient,
+    endpoint_namespace: Namespace,
+    aws_access_key_id: str,
+    aws_secret_access_key: str,
+    s3_bucket_name_wisdom: str,
+) -> Generator[Secret, None, None]:
+    data = {
+        "AWS_ACCESS_KEY_ID": b64_encoded_string(aws_access_key_id),
+        "AWS_DEFAULT_REGION": b64_encoded_string(AWS_REGION_EAST_2),
```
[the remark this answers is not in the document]
lugi0: This is the conftest for private_endpoint, so I think it makes sense to not handle every possible case. If we were to generalize this for any test then I'd agree with you that it would need to handle other regions

tests/model_serving/model_server/private_endpoint/conftest.py
```diff
@@ -0,0 +1,195 @@
...
+    endpoint_sr: ServingRuntime,
+    endpoint_s3_secret: Secret,
+    storage_config_secret: Secret,
+    endpoint_namespace: Namespace,
+) -> Generator[InferenceService, None, None]:
+    with create_isvc(
+        client=admin_client,
+        name="test",
+        namespace=endpoint_namespace.name,
+        deployment_mode="Serverless",
```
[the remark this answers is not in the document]
lugi0: Not sure I understand?

tests/model_serving/model_server/private_endpoint/utils.py
lugi0: I have tried implementing it as contextmanager but it then created issues when trying to execute the command in the pod; IIRC the issue was that I couldn't figure out why it wouldn't return a Pod object but rather a GeneratorSomething

tests/model_serving/model_server/storage/pvc/utils.py
lugi0: Without the ignore I'd get a complaint from mypy, `Unsupported target for indexed assignment ("object")`

tests/model_serving/model_server/storage/pvc/utils.py
lugi0: Mh, given the way it's written it would use storage_uri, but it should probably fail since I don't believe both formats are supported at the same time

tests/model_serving/model_server/private_endpoint/utils.py
lugi0: Running is different from being Ready, IIUC it's the same difference between a liveness and readiness probe. In our case we need the pod to be Ready in order to serve requests

tests/model_serving/model_server/private_endpoint/conftest.py
lugi0: See https://github.com/opendatahub-io/opendatahub-tests/pull/22/files/38b1f728c8d82955805212d36cf8c47f03d17a10#r1846653516

tests/model_serving/model_server/private_endpoint/conftest.py
lugi0: see https://github.com/opendatahub-io/opendatahub-tests/pull/22/files/38b1f728c8d82955805212d36cf8c47f03d17a10#r1846627863

conftest.py
```diff
@@ -31,6 +33,12 @@ def pytest_addoption(parser):
         "--ci-s3-bucket-name", default=os.environ.get("CI_S3_BUCKET_NAME"), help="Ci S3 bucket name"
     )
 
+    buckets_group.addoption(
```
rnetser, line 36: what about region?

conftest.py
```diff
@@ -83,3 +91,15 @@ def pytest_sessionfinish(session, exitstatus):
 
     reporter = session.config.pluginmanager.get_plugin("terminalreporter")
     reporter.summary_stats()
+
+
+@pytest.fixture(scope="session")
+def s3_bucket_name_wisdom(pytestconfig: pytest.Config) -> Optional[str]:
```
rnetser, line 97:
```suggestion
def s3_bucket_name_wisdom(pytestconfig: pytest.Config) -> str:
```
will always return because if not we raise

please rename to something like `models_s3_bucket_name` (and not include `wisdom`)

conftest.py
```diff
@@ -83,3 +91,15 @@ def pytest_sessionfinish(session, exitstatus):
...
+    wisdom_bucket = pytestconfig.option.models_s3_bucket_name
```
rnetser, line 98: ditto

mypy.ini
rnetser: did you check the vscode suggestion above?

tests/model_serving/model_server/private_endpoint/conftest.py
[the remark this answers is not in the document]
rnetser: we need to get the region from the user as the bucket has a region

tests/model_serving/model_server/private_endpoint/conftest.py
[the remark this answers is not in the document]
rnetser: we have "Serverless" in a number of places. so we need to start saving what is used more than once. this way, if needed, we change in one place

tests/model_serving/model_server/private_endpoint/conftest.py
```diff
@@ -0,0 +1,195 @@
...
+        name="storage-config",
+    ) as storage_config:
+        yield storage_config
+
+
+@pytest.fixture()
+def service_mesh_member(
+    admin_client: DynamicClient, diff_namespace: Namespace
+) -> Generator[ServiceMeshMember, None, None]:
+    with ServiceMeshMember(
```
rnetser, line 129: for serverless we do not need to create smm, it will be created by kserve

tests/model_serving/model_server/private_endpoint/conftest.py
rnetser: not sure i understand the ref to the other comment. let's talk about this

tests/model_serving/model_server/private_endpoint/utils.py
rnetser: please address

tests/model_serving/model_server/private_endpoint/utils.py
rnetser: let's go over this together

tests/model_serving/model_server/private_endpoint/utils.py
```diff
@@ -0,0 +1,112 @@
...
+    decodes the b64 to a string and returns it. This is needed for openshift
+    resources expecting b64 encoded values in the yaml.
+
+    Args:
+        string_to_encode: The string to encode in base64
+
+    Returns:
+        A base64 encoded string that is compliant with openshift's yaml format
+    """
+    # encodes the string to bytes-like, encodes the bytes-like to base 64, decodes the b64 to a string and returns it
```
rnetser, line 110: comment should be removed (is redundant with docscting)

tests/model_serving/model_server/storage/pvc/utils.py
```diff
@@ -10,12 +12,32 @@ def create_isvc(
...
+) -> Generator[InferenceService, Any, Any]:
+    predictor_dict = {
+        "minReplicas": min_replicas,
+        "model": {
+            "modelFormat": {"name": model_format},
+            "version": "1",
+            "runtime": runtime,
+        },
+    }
+    # "type:ignore" is needed because otherwise mypy will complain about "Unsupported target for indexed assignment ("object")"
```
rnetser, line 31: please add typing to `predictor_dict` and check again

tests/model_serving/model_server/storage/pvc/utils.py
```diff
@@ -10,12 +12,32 @@ def create_isvc(
...
+    if storage_uri and storage_key and storage_path:
+        raise InvalidStorageArgument(storage_uri, storage_key, storage_path)
+    elif storage_uri:
+        predictor_dict["model"]["storageUri"] = storage_uri  # type: ignore
```
rnetser, line 35: once `predictor_dict: Dict[str, Any]` is set I believe `# type: ignore` could be removed

tests/model_serving/model_server/storage/pvc/utils.py
```diff
@@ -10,12 +12,32 @@ def create_isvc(
...
+    elif storage_key and storage_path:
+        predictor_dict["model"]["storage"] = {"key": storage_key, "path": storage_path}  # type: ignore
+    else:
```
rnetser, line 38: how about creating a helper function to check the combinations so all logic is next to each other (i.e `if storage_uri and storage_key and storage_path` and making sure that storage_key and storage_path must be provided together?

rnetser (changes requested):

tests/model_serving/model_server/private_endpoint/conftest.py
```diff
@@ -0,0 +1,189 @@
...
+from kubernetes.dynamic import DynamicClient
+
+from utilities.serving_runtime import ServingRuntimeFromTemplate
+from tests.model_serving.model_server.storage.pvc.utils import create_isvc
+from tests.model_serving.model_server.private_endpoint.utils import (
+    create_sidecar_pod,
+    get_flan_deployment,
+    b64_encoded_string,
+)
+from tests.model_serving.model_server.private_endpoint.infra import create_ns
```
rnetser, line 19: we have `from tests.utils import create_ns` 

please move `create_ns` to utilities.infra.py and use in both places

tests/model_serving/model_server/private_endpoint/conftest.py
```diff
@@ -0,0 +1,189 @@
...
+from ocp_resources.inference_service import InferenceService
+from ocp_resources.secret import Secret
+from ocp_resources.namespace import Namespace
+from ocp_resources.pod import Pod
+from simple_logger.logger import get_logger
+from ocp_resources.serving_runtime import ServingRuntime
+from kubernetes.dynamic import DynamicClient
+
+from utilities.serving_runtime import ServingRuntimeFromTemplate
+from tests.model_serving.model_server.storage.pvc.utils import create_isvc
```
rnetser, line 13: please move to upper level, in this case should be under `tests.model_serving.model_server.utils`

tests/model_serving/model_server/private_endpoint/utils.py
```diff
@@ -0,0 +1,121 @@
...
+            "storageUri": {self.storageUri}
+            "storage_key": {self.storage_key}
+            "storage_path: {self.storage_path}
+            In order to create a valid ISVC you need to specify either a storageUri value
+            or both a storage key and a storage path.
+        """
+        return msg
+
+
+def get_flan_deployment(client: DynamicClient, namespace: str, name_prefix: str) -> Deployment:
```
rnetser, line 50: the function does not return anything at the moment, so update to `return deployment`

rnetser (changes requested):

rnetser (dismissed):

tests/conftest.py
```diff
@@ -12,6 +15,99 @@ def admin_client() -> DynamicClient:
...
+        )
+    return models_bucket
+
+
+@pytest.fixture(scope="session")
+def models_s3_bucket_region(pytestconfig: pytest.Config) -> str:
+    models_bucket_region = pytestconfig.option.models_s3_bucket_region
+    if not models_bucket_region:
+        raise ValueError(
+            "Bucket name for the models bucket is not defined."
```
adolfo-ab, line 99: Wrong error message (bucket name)

tests/conftest.py
```diff
@@ -12,6 +15,99 @@ def admin_client() -> DynamicClient:
...
+        )
+    return models_bucket_region
+
+
+@pytest.fixture(scope="session")
+def models_s3_bucket_endpoint(pytestconfig: pytest.Config) -> str:
+    models_bucket_endpoint = pytestconfig.option.models_s3_bucket_endpoint
+    if not models_bucket_endpoint:
+        raise ValueError(
+            "Bucket name for the models bucket is not defined."
```
adolfo-ab, line 110: Wrong error message (bucket name)

tests/conftest.py
```diff
@@ -12,6 +15,99 @@ def admin_client() -> DynamicClient:
...
+
+@pytest.fixture(scope="session")
+def models_s3_bucket_name(pytestconfig: pytest.Config) -> str:
+    models_bucket = pytestconfig.option.models_s3_bucket_name
+    if not models_bucket:
+        raise ValueError(
+            "Bucket name for the models bucket is not defined."
+            "Either pass with `--models-s3-bucket-name` or set `MODELS_S3_BUCKET_NAME` environment variable"
+        )
+    return models_bucket
```
adolfo-ab, line 91: I'm not sure if there's a better way to do it, but all these fixtures share the same logic. Is there a way to avoid duplication here?

tests/conftest.py
```diff
@@ -12,6 +15,99 @@ def admin_client() -> DynamicClient:
...
+        )
+    return bucket_name
+
+
+@pytest.fixture(scope="session")
+def ci_s3_bucket_region(pytestconfig: pytest.Config) -> str:
+    ci_bucket_region = pytestconfig.option.ci_s3_bucket_region
+    if not ci_bucket_region:
+        raise ValueError(
+            "Bucket name for the models bucket is not defined."
```
adolfo-ab, line 66: Wrong error message (bucket name)

tests/conftest.py
```diff
@@ -12,6 +15,99 @@ def admin_client() -> DynamicClient:
...
+        )
+    return ci_bucket_region
+
+
+@pytest.fixture(scope="session")
+def ci_s3_bucket_endpoint(pytestconfig: pytest.Config) -> str:
+    ci_bucket_endpoint = pytestconfig.option.ci_s3_bucket_endpoint
+    if not ci_bucket_endpoint:
+        raise ValueError(
+            "Bucket name for the models bucket is not defined."
```
adolfo-ab, line 77: Wrong error message (bucket name)

tests/model_serving/constants.py
```diff
@@ -0,0 +1,2 @@
+KSERVE_SERVERLESS: str = "Serverless"
+KSERVE_RAWDEPLOYMENT: str = "RawDeployment"
```
adolfo-ab, line 2: Just a nitpick, but maybe it would be a bit cleaner to have an enum, e.g.:

```
class KServeDeploymentType(Enum):
    SERVERLESS = "Serverless"
    RAW_DEPLOYMENT = "RawDeployment"
```

tests/model_serving/model_server/private_endpoint/conftest.py
```diff
@@ -0,0 +1,191 @@
...
+    endpoint_namespace: Namespace,
+) -> Generator[InferenceService, None, None]:
+    with create_isvc(
+        client=admin_client,
+        name="test",
+        namespace=endpoint_namespace.name,
+        deployment_mode=KSERVE_SERVERLESS,
+        storage_key="endpoint-s3-secret",
+        storage_path="flan-t5-small/flan-t5-small-caikit",
+        model_format="caikit",
```
adolfo-ab, line 94: Maybe create constants for `flan-t5` and `caikit`, it's very likely that we'll end up using them a lot throughout the repo

tests/model_serving/model_server/private_endpoint/conftest.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+@pytest.fixture(scope="session")
+def endpoint_namespace(admin_client: DynamicClient) -> Generator[Namespace, None, None]:
+    with create_ns(admin_client=admin_client, name="endpoint-namespace") as ns:
+        yield ns
+
+
+@pytest.fixture(scope="session")
+def diff_namespace(admin_client: DynamicClient) -> Generator[Namespace, None, None]:
```
adolfo-ab, line 33: Another nitpick, but I don't like these `diff_*` fixture names; they are not very descriptive and what if you need two namespaces in another test again?

tests/model_serving/model_server/private_endpoint/test_kserve_private_endpoint.py
```diff
@@ -0,0 +1,85 @@
...
+from ocp_resources.inference_service import InferenceService
+from ocp_resources.pod import Pod
+from ocp_resources.deployment import Deployment
+from tests.model_serving.model_server.private_endpoint.utils import curl_from_pod
+
+
+LOGGER = get_logger(name=__name__)
+
+
+class TestKserveInternalEndpoint:
```
adolfo-ab, line 14: Can you add a comment at the start of the class explaining the goal of these tests, so that someone not familiar with the feature can get a basic idea of what you are testing here?

tests/model_serving/model_server/private_endpoint/test_kserve_private_endpoint.py
```diff
@@ -0,0 +1,85 @@
...
+        endpoint_pod_with_istio_sidecar: Pod,
+    ) -> None:
+        LOGGER.info("Testing curl from the same namespace with a pod part of the service mesh")
+
+        curl_stdout = curl_from_pod(
+            isvc=endpoint_isvc,
+            pod=endpoint_pod_with_istio_sidecar,
+            endpoint="health",
+        )
+        assert curl_stdout == "OK"
```
adolfo-ab, line 40: Create constant and reuse

tests/model_serving/model_server/private_endpoint/test_kserve_private_endpoint.py
```diff
@@ -0,0 +1,85 @@
...
+        self: Self,
+        endpoint_isvc: InferenceService,
+        endpoint_pod_with_istio_sidecar: Pod,
+    ) -> None:
+        LOGGER.info("Testing curl from the same namespace with a pod part of the service mesh")
+
+        curl_stdout = curl_from_pod(
+            isvc=endpoint_isvc,
+            pod=endpoint_pod_with_istio_sidecar,
+            endpoint="health",
```
adolfo-ab, line 38: Create constant and reuse (maybe even create enum class)

tests/model_serving/model_server/private_endpoint/utils.py
adolfo-ab, line 50: There is nothing specific to the flan model in this fixture, right? Would it make sense to call it `get_kserve_predictor_deployment` or similar?

tests/model_serving/model_server/private_endpoint/utils.py
```diff
@@ -0,0 +1,121 @@
...
+        parsed = urlparse(host)
+        host = parsed._replace(scheme="http").geturl()
+    return pod.execute(command=shlex.split(f"curl -k {host}/{endpoint}"), ignore_rc=True)
+
+
+@contextmanager
+def create_sidecar_pod(
+    admin_client: DynamicClient,
+    namespace: str,
+    istio: bool,
```
adolfo-ab, line 89: Nitpick, maybe `use_istio` or similar would be easier to understand

tests/conftest.py
lugi0: Let's open an issue and think about this

tests/model_serving/model_server/private_endpoint/conftest.py
lugi0: Let's open an issue and think about this

tests/conftest.py
adolfo-ab: Works for me

tests/model_serving/model_server/private_endpoint/conftest.py
adolfo-ab: :+1:

adolfo-ab (dismissed):

tests/constants.py
```diff
@@ -0,0 +1,22 @@
+from enum import Enum
```
rnetser, line 1: can you please place under `tests\utilites`?

tests/constants.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class KServeDeploymentType(Enum):
+    SERVERLESS: str = "Serverless"
+    RAW_DEPLOYMENT: str = "RawDeployment"
+
+
+class ModelStoragePath(Enum):
+    FLAN_T5_SMALL: str = "flan-t5-small/flan-t5-small-caikit"
```
rnetser, line 10:
```suggestion
    FLAN_T5_SMALL: str = f"flan-t5-small/flan-t5-small-{ModelFormat.CAIKIT}"
```
and move ModelFormatup

rnetser (changes requested):

tests/conftest.py
lugi0: Opened https://github.com/opendatahub-io/opendatahub-tests/issues/44

tests/model_serving/model_server/private_endpoint/conftest.py
lugi0: Opened https://github.com/opendatahub-io/opendatahub-tests/issues/45

adolfo-ab (dismissed):

rnetser (approved):

adolfo-ab (approved):

PR #23, opened 2024-10-25
Add support for plugins to query the model as of now only for tgis-grpc and openai endpoints

tarukumar:
This PR add support for the tgis_grpc_plugin and openai . In the future we will add suppourt for other plugin like caiki,culr etc


pyproject.toml
```diff
@@ -45,6 +45,11 @@ dependencies = [
     "pytest-testconfig",
     "python-simple-logger",
     "pyyaml",
+    "syrupy>=4.0.1",
+    "caikit-nlp-client>=0.0.8",
+    "requests>=2.31.0",
+    "aiohttp>=3.8.5",
+    "pytest-asyncio>=0.21.1",
```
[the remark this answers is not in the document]
rnetser: `grpcio` is missing

plugins/tgis_grpc_plugin.py
```diff
@@ -0,0 +1,124 @@
...
+        if self.use_tls:
+            cert = self._get_server_certificate(self.host, 443).encode()
+            return grpc.ssl_channel_credentials(root_certificates=cert)
+        return None
+
+    def _create_channel(self) -> grpc.Channel:
+        credentials = self._channel_credentials()
+        return grpc.secure_channel(self.host, credentials) if credentials else grpc.insecure_channel(self.host)
+
+    def make_grpc_request(self, query: dict):
```
rnetser, line 49: please run with `pre-commit`. return tye is missing

plugins/tgis_grpc_plugin.py
```diff
@@ -0,0 +1,124 @@
...
+            response = response.responses[0]
+            return {
+                "input_tokens": response.input_token_count,
+                "stop_reason": response.stop_reason,
+                "output_text": response.text,
+                "output_tokens": response.generated_token_count,
+            }
+        except grpc.RpcError as err:
+            logger.error("gRPC Error: %s", err.details())
+            return None
```
rnetser, line 73: Why catch the exception and return `None`? 
This behavior should be controlled by the one calling the function

plugins/tgis_grpc_plugin.py
```diff
@@ -0,0 +1,124 @@
...
+            params=generation_pb2_grpc.generation__pb2.Parameters(
+                method=generation_pb2_grpc.generation__pb2.GREEDY,
+                sampling=generation_pb2_grpc.generation__pb2.SamplingParameters(seed=1037),
+                response=generation_pb2_grpc.generation__pb2.ResponseOptions(generated_tokens=True),
+            ),
+        )
+
+        try:
+            resp_stream = stub.GenerateStream(request=request)
+            for resp in resp_stream:
```
rnetser, line 92:
```suggestion
            for resp in stub.GenerateStream(request=request):
```
used once, can be called directly

plugins/tgis_grpc_plugin.py
```diff
@@ -0,0 +1,124 @@
...
+        except grpc.RpcError as err:
+            logger.error("gRPC Error: %s", err.details())
+            return None
+
+        return {
+            "input_tokens": query.get("input_tokens"),
+            "stop_reason": None,
+            "output_text": "".join(tokens),
+            "output_tokens": len(tokens),
+        }
```
rnetser, line 111: why is the return outside the `try`?

rnetser (changes requested):

pyproject.toml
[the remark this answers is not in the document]
tarukumar: grpcio will be installed as part of caikit-nlp-library no need to mention that. Also i normally add the version because i don't want pip  backtracking. but i can remove those

plugins/openai_plugin.py
```diff
@@ -0,0 +1,212 @@
...
+import pytest
+import requests
+from urllib3.exceptions import InsecureRequestWarning
+
+requests.packages.urllib3.disable_warnings(category=InsecureRequestWarning)
+logging.basicConfig(level=logging.DEBUG)
+logger = logging.getLogger(__name__)
+
+
+class OpenAIClient:
```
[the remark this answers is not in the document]
tarukumar: our customer use curl for most of the work and to have equivalent of that i don't want to use the openai library.

plugins/tgis_grpc_plugin.py
tarukumar: it is  mistake . I overlooked it.

plugins/tgis_grpc_plugin.py
```diff
@@ -0,0 +1,124 @@
...
+        Initialize the TGISGRPCPlugin with necessary parameters.
+
+        Args:
+            model_name (str): The model name to use.
+            host (str): The gRPC server host.
+            streaming (bool): Whether to use streaming.
+            use_tls (bool): Whether to use TLS for the connection.
+        """
+        if not all([model_name, host]):
+            raise ValueError("Model name and host are required arguments.")
```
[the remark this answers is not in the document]
tarukumar: This was just to throw error if someone doesn't provide the parameter. But it is not necessary. i will remove it

tarukumar:
I will send the another pr with all the package being  used with new tc . That would make more sense  l. I am going to close this pr.

PR #30, opened 2024-10-30
[WIP] fix isvc for all kind of storage

mwaykole:
is func should work with AWS , GCP , OCI   stoarge

### DIFF
--- a/tests/model_serving/model_server/storage/pvc/utils.py
+++ b/tests/model_serving/model_server/storage/pvc/utils.py
@@ -1,40 +1,84 @@
 from contextlib import contextmanager
+from typing import Any, Dict, Optional
 
 from kubernetes.dynamic import DynamicClient
 from ocp_resources.inference_service import InferenceService
 
-
 @contextmanager
 def create_isvc(
-    client: DynamicClient,
-    name: str,
-    namespace: str,
-    deployment_mode: str,
-    storage_uri: str,
-    model_format: str,
-    runtime: str,
-    min_replicas: int = 1,
-    wait: bool = True,
+        client: DynamicClient,
+        name: str,
+        namespace: str,
+        deployment_mode: str,
+        storage_key: str,
+        storage_uri: str,
+        model_format: str,
+        runtime: str,
+        max_replicas: Optional[int] = 1,
+        min_replicas: Optional[int] = 1,
+        cpu_limit: str = "2",
+        memory_limit: str = "8Gi",
+        cpu_request: str = "1",
+        memory_request: str = "4Gi",
+        wait: bool = True,
+        enable_auth: bool = False,
+        model_service_account: Optional[str] = ""
 ) -> InferenceService:
-    with InferenceService(
-        client=client,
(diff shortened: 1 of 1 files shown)


tests/model_serving/model_server/storage/pvc/utils.py
```diff
@@ -10,31 +11,64 @@ def create_isvc(
     name: str,
     namespace: str,
     deployment_mode: str,
+    storage_key: str,
     storage_uri: str,
     model_format: str,
     runtime: str,
-    min_replicas: int = 1,
+    max_replicas: Optional[int] = 1,
```
rnetser, line 18: """Optional[X] is equivalent to Union[X, None].""" so should be 
```suggestion
    max_replicas: Optional[int] = None,
```
and in the code, add to the dict if max_replicas is not None

tests/model_serving/model_server/storage/pvc/utils.py
```diff
@@ -10,31 +11,64 @@ def create_isvc(
...
+    min_replicas: Optional[int] = 1,
+    cpu_limit: str = "2",
+    memory_limit: str = "8Gi",
+    cpu_request: str = "1",
+    memory_request: str = "4Gi",
```
rnetser, line 23: why do we want to set these defaults? why not have the same values set for limits and requests? what happens if you do not set defaults?

tests/model_serving/model_server/storage/pvc/utils.py
```diff
@@ -10,31 +11,64 @@ def create_isvc(
...
+                "requests": {"cpu": cpu_request, "memory": memory_request},
+            },
+        },
+        "maxReplicas": max_replicas,
+        "minReplicas": min_replicas,
+    }
+
+    if model_service_account:
+        predictor_config["serviceAccountName"] = model_service_account
+
```
rnetser, line 45:
```
base_annotations = {
            "serving.kserve.io/deploymentMode": deployment_mode,
        }

if deployment_mode == "RawDeployment":
    annotations = base_annotations
    ....
else:
    annotations = base_annotations.update(....)
```

tests/model_serving/model_server/storage/pvc/utils.py
```diff
@@ -10,31 +11,64 @@ def create_isvc(
...
+    if deployment_mode == "RawDeployment":
+        annotations = {
+            "serving.kserve.io/deploymentMode": deployment_mode,
+        }
+        labels = {
+            "networking.kserve.io/visibility": "enable-route",
+            "security.opendatahub.io/enable-auth": "true",
```
rnetser, line 52: should be set based on `enable_auth` right?

tests/model_serving/model_server/storage/pvc/utils.py
```diff
@@ -10,31 +11,64 @@ def create_isvc(
...
+        }
+    else:
```
rnetser, line 54: """
elif deployment_mode == "Serverless"
"""

Issue #46, opened 2024-11-26
Use pytest-pspec instead of pytest-progress


lugi0:
[pytest-pspec](https://github.com/gwthm-in/pytest-pspec) seems to provide a nicer terminal output that takes into account docstrings compared to the plugin currently in use (pytest-progress).
If there's enough interest we could switch to this plugin by default

lugi0:
Current output:
![Screenshot 2024-11-26 at 17 10 29](https://github.com/user-attachments/assets/7a140908-8239-46c5-ad38-efa7208cd011)

pytest-pspec output:
<img width="1219" alt="Screenshot 2024-11-26 at 17 06 45" src="https://github.com/user-attachments/assets/777d0f03-9130-44a2-b028-7ed92e349a0a">

PR #47, opened 2024-11-26
Use pytest-pspec instead of pytest-progress

lugi0:
Implementation of https://github.com/opendatahub-io/opendatahub-tests/issues/46

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,11 +41,11 @@ classifiers = [
 dependencies = [
     "openshift-python-utilities",
     "openshift-python-wrapper",
-    "pytest-progress",
     "python-simple-logger",
     "pyyaml",
     "tenacity",
     "types-requests>=2.32.0.20241016",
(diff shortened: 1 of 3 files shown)


rnetser:
@lugi0 looked into using this plugin. i think we should continue using pytest-progress as it provides visibility on tests progress which is important mainly in ci. i think logging the test docs in this context is less valuable. wdyt?

lugi0:
@rnetser I honestly don't like how verbose pytest-progress is most of the time, but I agree that in CI it would be the better choice. 
IMHO we can either use two configs (local, ci) if we wanted to have a common config for all, or we could just default to pytest-progress and let folks modify their local env to best suit their preferences

rnetser:
@lugi0 I prefer not to have 2 configs and as the target here is mainly ci I think we should go with pytest-progress.

rnetser:
@lugi0 can we close this one?

PR #54, opened 2024-12-05
Add first Model Registry tests

lugi0:
## Merge criteria:


lugi0:
Needs https://github.com/RedHatQE/openshift-python-wrapper/pull/2220 to be merged first

tests/model_registry/conftest.py
```diff
@@ -0,0 +1,307 @@
...
+
+@pytest.fixture(scope="class")
+def model_registry_namespace(admin_client: DynamicClient) -> Generator[Namespace, Any, Any]:
+    # This namespace should exist after Model Registry is enabled, but it can also be deleted
+    # from the cluster and does not get reconciled. Fetch if it exists, create otherwise.
+    ns = Namespace(name="rhoai-model-registries", client=admin_client)
+    if ns.exists:
+        yield ns
+    else:
+        with create_ns(
```
tarukumar, line 30: I think logging warning could be helpful here. If it is already deleted and we are recreating it

tests/model_registry/conftest.py
```diff
@@ -0,0 +1,307 @@
...
+    model_registry_db_service: Service,
+) -> Generator[ModelRegistry, Any, Any]:
+    with ModelRegistry(
+        name="model-registry",
+        namespace=model_registry_namespace.name,
+        label={
+            "app.kubernetes.io/name": "model-registry",
+            "app.kubernetes.io/instance": "model-registry",
+            "app.kubernetes.io/part-of": "model-registry-operator",
+            "app.kubernetes.io/managed-by": "kustomize",
```
tarukumar, line 270: I think we can removed kustomize . I don't think it is mandatory, as i understand when we use kustomize tool people tend to add this label like fyi . Is it mandatory?

tarukumar:
overall looks good. Add small comment

tests/conftest.py
```diff
@@ -14,6 +15,14 @@ def admin_client() -> DynamicClient:
     return get_client()
 
 
+@pytest.fixture(scope="session")
+def admin_client_token(admin_client: DynamicClient) -> str:
+    cmd = "oc whoami -t"
```
adolfo-ab, line 20: Only used once, no need to define a literal

tests/model_registry/conftest.py
```diff
@@ -0,0 +1,307 @@
...
+        "label": {
+            "app.kubernetes.io/name": DB_RESOURCES_NAME,
+            "app.kubernetes.io/instance": DB_RESOURCES_NAME,
+            "app.kubernetes.io/part-of": DB_RESOURCES_NAME,
+            "app.kubernetes.io/managed-by": "kustomize",
+        },
+    }
+
+    with PersistentVolumeClaim(**pvc_kwargs) as pvc:
+        yield pvc
```
adolfo-ab, line 93: Why use kwargs here instead of passing the arguments to `PersistentVolumeClaim` as usual?

tests/model_registry/test_model_registry_creation.py
```diff
@@ -0,0 +1,31 @@
...
+LOGGER = get_logger(name=__name__)
+
+
+class TestModelRegistryCreation:
+    """Tests the creation of a model registry"""
+
+    # TODO: Enable Model Registry in DSC if needed
+
+    def test_model_registry_instance_creation(self: Self, model_registry_instance: ModelRegistry):
+        assert model_registry_instance.name == "model-registry"
```
adolfo-ab, line 18: Would it make more sense to use `model_registry_instance.exists`?

tests/model_registry/utils.py
```diff
@@ -0,0 +1,62 @@
...
+from ocp_resources.service import Service
+from ocp_resources.model_registry import ModelRegistry
+from kubernetes.dynamic.exceptions import ResourceNotFoundError
+
+from tests.utils import ProtocolNotSupported
+from utilities.constants import Protocols
+from tests.model_registry.constants import ModelRegistryEndpoints
+
+
+def get_mr_service_by_label(client: DynamicClient, ns: Namespace, mr_instance: ModelRegistry) -> Service:
```
adolfo-ab, line 12: Can we use `ensure_exists` here?

tests/model_registry/utils.py
```diff
@@ -0,0 +1,62 @@
...
+        )
+    ]:
+        return svc[0]
+
+    raise ResourceNotFoundError(f"{mr_instance.name} has no Service")
+
+
+def get_endpoint_from_mr_service(client: DynamicClient, svc: Service, protocol: str) -> str | ProtocolNotSupported:
+    if protocol == Protocols.REST:
+        return svc.instance.metadata.annotations["routing.opendatahub.io/external-address-rest"]
```
adolfo-ab, line 40: Save and reuse `routing.opendatahub.io/external-address-`

tests/model_registry/utils.py
```diff
@@ -0,0 +1,62 @@
...
+    elif protocol == Protocols.GRPC:
+        return svc.instance.metadata.annotations["routing.opendatahub.io/external-address-grpc"]
+    else:
+        raise ProtocolNotSupported(protocol)
+
+
+def generate_register_model_command(endpoint: str, token: str) -> str:
+    auth_header = f" -H 'Authorization: Bearer {token}'"
```
adolfo-ab, line 48: I think `Authorization: Bearer` is used somewhere else, should we add it to a common constans.py?

tests/model_registry/conftest.py
lugi0: I think I copied this from another conftest file and it was defined like this, in the end I'd say it's the same

tests/model_registry/test_model_registry_creation.py
lugi0: The resource fixtures already wait for everything to be available, so when we get here we're pretty sure everything is up and running. This was just a dumb assert to have a test on instance creation.
```
mr_db_deployment.wait_for_replicas(deployed=True)
[...]
mr.wait_for_condition(condition="Available", status="True")
```

tests/model_registry/utils.py
lugi0: I'm not sure what you're suggesting. This function would raise if the Service didn't exist?

adolfo-ab: Yes, it would raise if the service doesn't exist. I'm not sure if you could use it here or not --feel free to resolve this comment if you can't use it.

adolfo-ab (approved):

tarukumar (dismissed):

tests/model_registry/conftest.py
```diff
@@ -0,0 +1,315 @@
...
+
+@pytest.fixture(scope="class")
+def model_registry_db_deployment(
+    admin_client: DynamicClient,
+    model_registry_namespace: Namespace,
+    model_registry_db_secret: Secret,
+    model_registry_db_pvc: PersistentVolumeClaim,
+    model_registry_db_service: Service,
+) -> Generator[Deployment, Any, Any]:
+    with Deployment(
```
rnetser, line 132: is there a template for this deployment in the cluster?

tests/model_registry/conftest.py
```diff
@@ -0,0 +1,315 @@
...
+    model_registry_namespace: Namespace,
+    model_registry_db_deployment: Deployment,
+    model_registry_db_secret: Secret,
+    model_registry_db_service: Service,
+) -> Generator[ModelRegistry, Any, Any]:
+    with ModelRegistry(
+        name="model-registry",
+        namespace=model_registry_namespace.name,
+        label={
+            "app.kubernetes.io/name": "model-registry",
```
rnetser, line 265: `model-registry` - save and re-use
`model-registry-operator` - save and re-use

if something changes, we then change in one place

tests/model_registry/test_model_registry_creation.py
rnetser: i agree with @adolfo-ab 
the name is set in the fixture so at this point in the test, this will always be true.
if the test is to verify creation, then the validation should be in the test not the fixture.

tests/model_registry/test_rest_api.py
```diff
@@ -0,0 +1,20 @@
...
+import pytest
+from schemathesis.checks import ignored_auth
+from simple_logger.logger import get_logger
+
+LOGGER = get_logger(name=__name__)
+
+schema = schemathesis.from_pytest_fixture("generate_schema")
+
+
+@pytest.mark.skip(reason="Only run manually for now")
```
rnetser, line 11: why?


please add `TODO` with what needs to be done to remove this skip

tests/model_registry/test_rest_api.py
```diff
@@ -0,0 +1,20 @@
+import schemathesis
+import pytest
+from schemathesis.checks import ignored_auth
+from simple_logger.logger import get_logger
+
+LOGGER = get_logger(name=__name__)
+
+schema = schemathesis.from_pytest_fixture("generate_schema")
```
rnetser, line 8: why do we need to wrap with calling the fixture and not using `from_uri` with the uri?

tests/model_registry/utils.py
```diff
@@ -0,0 +1,65 @@
...
+    """
+    if svc := [
+        svcs
+        for svcs in Service.get(
+            dyn_client=client,
+            namespace=ns.name,
+            label_selector=f"app={mr_instance.name},component=model-registry",
+        )
+    ]:
+        return svc[0]
```
rnetser, line 36: can there be more than one? if yes - why return the 1st?

tests/model_registry/utils.py
```diff
@@ -0,0 +1,65 @@
...
+
+    raise ResourceNotFoundError(f"{mr_instance.name} has no Service")
+
+
+def get_endpoint_from_mr_service(client: DynamicClient, svc: Service, protocol: str) -> str | ProtocolNotSupported:
```
rnetser, line 41:
```suggestion
def get_endpoint_from_mr_service(client: DynamicClient, svc: Service, protocol: str) -> str:
```

no need to include the raise

utilities/constants.py
```diff
@@ -42,6 +42,14 @@ class Protocols:
     HTTP: str = "http"
     HTTPS: str = "https"
     GRPC: str = "grpc"
+    REST: str = "rest"
+
+
+class HTTPRequest:
+    # Pay attention to the single `'`, needs to be closed after your token
+    # See e.g. tests/model_registry/utils.py#L51
+    AUTH_HEADER: str = "-H 'Authorization: Bearer"
```
rnetser, line 51: this will always be used with a token so how about one of:

```
AUTH_HEADER: str = "-H 'Authorization: Bearer {token}'"
AUTH_HEADER.format(token="xx")

```

or
```
from string import Template
 AUTH_HEADER = Template("-H 'Authorization: Bearer $token'")
AUTH_HEADER.substitute(token="xx")

```

rnetser (changes requested):

tests/model_registry/conftest.py
```diff
@@ -0,0 +1,315 @@
...
+
+LOGGER = get_logger(name=__name__)
+
+DB_RESOURCES_NAME = "model-registry-db"
+
+
+@pytest.fixture(scope="class")
+def model_registry_namespace(admin_client: DynamicClient) -> Generator[Namespace, Any, Any]:
+    # This namespace should exist after Model Registry is enabled, but it can also be deleted
+    # from the cluster and does not get reconciled. Fetch if it exists, create otherwise.
```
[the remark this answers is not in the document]
lugi0: No

tests/model_registry/conftest.py
```diff
@@ -0,0 +1,315 @@
...
+    ns = Namespace(name="rhoai-model-registries", client=admin_client)
+    if ns.exists:
+        yield ns
+    else:
+        LOGGER.warning("rhoai-model-registries namespace was not present, creating it")
+        with create_ns(
+            name="rhoai-model-registries",
+            admin_client=admin_client,
+            teardown=False,
```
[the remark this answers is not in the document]
lugi0: In this case I think I disagree - the Namespace is expected to be there on a "standard" cluster we conduct testing on. If it's missing it might be because of an error in setup or some other test that somehow deleted it. I'd vote to keep it as is.

tests/model_registry/test_model_registry_creation.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def test_registering_model(self: Self, model_registry_instance_rest_endpoint: str, admin_client_token: str):
+        cmd = generate_register_model_command(model_registry_instance_rest_endpoint, admin_client_token)
+        res, out, _ = run_command(command=shlex.split(cmd), verify_stderr=False, check=False)
+        assert res
```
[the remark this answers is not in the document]
lugi0: What will it handle for me?

tests/model_registry/test_rest_api.py
lugi0: Because I want it to use the real host of the MR instance rather than grab the localhost from the openAPI spec. In order to do that and pass it as a base_url, I need it to be in a fixture so that I can wait for the MR instance and grab it.
If I were to call it as from_uri here, it would not be doable, as this happens before the tests are run and it does not have the ability to see the fixtures yet.

tests/model_registry/utils.py
lugi0: No, the label selector should guarantee that only one is ever returned. One MR instance = One service.

utilities/constants.py
lugi0: In every case we need to be aware of the need to handle the token somehow, so I honestly don't see much of a difference. I think your first suggestion is perhaps the cleanest, so I might go with that.

CONTRIBUTING.md
```diff
@@ -36,3 +36,29 @@ pre-commit install
 If you use Visual Studio Code as your IDE, we recommend using the [Mypy Type Checker](https://marketplace.visualstudio.com/items?itemName=ms-python.mypy-type-checker) extension.
 After installing it, make sure to update the `Mypy-type-checkers: Args` setting
 to `"mypy-type-checker.args" = ["--config-file=pyproject.toml"]`.
+
+### Debugging in Visual Studio Code
```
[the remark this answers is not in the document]
lugi0: That feels overkill right now, I wasn't even sure if it was worth it to have this in the first place as at the end of the day anyone is free to use whatever editor/IDE they want.

tests/conftest.py
```diff
@@ -15,6 +16,13 @@ def admin_client() -> DynamicClient:
     return get_client()
 
 
+@pytest.fixture(scope="session")
```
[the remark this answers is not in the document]
lugi0: >you can check by getting "oc whoami" first

Well, that'd only tell us the name of the user, not if it was an admin or not.

tests/model_registry/conftest.py
```diff
@@ -0,0 +1,315 @@
...
+    ) as mr_db_service:
+        yield mr_db_service
+
+
+@pytest.fixture(scope="class")
+def model_registry_db_pvc(
+    admin_client: DynamicClient,
+    model_registry_namespace: Namespace,
+) -> Generator[PersistentVolumeClaim, Any, Any]:
+    pvc_kwargs = {
```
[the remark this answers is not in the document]
lugi0: I think it does not make any difference in this context

tests/model_registry/conftest.py
lugi0: No

tests/conftest.py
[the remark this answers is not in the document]
lugi0: With this command we could get the list of cluster admin users
`oc get clusterrolebindings -o json |jq '.items[] |select(.roleRef.name=="cluster-admin") |.subjects[] |select(.kind=="User") |.name'`
but honestly it makes more sense to me to rename this to `current_client_token`

tests/model_registry/test_model_registry_creation.py
```diff
@@ -0,0 +1,31 @@
...
+        out_dict = eval(out)
+        assert out_dict["name"] == "model-name"
+        assert out_dict["description"] == "test-model"
+        assert out_dict["externalId"] == "1"
+        assert out_dict["owner"] == "opendatahub-tests-client"
+        assert out_dict["state"] == "LIVE"
```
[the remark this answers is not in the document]
lugi0: We could also look into https://github.com/astraw38/pytest-assume

CONTRIBUTING.md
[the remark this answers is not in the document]
rnetser: but we can make their lives easier :) but i think that the contirbuting guide should be only about that.

tests/conftest.py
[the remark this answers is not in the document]
rnetser: in https://github.com/opendatahub-io/opendatahub-tests/pull/54/files#r1881639137 i meant to check if the current context is admin then return the token.

so if you do not need admin, but current token, please use https://github.com/opendatahub-io/opendatahub-tests/blob/main/tests/trustyai/conftest.py#L55

tests/model_registry/conftest.py
```diff
@@ -0,0 +1,318 @@
...
+
+@pytest.fixture(scope="class")
+def model_registry_namespace(admin_client: DynamicClient) -> Generator[Namespace, Any, Any]:
+    # This namespace should exist after Model Registry is enabled, but it can also be deleted
+    # from the cluster and does not get reconciled. Fetch if it exists, create otherwise.
+    ns = Namespace(name="rhoai-model-registries", client=admin_client)
+    if ns.exists:
+        yield ns
+    else:
+        LOGGER.warning("rhoai-model-registries namespace was not present, creating it")
```
rnetser, line 33: please save `rhoai-model-registries` and re-use

tests/model_registry/conftest.py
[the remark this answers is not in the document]
rnetser: why set the dict up and then call it instead of using it directly?

tests/model_registry/conftest.py
```diff
@@ -0,0 +1,318 @@
...
+                "targetPort": 3306,
+            }
+        ],
+        selector={
+            "name": DB_RESOURCES_NAME,
+        },
+        label={
+            KubernetesAnnotations.NAME: DB_RESOURCES_NAME,
+            KubernetesAnnotations.INSTANCE: DB_RESOURCES_NAME,
+            KubernetesAnnotations.PART_OF: DB_RESOURCES_NAME,
```
rnetser, line 67: repeated - please save and re-use

tests/model_registry/test_model_registry_creation.py
rnetser: if the test always passes (because all checks are done before) then it should be removed

tests/model_registry/test_model_registry_creation.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def test_model_registry_instance_creation(self: Self, model_registry_instance: ModelRegistry):
+        assert model_registry_instance.name == "model-registry"
+
+    def test_registering_model(self: Self, model_registry_instance_rest_endpoint: str, current_client_token: str):
+        cmd = generate_register_model_command(
+            endpoint=model_registry_instance_rest_endpoint, token=current_client_token
+        )
+        _, out, _ = run_command(command=shlex.split(cmd))
+        out_dict = eval(out)
```
rnetser, line 25: better use `ast.literal_eval`

tests/model_registry/utils.py
rnetser: what if there's a bug and on MR has more than 1 service?

rnetser (changes requested):

utilities/constants.py
```diff
@@ -42,10 +42,25 @@ class Protocols:
     HTTP: str = "http"
     HTTPS: str = "https"
     GRPC: str = "grpc"
+    REST: str = "rest"
+
+
+class HTTPRequest:
+    # You will need to `.format(token='foo')`
+    # See e.g. tests/model_registry/utils.py#L51
```
rnetser, line 50: please remove as we do not want to maintain this if lines change; i'd expect a user to understand how to use with your given example

you may want to consider rephrasing the 1st line to 
```
use string formatting to set the token value
```

tests/model_registry/utils.py
```diff
@@ -0,0 +1,65 @@
...
+        return svc.instance.metadata.annotations[f"{ADDRESS_ANNOTATION_PREFIX}{Protocols.GRPC}"]
+    else:
+        raise ProtocolNotSupported(protocol)
+
+
+def generate_register_model_command(endpoint: str, token: str) -> str:
+    auth_header = f" {HTTPRequest.AUTH_HEADER.format(token=token)}"
+    content_header = f" {HTTPRequest.CONTENT_JSON}"
+    data = ' -d \'{"name": "model-name", "description": "test-model", "owner": "opendatahub-tests-client", "externalId": "1", "state": "LIVE"}\''
+    cmd = (
```
rnetser, line 54: - please use f-string format 
- `auth_header` and `content_header` can be use directly here and not set and then used

rnetser (changes requested):

tests/conftest.py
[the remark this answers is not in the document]
lugi0: I don't agree in using a fixture from the trustyai folder for an operation that everyone will need sooner or later. I think in future PRs it would be better to refactor TrustyAI to use this new fixture.

tests/model_registry/conftest.py
lugi0: These are already constants

tests/model_registry/utils.py
lugi0: >please use f-string format

It is already being used. If you meant in `data`, I don't think it makes sense right now since it is only sending one request.

adolfo-ab (approved):

tarukumar (dismissed):

lugi0:
Do not merge right now, still needs testing

tests/trustyai/drift/test_drift.py
```diff
@@ -23,14 +23,14 @@ class TestDriftMetrics:
     def test_send_inference_request_and_verify_trustyai_service(
         self,
         admin_client,
-        openshift_token,
+        current_client_token,
```
rnetser, line 26: `openshift_token` should be deleted

rnetser (changes requested):

tests/trustyai/drift/test_drift.py
lugi0: Do you mean the fixture?

tests/conftest.py
```diff
@@ -41,6 +42,13 @@ def tests_tmp_dir(request: FixtureRequest, tmp_path_factory: TempPathFactory) ->
     shutil.rmtree(str(tests_tmp_path), ignore_errors=True)
 
 
+@pytest.fixture(scope="session")
+def current_client_token(admin_client: DynamicClient) -> str:
+    _, out, _ = run_command(command=shlex.split("oc whoami -t"), verify_stderr=False, check=False)
```
rnetser, line 47: please use https://github.com/opendatahub-io/opendatahub-tests/blob/main/utilities/infra.py#L299 `get_openshift_token`

tests/model_registry/test_model_registry_creation.py
```diff
@@ -0,0 +1,41 @@
...
+
+LOGGER = get_logger(name=__name__)
+
+
+class TestModelRegistryCreation:
+    """Tests the creation of a model registry"""
+
+    # TODO: Enable Model Registry in DSC if needed
+
+    def test_model_registry_instance_creation(self: Self, model_registry_instance: ModelRegistry):
```
rnetser, line 18: any test markers in terms of tiers? smoke etc

tests/model_registry/test_model_registry_creation.py
```diff
@@ -0,0 +1,41 @@
...
+        assert model_registry_instance.exists
```
rnetser, line 19: you already have the check in `model_registry_instance` 
`wait_for_condition` is  waiting on `self.exists` so if the resource does not exist the setup will not pass leading to pytest error.
if setup succeeds, the test will always pass.
`wait_for_resource=True` will also make sure that the resource is created.

tests/model_registry/test_model_registry_creation.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def test_registering_model(self: Self, model_registry_instance_rest_endpoint: str, current_client_token: str):
+        cmd = generate_register_model_command(
+            endpoint=model_registry_instance_rest_endpoint, token=current_client_token
+        )
+        _, out, _ = run_command(command=shlex.split(cmd))
+        out_dict = literal_eval(out)
+        errors = []
+        if not out_dict["name"] == "model-name":
```
rnetser, line 28: will you always have these keys or would it be better to use `get`?

tests/model_registry/utils.py
```diff
@@ -0,0 +1,72 @@
...
+        svcs
+        for svcs in Service.get(
+            dyn_client=client,
+            namespace=ns.name,
+            label_selector=f"app={mr_instance.name},component=model-registry",
+        )
+    ]:
+        if len(svc) > 1:
+            raise TooManyServices(svc)
+        return svc[0]
```
rnetser, line 46: a suggestion, i think it is clearer to first have the positive path and then fail, i.e
```suggestion
        if len(svc) == 1:
            return svc[0]

        raise TooManyServices(svc)
```

tests/model_registry/utils.py
```diff
@@ -0,0 +1,72 @@
...
+    #     "curl -k "
+    #     + f" {HTTPRequest.AUTH_HEADER.format(token=token)}"
+    #     + f" {HTTPRequest.CONTENT_JSON}"
+    #     + data
+    #     + " "
+    #     + Protocols.HTTPS
+    #     + "://"
+    #     + endpoint
+    #     + ModelRegistryEndpoints.REGISTERED_MODELS
+    # )
```
rnetser, line 70: leftover?

tests/model_registry/utils.py
```diff
@@ -0,0 +1,72 @@
...
+    cmd = f'curl -k {HTTPRequest.AUTH_HEADER.format(token=token)} {HTTPRequest.CONTENT_JSON} -d \'{{"name": "model-name", "description": "test-model", "owner": "opendatahub-tests-client", "externalId": "1", "state": "LIVE"}}\' {Protocols.HTTPS}://{endpoint}{ModelRegistryEndpoints.REGISTERED_MODELS}'
```
rnetser, line 71:
```suggestion
   return f'curl -k {HTTPRequest.AUTH_HEADER.format(token=token)} {HTTPRequest.CONTENT_JSON} -d \'{{"name": "model-name", "description": "test-model", "owner": "opendatahub-tests-client", "externalId": "1", "state": "LIVE"}}\' {Protocols.HTTPS}://{endpoint}{ModelRegistryEndpoints.REGISTERED_MODELS}'
```

tests/utils.py
```diff
@@ -0,0 +1,6 @@
+class ProtocolNotSupported(Exception):
```
rnetser, line 1: please create `utilities/exceptions.py` and put that code there

utilities/constants.py
```diff
@@ -88,8 +88,15 @@ class Protocols:
...
-    TCP_PROTOCOLS: set[str] = {HTTP, HTTPS}
-    ALL_SUPPORTED_PROTOCOLS: set[str] = TCP_PROTOCOLS.union({GRPC})
+    REST: str = "rest"
+    TCP_PROTOCOLS: set[str] = {"HTTP", "HTTPS"}
+    ALL_SUPPORTED_PROTOCOLS: set[str] = TCP_PROTOCOLS.union({"GRPC"})
+
+
+class HTTPRequest:
+    # Use string formatting to set the token value when using this constant
+    AUTH_HEADER: str = "-H 'Authorization: Bearer {token}'"
```
rnetser, line 98: please also use in https://github.com/opendatahub-io/opendatahub-tests/blob/main/utilities/inference_utils.py#L197

utilities/constants.py
```diff
@@ -99,6 +106,13 @@ class AcceleratorType:
     SUPPORTED_LISTS: list[str] = [NVIDIA, AMD, GAUDI]
 
 
+class KubernetesAnnotations:
+    NAME: str = "app.kubernetes.io/name"
```
rnetser, line 110:
```suggestion
    NAME: str = f"{Resource.ApiGroup.APP_KUBERNETES_IO}/name"
```

rnetser (changes requested):

tests/model_registry/test_model_registry_creation.py
lugi0: Unless the API changes the response will remain consistent

adolfo-ab (approved):

tarukumar (approved):

rnetser (approved):

PR #59, opened 2024-12-15
Add RawDeployment route tests

rnetser:


tests/model_serving/model_server/raw_deployment/conftest.py
```diff
@@ -0,0 +1,45 @@
+import pytest
```
mwaykole, line 1: please add proper label according to new raw kserve

mwaykole (changes requested):

tests/model_serving/model_server/raw_deployment/conftest.py
rnetser: which ones? added if auth or route are needed

PR #66, opened 2024-12-24
[Smoke] migrate `Verify Model Serving Installation` from `ods-ci`

rnetser:
depends on https://github.com/opendatahub-io/opendatahub-tests/pull/59


rnetser:
verified

PR #95, opened 2025-01-16
Reduce isvc fail timeout, fix `TCP_PROTOCOLS` constant, add `Error` ro exception class names

rnetser:
## Merge criteria:


mwaykole (dismissed):

utilities/constants.py
```diff
@@ -90,7 +90,7 @@ class Protocols:
     HTTPS: str = "https"
     GRPC: str = "grpc"
     REST: str = "rest"
-    TCP_PROTOCOLS: set[str] = {"HTTP", "HTTPS"}
+    TCP_PROTOCOLS: set[str] = {HTTP, HTTPS}
     ALL_SUPPORTED_PROTOCOLS: set[str] = TCP_PROTOCOLS.union({"GRPC"})
```
lugi0, line 94: You probably need to change this one too

utilities/exceptions.py
```diff
@@ -23,21 +23,21 @@ class InferenceResponseError(Exception):
...
         self.storage_path = storage_path
 
     def __str__(self) -> str:
         msg = f"""
             You've passed the following parameters:
-            "storageUri": {self.storageUri}
+            "storage_uri": {self.storageUri}
             "storage_key": {self.storage_key}
             "storage_path: {self.storage_path}
             In order to create a valid ISVC you need to specify either a storageUri value
```
lugi0, line 43: storageUri to storage_uri

rnetser: because python vars should be lower case

lugi0 (changes requested):
couple changes otherwise lgtm

utilities/exceptions.py
lugi0: I meant you should also update the name in the docstring at L43, sorry

utilities/constants.py
rnetser: done

utilities/exceptions.py
rnetser: done

lugi0 (approved):

PR #96, opened 2025-01-19
flake8 - add FunctionCallForceNames

rnetser:


tests/model_registry/conftest.py
```diff
@@ -68,7 +68,7 @@ def model_registry_db_service(
         },
         label=DEFAULT_LABEL_DICT_DB,
         annotations={
-            "template.openshift.io/expose-uri": "mysql://{.spec.clusterIP}:{.spec.ports[?(.name==\mysql\)].port}",
+            "template.openshift.io/expose-uri": r"mysql://{.spec.clusterIP}:{.spec.ports[?(.name==\mysql\)].port}",
```
lugi0, line 71: why are you making this a raw string? does the selector `.name==\mysql\` still work like this?

utilities/general.py
```diff
@@ -69,7 +69,7 @@ def download_model_data(
     containers = [
         {
             "name": "model-downloader",
-            "image": "quay.io/modh/kserve-storage-initializer@sha256:330af2d517b17dbf0cab31beba13cdbe7d6f4b9457114dea8f8485a011e3b138",
+            "image": "quay.io/modh/kserve-storage-initializer@sha256:330af2d517b17dbf0cab31beba13cdbe7d6f4b9457114dea8f8485a011e3b138",  # noqa: E501
```
lugi0, line 72: this might not be needed due to https://github.com/opendatahub-io/opendatahub-tests/pull/97/files#diff-ac6c824bcc7455fa0967c33f2062f5c5cf7cee2b62d6402eec5b39c359530e02R78

lugi0 (commented):
can you make sure the raw string still works here before we merge?

utilities/general.py
rnetser: yes; just to pass flake8 check in this pr

tests/model_registry/conftest.py
rnetser: yes, tested with and without

lugi0 (approved):

mwaykole (approved):

PR #97, opened 2025-01-19
Use `data connection` secret to store s3 information

rnetser:
by using `data connection` secrets, storage-config secret will be auto-generated by odh-model-controller and this is currently the recommended way (and the one used by the dashboard) to create a connection secret


utilities/constants.py
```diff
@@ -112,6 +112,7 @@ class KubernetesAnnotations:
     INSTANCE: str = f"{Resource.ApiGroup.APP_KUBERNETES_IO}/instance"
     PART_OF: str = f"{Resource.ApiGroup.APP_KUBERNETES_IO}/part-of"
     CREATED_BY: str = f"{Resource.ApiGroup.APP_KUBERNETES_IO}/created-by"
+    DEPLOYMENT_MODE: str = "serving.kserve.io/deploymentMode"
```
lugi0, line 115: Wouldn't it be better to put this under a different class, e.g. `KserveAnnotations`, or rename the class to be more general?

utilities/infra.py
```diff
@@ -116,7 +108,9 @@ def s3_endpoint_secret(
         name=name,
         namespace=namespace,
         annotations={"opendatahub.io/connection-type": "s3"},
-        data_dict=get_s3_secret_dict(
+        # the labels are needed to set the secret as data connection by odh-model-controller
+        label={"opendatahub.io/managed": "true", "opendatahub.io/dashboard": "true"},
```
lugi0, line 112: I don't really like the fact that we're putting a label here that signals this resource was created by the dashboard when it isn't so. If we want to test in a way that is decouple from the dashboard we shouldn't rely on its logic and behaviour IMHO

lugi0 (commented):
I'm not sure this is the right approach moving forward, we're coupling ourselves with dashboard's behaviour over which we have no control

utilities/infra.py
rnetser: i do not like it either, see https://github.com/opendatahub-io/odh-model-controller/issues/348
but this how the product is working now

utilities/constants.py
rnetser: split into subclasses

lugi0 (dismissed):

lugi0 (approved):

mwaykole (approved):

PR #99, opened 2025-01-22
Add fixture to set managementState for components

lugi0:
## Description

Added two fixtures, one to get the DSC resource in the cluster which is run in a session scope and the other to patch the managementState of a given component to the desired state with ResourceEditor, which will revert the change after the fixture call is done. 
For now it has been set with scope "module", assuming that a given component needs to be modified for the whole test module.
There is an accompanying helper function that gets the current state of a component, to avoid a useless patch if the desired state is the one already set in the DSC.
Small improvement to the "test_model_registry_creation" test module to enable the MR component if currently set to Removed (default behaviour while the component is in Tech Preview)

## How Has This Been Tested?

Tested running the model registry creation test with the component state set to both Managed and Removed beforehand

## Merge criteria:

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -245,11 +245,9 @@ def unprivileged_client(
 @pytest.fixture(scope="session")
 def dsc_resource(admin_client: DynamicClient):
     name = py_config["dsc_name"]
-    try:
-        for dsc in DataScienceCluster.get(dyn_client=admin_client, name=name):
-            yield dsc
-    except Exception:
-        raise ResourceNotFoundError
+    for dsc in DataScienceCluster.get(dyn_client=admin_client, name=name):
+        return dsc
+    raise ResourceNotFoundError(f"DSC resource {name} not found")
 
 
 @pytest.fixture(scope="module")
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -12,8 +12,9 @@ from ocp_resources.secret import Secret
 from pyhelper_utils.shell import run_command
 from pytest import FixtureRequest, Config
 from kubernetes.dynamic import DynamicClient
+from ocp_resources.data_science_cluster import DataScienceCluster
 from ocp_resources.namespace import Namespace
-from ocp_resources.resource import get_client
+from ocp_resources.resource import get_client, ResourceEditor
 from pytest_testconfig import config as py_config
 from simple_logger.logger import get_logger
 
@@ -238,3 +239,38 @@ def unprivileged_client(
 
         else:
             yield admin_client
+
+
+@pytest.fixture(scope="session")
+def dsc_resource(admin_client: DynamicClient, name: str = "default-dsc"):
+    for dsc in DataScienceCluster.get(dyn_client=admin_client, name=name):
+        yield dsc
+
+
+def get_dsc_component_state(dsc_resource: DataScienceCluster, component_name: str) -> str:
+    return dsc_resource.instance.spec.components[component_name].managementState
+
+
+@pytest.fixture(scope="module")
+def set_dsc_component_state(
+    dsc_resource: DataScienceCluster,
+    request: FixtureRequest,
+) -> Generator[DataScienceCluster, Any, Any]:
+    component_name = getattr(request.module, "component_name")
+    desired_state = getattr(request.module, "desired_state")
+    if get_dsc_component_state(dsc_resource=dsc_resource, component_name=component_name) != desired_state:
+        with ResourceEditor(
+            patches={
+                dsc_resource: {
+                    "spec": {
+                        "components": {
+                            f"{component_name}": {"managementState": f"{desired_state}"},
+                        }
+                    }
+                }
+            }
+        ):
+            yield dsc_resource
+    else:
+        LOGGER.warning(f"Component {component_name} was already set to managementState {desired_state}")
+        yield dsc_resource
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -238,3 +239,38 @@ def unprivileged_client(
 
         else:
             yield admin_client
+
+
+@pytest.fixture(scope="session")
+def dsc_resource(admin_client: DynamicClient, name: str = "default-dsc"):
+    for dsc in DataScienceCluster.get(dyn_client=admin_client, name=name):
+        yield dsc
+
```
rnetser, line 248:
```suggestion
    for dsc in DataScienceCluster.get(dyn_client=admin_client, name=name):
        yield dsc
    raise ResourceNotFound....
```
if dsc is not there we should raise

tests/conftest.py
```diff
@@ -238,3 +239,38 @@ def unprivileged_client(
 
         else:
             yield admin_client
+
+
+@pytest.fixture(scope="session")
+def dsc_resource(admin_client: DynamicClient, name: str = "default-dsc"):
```
rnetser, line 245: name: str = "default-dsc" - you should use `request` and then `request.param` to control the name
i think that this should be set in global_config.py

tests/conftest.py
```diff
@@ -238,3 +239,38 @@ def unprivileged_client(
...
+    for dsc in DataScienceCluster.get(dyn_client=admin_client, name=name):
+        yield dsc
+
+
+def get_dsc_component_state(dsc_resource: DataScienceCluster, component_name: str) -> str:
```
rnetser, line 250: used once and it is just wrapping the access to the resource `managementState` . use directly where needed

tests/conftest.py
```diff
@@ -238,3 +239,38 @@ def unprivileged_client(
...
+    return dsc_resource.instance.spec.components[component_name].managementState
+
+
+@pytest.fixture(scope="module")
+def set_dsc_component_state(
```
rnetser, line 255: a fixture name should be a noun and not a verb; i.e describes the result.
e.g updated_dsc_component_state`

tests/conftest.py
```diff
@@ -238,3 +239,38 @@ def unprivileged_client(
...
+    dsc_resource: DataScienceCluster,
+    request: FixtureRequest,
```
rnetser, line 257: just a nitpick - try to keep fixture order so pytest ones are called first

tests/conftest.py
```diff
@@ -238,3 +239,38 @@ def unprivileged_client(
...
+) -> Generator[DataScienceCluster, Any, Any]:
+    component_name = getattr(request.module, "component_name")
+    desired_state = getattr(request.module, "desired_state")
```
rnetser, line 260: when more than one param is needed, please use pytest.params with a dict to pass values so it is clear when calling the fixture what is passed
e.g

@pytest.mark.parametrize(
    "fixture_name",
[
        pytest.param(
            {"component_name": "xxxx", "desired_state": "yyy"},
        )
    ],
    indirect=True,
)

tests/conftest.py
```diff
@@ -238,3 +239,38 @@ def unprivileged_client(
...
+    if get_dsc_component_state(dsc_resource=dsc_resource, component_name=component_name) != desired_state:
+        with ResourceEditor(
```
rnetser, line 262: note that  `ResourceEditor` does the check

```
LOGGER.warning(f"ResourceEdit: no diff found in patch for {resource.name} -- skipping")
```

tests/model_registry/test_model_registry_creation.py
```diff
@@ -9,12 +9,18 @@
...
 LOGGER = get_logger(name=__name__)
 
+component_name = "modelregistry"
+desired_state = "Managed"
 
-class TestModelRegistryCreation:
-    """Tests the creation of a model registry"""
 
-    # TODO: Enable Model Registry in DSC if needed
+@pytest.mark.usefixtures("set_dsc_component_state")
```
rnetser, line 16: please see my comment on `set_dsc_component_state`

rnetser (changes requested):

tests/conftest.py
lugi0: Mh, but if you didn't know you wouldn't be able to tell from looking at this code. Honestly I'd rather keep it explicit here.

tests/conftest.py
lugi0: Rather than use request/request.param, I think this is a good candidate for using py_config. The name of the DSC is known and stable, and if we need to update it in some specific scenario we can override it like we do for applications_namespace

tests/conftest.py
```diff
@@ -238,3 +240,41 @@ def unprivileged_client(
...
+
+
+@pytest.fixture(scope="session")
+def dsc_resource(admin_client: DynamicClient):
+    name = py_config["dsc_name"]
+    try:
+        for dsc in DataScienceCluster.get(dyn_client=admin_client, name=name):
+            yield dsc
+    except Exception:
+        raise ResourceNotFoundError
```
rnetser, line 252: how about:

```suggestion
    name = py_config["dsc_name"]
    for dsc in DataScienceCluster.get(dyn_client=admin_client, name=name):
            return dsc

        raise ResourceNotFoundError(f"(could not find {name} DCS")
```
no need to yield here as there is not teardown

rnetser (approved):

mwaykole (approved):

PR #100, opened 2025-01-22
Use Model Registry's python client for its test

lugi0:
## Description

Switch from using a direct API call with curl to using the official python library/client for model registry

## How Has This Been Tested?

Ran the test, which passed

## Merge criteria:

On hold until an update to model registry's client is made


lugi0:
On hold until https://github.com/kubeflow/model-registry/pull/726 is merged and released

tests/model_registry/test_model_registry_creation.py
```diff
@@ -17,23 +15,49 @@ class TestModelRegistryCreation:
...
+            storage_key="my-data-connection",
+            storage_path="path/to/model",
+            metadata={
+                # can be one of the following types
+                "int_key": 1,
+                "bool_key": False,
+                "float_key": 3.14,
+                "str_key": "str_value",
+            },
+        )
```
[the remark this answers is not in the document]
lugi0: No, this is the operation that registers the model to the registry, it is part of the test

lugi0:
https://github.com/kubeflow/model-registry/pull/726 has been merged and v0.2.13 of the library includes the needed changes, this can now be merged

tests/model_registry/test_model_registry_creation.py
```diff
@@ -32,28 +31,51 @@ class TestModelRegistryCreation:
     # TODO: Switch to Python client
     @pytest.mark.smoke
     def test_registering_model(
-        self: Self,
-        model_registry_instance_rest_endpoint: str,
-        current_client_token: str,
-        updated_dsc_component_state,
+        self: Self, model_registry_instance_rest_endpoint: str, current_client_token: str, updated_dsc_component_state
```
rnetser, line 34: nitpick  `updated_dsc_component_state` is missing typing (i do not think that typing is needed in tests as they are not called directly / typing here can provide additional info)

tests/model_registry/test_model_registry_creation.py
```diff
@@ -32,28 +31,51 @@ class TestModelRegistryCreation:
...
     ):
-        cmd = generate_register_model_command(
-            endpoint=model_registry_instance_rest_endpoint, token=current_client_token
+        # address and port need to be split in the client instantiation
+        server = model_registry_instance_rest_endpoint.split(":")[0]
+        port = model_registry_instance_rest_endpoint.split(":")[1]
```
rnetser, line 38:
```suggestion
        server, port = model_registry_instance_rest_endpoint.split(":")
```
can assign both directly

tests/model_registry/test_model_registry_creation.py
```diff
@@ -32,28 +31,51 @@ class TestModelRegistryCreation:
...
+            name="my-model",  # model name
+            uri="https://storage-place.my-company.com",  # model URI
+            version="2.0.0",
+            description="lorem ipsum",
+            model_format_name="onnx",
+            model_format_version="1",
+            storage_key="my-data-connection",
+            storage_path="path/to/model",
+            metadata={
+                # can be one of the following types
```
rnetser, line 56: the comment to me is a bit confusing - what does it mean `can be one of the following types`? in which usage?

tests/model_registry/test_model_registry_creation.py
```diff
@@ -32,28 +31,51 @@ class TestModelRegistryCreation:
...
+                "int_key": 1,
+                "bool_key": False,
+                "float_key": 3.14,
+                "str_key": "str_value",
+            },
         )
-        _, out, _ = run_command(command=shlex.split(cmd))
-        out_dict = literal_eval(node_or_string=out)
+        registered_model = registry.get_registered_model("my-model")
```
rnetser, line 63:
```suggestion
        registered_model = registry.get_registered_model(model.name)
```
?

tests/model_registry/test_model_registry_creation.py
```diff
@@ -32,28 +31,51 @@ class TestModelRegistryCreation:
...
+        model = registry.register_model(
+            name="my-model",  # model name
+            uri="https://storage-place.my-company.com",  # model URI
+            version="2.0.0",
+            description="lorem ipsum",
+            model_format_name="onnx",
+            model_format_version="1",
+            storage_key="my-data-connection",
+            storage_path="path/to/model",
+            metadata={
```
rnetser, line 55: should be in a fixture; if this fails for some reason (e.g something changed from client side, this is not a test failure but setup error)

lugi0: I disagree, this is also implicitly testing the Python client. If something changes in instantiation it is a failure I want to be aware of

tests/model_registry/test_model_registry_creation.py
lugi0: When adding metadata to a model you're registering in MR, just there for easy accessibility while developing in the same file new tests

tests/model_registry/test_model_registry_creation.py
lugi0: I think I'd rather keep it separate here, because then I do 
```
if not registered_model.name == model.name:
            errors.append(f"Unexpected name, received {registered_model.name}")
```
checking that was is returned and what is present in model.name is the same is part of the test

adolfo-ab (approved):

tests/model_registry/test_model_registry_creation.py
lugi0: Removed a few comments from the latest version, resolving this thread

tests/model_registry/test_model_registry_creation.py
rnetser: so please save "my-model" and re-use

rnetser (changes requested):

tests/model_registry/test_model_registry_creation.py
rnetser: > this is also implicitly testing the Python client.
why is testing the client part of this repo? should be part of the client's repo.

tests/model_registry/test_model_registry_creation.py
lugi0: done

tests/model_registry/test_model_registry_creation.py
lugi0: Same idea for all of our components under test, upstream vs downstream testing :)

rnetser (changes requested):

tests/model_registry/test_model_registry_creation.py
```diff
@@ -1,14 +1,14 @@
...
 
-from tests.model_registry.utils import generate_register_model_command
-from utilities.constants import DscComponents
+from ocp_resources.data_science_cluster import DataScienceCluster
+from utilities.constants import Protocols, DscComponents
+from model_registry import ModelRegistry
+
 
 LOGGER = get_logger(name=__name__)
+model_name: str = "my-model"
```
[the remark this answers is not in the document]
lugi0: I will most likely reuse it in upcoming additional tests

tests/model_registry/test_model_registry_creation.py
lugi0: Yes, there won't be a downstream distribution of it, but it is still to be used and tested in conjunction with the downstream bits of MR. IMHO it is a very valid testing scenario.

rnetser: so the tests should be in https://github.com/kubeflow/model-registry like https://github.com/kubeflow/model-registry/blob/main/test/python/test_mr_conn.py

tests/model_registry/test_model_registry_creation.py
```diff
@@ -29,31 +29,54 @@ class TestModelRegistryCreation:
...
+        server, port = model_registry_instance_rest_endpoint.split(":")
+        registry = ModelRegistry(
+            server_address=f"{Protocols.HTTPS}://{server}",
+            port=port,
+            author="opendatahub-test",
+            user_token=current_client_token,
+            is_secure=False,
+        )
+        model = registry.register_model(
+            name=f"{MODEL_NAME}",
```
adolfo-ab, line 49: No need for fstring here, just pass the constant

tests/model_registry/test_model_registry_creation.py
```diff
@@ -29,31 +29,54 @@ class TestModelRegistryCreation:
...
+            uri="https://storage-place.my-company.com",
+            version="2.0.0",
+            description="lorem ipsum",
+            model_format_name="onnx",
```
adolfo-ab, line 53: Use ModelFormat.ONNX (already exists on constants.py)

tests/model_registry/test_model_registry_creation.py
```diff
@@ -29,31 +29,54 @@ class TestModelRegistryCreation:
...
+            metadata={
+                "int_key": 1,
+                "bool_key": False,
+                "float_key": 3.14,
+                "str_key": "str_value",
+            },
         )
-        _, out, _ = run_command(command=shlex.split(cmd))
-        out_dict = literal_eval(node_or_string=out)
+        registered_model = registry.get_registered_model(f"{MODEL_NAME}")
```
adolfo-ab, line 64: No need to use fstring here, just pass the constant

tests/model_registry/test_model_registry_creation.py
```diff
@@ -29,31 +29,54 @@ class TestModelRegistryCreation:
...
+        if not registered_model.name == model.name:
+            errors.append(f"Unexpected name, received {registered_model.name}")
+        if not registered_model.description == model.description:
+            errors.append(f"Unexpected description, received {registered_model.description}")
+        if not registered_model.owner == model.owner:
+            errors.append(f"Unexpected owner, received {registered_model.owner}")
+        if not registered_model.state == model.state:
+            errors.append(f"Unexpected state, received {registered_model.state}")
 
         assert not errors, "errors occured:\n{}".format("\n".join(errors))
```
adolfo-ab, line 77: Not a big deal, but maybe make the base error message a little bit more descriptive. E.g. "Errors found during model registry validation" or something similar instead of "errors occured"

tests/model_registry/test_model_registry_creation.py
adolfo-ab: I can see a point that testing the integration between ODH/RHOAI and the model registry python client can be part of the scope of this repo. Also, I think that a test _in general_ should only test one thing, and that we should try to avoid implicit testing when possible, so I think it makes sense to have a `model_registry` fixture for this test, and if you want to test the instantiation of a model registry maybe you can have another test specific for that

lugi0: I will send a new PR in the near future overhauling the python client tests, and will also address this point. For the time being I'd suggest merging this to not leave it limbo

rnetser (approved):

adolfo-ab (approved):

PR #126, opened 2025-02-06
ADD test for cross model auth and enable token and verify

mwaykole:
verified 
: test will fail as [RHOAIRFE-534](https://issues.redhat.com/browse/RHOAIRFE-534)


tests/model_serving/model_server/authentication/test_kserve_token_authentication_raw.py
```diff
@@ -45,3 +45,32 @@ def test_disabled_raw_model_authentication(self, patched_remove_raw_authenticati
...
+            inference_service=http_s3_caikit_raw_inference_service,
+            inference_config=CAIKIT_TGIS_INFERENCE_CONFIG,
+            inference_type=Inference.ALL_TOKENS,
+            protocol=Protocols.HTTPS,
+            model_name=ModelFormat.CAIKIT,
+            use_default_query=True,
+            token=http_raw_inference_token,
+        )
+
+    @pytest.mark.skip(reason="Skipping test_cross_model_authentication_raw: RHOAIRFE-534")
```
rnetser, line 62: because auth fails (but a wrong code) i think that we should not skip the test but run the inference and check here that we fail (i.e do not use `verify_inference_response` )

tests/model_serving/model_server/conftest.py
```diff
@@ -88,6 +88,17 @@ def model_service_account(
         yield sa
 
 
+@pytest.fixture(scope="class")
+def model_service_account_2(admin_client: DynamicClient, models_endpoint_s3_secret: Secret) -> ServiceAccount:
```
rnetser, line 92: this is used only in authentication, please move to conftest in that dir

rnetser (changes requested):

tests/model_serving/model_server/authentication/test_kserve_token_authentication_raw.py
mwaykole: (i.e do not use verify_inference_response )
then what should be used ?

tests/model_serving/model_server/conftest.py
mwaykole: ack

tests/model_serving/model_server/authentication/test_kserve_token_authentication_raw.py
rnetser: if bug is open, use `run_inference_flow` and check we get is an expected error and not something else (i.e another bug)
else call `verify_inference_response`

rnetser (changes requested):

tests/model_serving/model_server/authentication/test_kserve_token_authentication_raw.py
mwaykole: we have multiple test which test `expected error` is it really necessary?

tests/model_serving/model_server/authentication/test_kserve_token_authentication_raw.py
```diff
@@ -80,3 +82,47 @@ def test_raw_disable_enable_authentication_no_pod_rollout(self, http_s3_caikit_r
...
+            inference_service=http_s3_caikit_raw_inference_service,
+            inference_config=CAIKIT_TGIS_INFERENCE_CONFIG,
+            inference_type=Inference.ALL_TOKENS,
+            protocol=Protocols.HTTPS,
+            model_name=ModelFormat.CAIKIT,
+            use_default_query=True,
+            token=http_raw_inference_token,
+        )
+
+    # @pytest.mark.skip(reason="Skipping test_cross_model_authentication_raw: RHOAIENG-19645")
```
rnetser, line 99: leftover? should be removed, right?

mwaykole: ack

rnetser (approved):

nehachopra27 (approved):

PR #132, opened 2025-02-07
Add Stateful API testing for Model Registry

lugi0:
## Description

This adds a test for stateful API testing of Model Registry. It also bumps Schemathesis to V4, which is still an alpha release but appears to work fine for this workload. 
I will keep this as a draft until the openAPI spec is updated in the Model Registry repo to support Stateful testing, and we can then decide if we are ok with an alpha release being pulled or if we want to wait / downgrade to V3 (requires some rework due to breaking changes between the two major versions).
I have also added a new mark type, "fuzzer", which I'd use to indicate tests which are bound to fail due to their random nature. 

## How Has This Been Tested?

Multiple runs made from local with an updated openAPI spec file which have already identified some failures to address in Model Registry.

## Merge criteria:


lugi0:
On hold until https://github.com/kubeflow/model-registry/pull/773 is merged

tests/model_registry/conftest.py
```diff
@@ -298,8 +303,44 @@ def model_registry_instance_rest_endpoint(
...
     )
+    schema.configure(base_url=f"https://{model_registry_instance_rest_endpoint}/")
+    return schema
+
+
+# Used when testing local changes
+# @pytest.fixture(scope="class")
+# def generated_schema_local(model_registry_instance_rest_endpoint):
+#     schema = schemathesis.openapi.from_path(
+#         path="utilities/manifests/model-registry-api.yaml",
```
rnetser, line 318: please set path in py_config so when a user wants to overwrite they can us tc=<param name>:<path>

tests/model_registry/conftest.py
```diff
@@ -298,8 +303,44 @@ def model_registry_instance_rest_endpoint(
...
+        def setup(self) -> None:
+            self.headers = {"Authorization": f"Bearer {current_client_token}", "Content-Type": "application/json"}
+
+        # these kwargs are passed to requests.request()
+        def get_call_kwargs(self, case: Case) -> dict[str, Any]:
+            return {"verify": False, "headers": self.headers}
+
+        def after_call(self, response: Response, case: Case) -> None:
+            LOGGER.info(
+                "%s %s -> %d",
```
rnetser, line 340: please use f-string formatting

tests/model_registry/conftest.py
```diff
@@ -298,8 +303,44 @@ def model_registry_instance_rest_endpoint(
...
+# def generated_schema_local(model_registry_instance_rest_endpoint):
+#     schema = schemathesis.openapi.from_path(
+#         path="utilities/manifests/model-registry-api.yaml",
+#     )
+#     schema.configure(base_url=f"https://{model_registry_instance_rest_endpoint}/")
+#     return schema
+
+
+@pytest.fixture
+def state_machine(generated_schema: BaseOpenAPISchema, current_client_token: str) -> APIStateMachine:
```
rnetser, line 325: this can be a regular function; there's no setup / teardown related to the test here

tests/model_registry/test_rest_api.py
```diff
@@ -4,13 +4,11 @@
...
-schema = schemathesis.from_pytest_fixture("generated_schema")
+schema = schemathesis.pytest.from_fixture("generated_schema")
 
 
-# TODO: This is a Stateless test due to how the openAPI spec is currently defined in upstream.
-# Once it is updated to support Stateful testing of the API we can enable this to run every time,
-# but for now having it run manually to check the existing failures is more than enough.
-@pytest.mark.skip(reason="Only run manually for now")
+@pytest.mark.fuzzer
+@pytest.mark.sanity
```
rnetser, line 11: use `pytestmark` to set all markers in the same place.

tests/model_registry/test_rest_api.py
```diff
@@ -4,13 +4,11 @@
...
 
-schema = schemathesis.from_pytest_fixture("generated_schema")
+schema = schemathesis.pytest.from_fixture("generated_schema")
 
 
-# TODO: This is a Stateless test due to how the openAPI spec is currently defined in upstream.
-# Once it is updated to support Stateful testing of the API we can enable this to run every time,
-# but for now having it run manually to check the existing failures is more than enough.
-@pytest.mark.skip(reason="Only run manually for now")
+@pytest.mark.fuzzer
```
rnetser, line 10: i do not think that sanity tests should be marked as `fuzzer`; sanity tests are a set of tests to check basic functionality.  fuzzing (or in this case tools which are based on it) is a different level of testing

rnetser (changes requested):

tests/model_registry/conftest.py
lugi0: Schemathesis' documentation recommends using a fixture when loading the schema lazily: https://schemathesis.readthedocs.io/en/stable/stateful.html#schemathesis.stateful.state_machine.APIStateMachine.validate_response

tests/model_registry/conftest.py
lugi0: Will remove as not needed

tests/model_registry/conftest.py
lugi0: Not sure to which scenario this would apply, this needs to be manually enabled and pointing to an updated openapi spec, I don't think it makes sense to have it exposed as a config variable.

lugi0: Removed

lugi0:
The test now works as we've updated the openAPI spec file in upstream, however I sometimes get a library error (`hypothesis.errors.FlakyStrategyDefinition`) I'm still trying to figure out. Opening the PR up for review but keeping it on hold until I get to the root case of the error

tests/model_registry/test_rest_api.py
dbasunag: if this is part of sanity, then this test must be stable and should not be mark as `fuzzer`
```
Mark tests that use fuzzing and are probably going to generate unanticipated failures.
```

adolfo-ab (dismissed):

tests/model_registry/conftest.py
```diff
@@ -301,11 +306,33 @@ def model_registry_instance_rest_endpoint(
...
+        def setup(self) -> None:
+            self.headers = {"Authorization": f"Bearer {current_client_token}", "Content-Type": "application/json"}
+
+        # these kwargs are passed to requests.request()
+        def get_call_kwargs(self, case: Case) -> dict[str, Any]:
+            return {"verify": False, "headers": self.headers}
+
+        def after_call(self, response: Response, case: Case) -> None:
+            LOGGER.info(f"{case.method} {case.path} -> {response.status_code}")
+            # LOGGER.info("This is the after_call function")
```
rnetser, line 333: leftover?

lugi0: Removed

tests/model_registry/test_rest_api.py
lugi0: Removed the Sanity mark

rnetser (approved):

tests/model_registry/conftest.py
```diff
@@ -330,7 +330,6 @@ def get_call_kwargs(self, case: Case) -> dict[str, Any]:
 
         def after_call(self, response: Response, case: Case) -> None:
             LOGGER.info(f"{case.method} {case.path} -> {response.status_code}")
-            # LOGGER.info("This is the after_call function")
```
dbasunag: This seems like leftover from debugging. Can you please remove?

lugi0: It is already removed @dbasunag

dbasunag (dismissed):

tests/model_registry/conftest.py
```diff
@@ -44,11 +49,15 @@
...
-    ) as ns:
+    ns = Namespace(name=MR_NAMESPACE, client=admin_client)
+    if ns.exists:
         yield ns
+    else:
+        with create_ns(
+            name=request.param.get("namespace_name", MR_NAMESPACE),
+            admin_client=admin_client,
+        ) as ns:
+            yield ns
```
dbasunag, line 60: We should not do this.  Tests that creates namespace, must clean them up.

dbasunag (changes requested):

tests/model_registry/conftest.py
dbasunag: If the namespace is supposed to be created automatically, and did not, we should fail. However, we should add a check that it is in active status before we proceed. Clean up should not happen here though, it should be part of update dsc call (I added it there, since update dsc, enabling model registry creates the ns, it should be explicitly cleaning it up.

lugi0:
Removing the `hold` label since I am not able to replicate the `hypothesis.errors.FlakyStrategyDefinition` error anymore

tests/model_registry/conftest.py
dbasunag: This code would be changing in a follow up PR based on our discussions offline. Resolving.

dbasunag (approved):

uv.lock
mwaykole, line 1: Hey @lugi0, I think the uv.lock file might have been pushed unintentionally. Could you please revert that change when you get a chance? 😊

tests/model_registry/test_rest_api_stateful.py
```diff
@@ -0,0 +1,30 @@
...
+                    },
+                },
+            },
+        )
+    ],
+    indirect=True,
+)
+@pytest.mark.usefixtures("model_registry_namespace", "updated_dsc_component_state_scope_class")
+class TestRestAPIStateful:
+    def test_mr_api_stateful(self, state_machine):
```
mwaykole, line 29: can u please add test doc may be one liner ?

uv.lock
dbasunag: schemathesis version updated, so lock file would be updated as well. This is expected behavior.

tests/model_registry/test_rest_api_stateful.py
dbasunag: Let's take it as it is and add the doc in a follow up pr.

dbasunag (approved):

dbasunag (approved):

fege (approved):
/lgtm

PR #135, opened 2025-02-10
Add timeout parameter as optional for isvc creation and it corresponding change 

tarukumar:
I observed that when testing a larger model, which takes more time to download, the default timeout parameter isn't sufficient, causing the model deployment to fail. As a result, I modified the timeout parameter to be an option in the inference function, allowing for easier control.

### DIFF
--- a/tests/model_serving/model_server/utils.py
+++ b/tests/model_serving/model_server/utils.py
@@ -32,7 +32,9 @@ from utilities.jira import is_jira_open
 LOGGER = get_logger(name=__name__)
 
 
-def verify_no_failed_pods(client: DynamicClient, isvc: InferenceService, runtime_name: str | None) -> None:
+def verify_no_failed_pods(
+    client: DynamicClient, isvc: InferenceService, runtime_name: str | None, timeout: int = 5 * 60
+) -> None:
     """
     Verify no failed pods.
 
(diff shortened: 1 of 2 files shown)


tests/model_serving/model_server/utils.py
```diff
@@ -34,22 +33,25 @@
 LOGGER = get_logger(name=__name__)
 
 
-def verify_no_failed_pods(client: DynamicClient, isvc: InferenceService, runtime_name: str | None) -> None:
+def verify_no_failed_pods(
+    client: DynamicClient, isvc: InferenceService, runtime_name: str | None, timeout: int = 5 * 60
```
rnetser, line 37: please use `TIMEOUT_5MIN` from constants

utilities/infra.py
```diff
@@ -88,6 +88,7 @@ def wait_for_inference_deployment_replicas(
     isvc: InferenceService,
     runtime_name: str | None,
     expected_num_deployments: int = 1,
+    timeout: int = 4 * 60,
```
rnetser, line 91: please save to constants and re-use (also in `create_ns`)

rnetser (changes requested):

tests/model_serving/model_server/utils.py
adolfo-ab: +1

utilities/infra.py
adolfo-ab: +1

tests/model_serving/model_server/utils.py
```diff
@@ -125,6 +127,7 @@ def create_isvc(
     wait_for_predictor_pods: bool = True,
     autoscaler_mode: Optional[str] = None,
     multi_node_worker_spec: Optional[dict[str, int]] = None,
+    timeout: int = 15 * 60,
```
adolfo-ab, line 130: use constant

PR #141, opened 2025-02-12
[runtime]Add support vllm tool calling

tarukumar:
validated with 2.18


tests/model_serving/model_runtime/vllm/toolcalling/test_granite_3_2_8b_instruct_preview.py
```diff
@@ -0,0 +1,201 @@
...
+            },
+        ),
+    ],
+    indirect=True,
+)
+class TestGranite32ToolModel:
+    def test_granite_model_inference(
+        self,
+        vllm_inference_service: Generator[InferenceService, Any, Any],
+        get_pod_name_resource: Pod,
```
rnetser, line 61: `get_pod_name_resource` was not added in the pr but should be renamed, a fixture name should be a noun (in this case something like `vllm_pod` ) and not a verb

tests/model_serving/model_runtime/vllm/toolcalling/test_granite_3_2_8b_instruct_preview.py
```diff
@@ -0,0 +1,201 @@
...
+        )
+        validate_inference_output(
+            model_info,
+            chat_responses,
+            completion_responses,
+            model_info_tgis,
+            completion_responses_tgis,
+            completion_responses_tgis_stream,
+            response_snapshot=response_snapshot,
+        )
```
rnetser, line 86: duplicate code in other tests, please extract and re-use

tests/model_serving/model_runtime/vllm/toolcalling/test_granite_3_2_8b_instruct_preview.py
```diff
@@ -0,0 +1,201 @@
...
+        vllm_inference_service: Generator[InferenceService, Any, Any],
+        get_pod_name_resource: Pod,
+        response_snapshot: Any,
+    ):
+        pod_name = get_pod_name_resource.name
+        model_info_w, chat_responses_w, completion_responses_w = run_raw_inference(
+            pod_name=pod_name,
+            isvc=vllm_inference_service,
+            port=8080,
+            endpoint=OPENAI_ENDPOINT_NAME,
```
rnetser, line 99: save as base kwargs and re-use

tests/model_serving/model_runtime/vllm/toolcalling/test_granite_3_2_8b_instruct_preview.py
```diff
@@ -0,0 +1,201 @@
...
+        )
+
+    def test_granite_model_tool_inference(
+        self,
+        vllm_inference_service: Generator[InferenceService, Any, Any],
+        get_pod_name_resource: Pod,
+        response_snapshot: Any,
+    ):
+        pod_name = get_pod_name_resource.name
+        model_info_w, chat_responses_w, completion_responses_w = run_raw_inference(
```
rnetser, line 95: what does `l` and `w` stand for?

tests/model_serving/model_runtime/vllm/toolcalling/test_granite_3_2_8b_instruct_preview.py
```diff
@@ -0,0 +1,201 @@
...
+        pytest.param(
+            {"name": "granite-32-8b-multi"},
+            {"model-dir": MODEL_PATH},
+            {"deployment_type": KServeDeploymentType.RAW_DEPLOYMENT},
+            {
+                "deployment_mode": KServeDeploymentType.RAW_DEPLOYMENT,
+                "runtime_argument": SERVING_ARGUMENT,
+                "gpu_count": 2,
+                "name": "granite-32-8b-multi",
+                "min-replicas": 1,
```
rnetser, line 134: beside gpu_count - save and re-use

tests/model_serving/model_runtime/vllm/toolcalling/test_granite_3_2_8b_instruct_preview.py
```diff
@@ -0,0 +1,201 @@
...
+        )
+        validate_inference_output(
+            model_info_w,
+            chat_responses_w,
+            completion_responses_w,
+            model_info_l,
+            chat_responses_l,
+            completion_responses_l,
+            response_snapshot=response_snapshot,
+        )
```
rnetser, line 201: duplicate code, please extract and re-use

tests/model_serving/model_runtime/vllm/utils.py
```diff
@@ -51,13 +51,16 @@ def fetch_openai_response(  # type: ignore
     model_name: str,
     chat_query=CHAT_QUERY,
     completion_query=COMPLETION_QUERY,
+    tool_calling: Optional[dict[Any, Any]] = None,
```
rnetser, line 54: please use  union (X | None)  https://docs.python.org/3/library/typing.html#typing.Optional

rnetser (changes requested):

rnetser:
@tarukumar   `validated with 2.18` - will it work on 2.17? if not, we should wait with merging will we branch 2.17

tests/model_serving/model_runtime/vllm/toolcalling/test_granite_3_2_8b_instruct_preview.py
tarukumar: I've opened an issue for this in the repo. Once this PR is merged, I will make the requested changes for all the vLLM tests in single PR

tests/model_serving/model_runtime/vllm/toolcalling/test_granite_3_2_8b_instruct_preview.py
tarukumar: Could you clarify what you mean by "duplicate code"? This function takes the output queries from the model's response and compares them with a variable for the model's response. While the response variable might be the same, the output differs. When you refer to "duplicate," what exactly do you mean? The purpose of this function is to handle multiple responses and match them accordingly. Also keeping here make it easy for the folks to understand better

tests/model_serving/model_runtime/vllm/toolcalling/test_granite_3_2_8b_instruct_preview.py
tarukumar: Could you clarify what you mean by "save and reuse"? Are you suggesting keeping the other parameters in a variable statically and only adding the GPU count and name here because those two are distinct?

tests/model_serving/model_runtime/vllm/toolcalling/test_granite_3_2_8b_instruct_preview.py
tarukumar: Could you clarify what you mean by "duplicate code"? This function takes the output queries from the model's response and compares them with a variable for the model's response. While the response variable might be the same, the output differs. When you refer to "duplicate," what exactly do you mean? The purpose of this function is to handle multiple responses and match them accordingly. Also keeping here make it easy for the folks to understand better

tests/model_serving/model_runtime/vllm/utils.py
tarukumar: Aside from personal preference, what benefits do you see? I’ve tried using both options, but it didn’t seem to make a difference in the test itself. If you’ve noticed any impact with suggested changes, I’d like to know.

tarukumar:
> @tarukumar `validated with 2.18` - will it work on 2.17? if not, we should wait with merging will we branch 2.17

Validated with 2.17 image as well and it seems to be working fine

tests/model_serving/model_runtime/vllm/toolcalling/test_granite_3_2_8b_instruct_preview.py
rnetser: `run_raw_inference`, `run_raw_inference` and `validate_inference_output` - should be wrapped in a function with a meaningful name that does it all.
this way, there's no need to call all 3 each time and if there's a change. it is done in one place

tests/model_serving/model_runtime/vllm/toolcalling/test_granite_3_2_8b_instruct_preview.py
rnetser: so please give meaningful names

tests/model_serving/model_runtime/vllm/toolcalling/test_granite_3_2_8b_instruct_preview.py
rnetser: to avoid duplication (and change in one place, if needed), you can save a base dict

tests/model_serving/model_runtime/vllm/toolcalling/test_granite_3_2_8b_instruct_preview.py
rnetser: please see comment above

tests/model_serving/model_runtime/vllm/utils.py
rnetser: https://docs.python.org/3/library/stdtypes.html#union-type

"The union type expression enables cleaner type hinting syntax "

rnetser (changes requested):

adolfo-ab (commented):
Will approve once @rnetser comments are addressed

tarukumar:
@rnetser i have addressed the comment please take a look

lugi0 (approved):

tests/model_serving/model_runtime/vllm/utils.py
```diff
@@ -132,3 +142,49 @@ def validate_supported_quantization_schema(q_type: str) -> None:
...
     for data in args:
         assert data == response_snapshot, f"output mismatch for {data}"
+
+
+def validate_raw_openai_inference_request(
+    pod_name: str,
+    isvc: InferenceService,
+    response_snapshot: Any,
+    chat_query: list[list[dict[str, str]]] = CHAT_QUERY,
+    completion_query: list[dict[str, str]] = COMPLETION_QUERY,
```
rnetser, line 152: using mutable default values ​​as default arguments is not recommended.

rnetser (changes requested):

tests/model_serving/model_runtime/vllm/utils.py
tarukumar: I have  have taken suggestion  into consideration. However, the suggestion doesn't make a difference in this case. It’s not recommended when we play around with mutable object and modify it  but in our code, we’re simply passing the value around without performing any  changes/operation. I understand the ambiguity around using a default parameter with a mutable object, but technically, that doesn’t apply here.

mwaykole (approved):

PR #143, opened 2025-02-12
[model server] verify product version in model mesh container

rnetser:
## Merge criteria:


tests/model_serving/model_server/model_mesh/test_model_mesh_multi_models.py
```diff
@@ -1,4 +1,7 @@
+import re
+
 import pytest
+from packaging.version import Version
```
lugi0, line 4: This only works if the version being tested is valid according to PEP 440, I would suggest using SemVer for handling more general versioning styles

tests/model_serving/model_server/model_mesh/test_model_mesh_multi_models.py
```diff
@@ -58,3 +62,25 @@ def test_model_mesh_tensorflow_with_openvino(
...
+        http_s3_openvino_second_model_mesh_inference_service,
+    ):
+        """Verify model mesh container contains the correct product version"""
+        pod = get_pods_by_isvc_label(
+            client=http_s3_openvino_model_mesh_inference_service.client,
+            isvc=http_s3_openvino_model_mesh_inference_service,
+            runtime_name=http_s3_ovms_model_mesh_serving_runtime.name,
+        )[0]
+
+        mm_log = pod.log(container="mm")
```
lugi0, line 80: is it possible you get into timing issues here and the logs are still empty or the versioning string still hasn't been printed?

tests/model_serving/model_server/model_mesh/test_model_mesh_multi_models.py
```diff
@@ -58,3 +62,25 @@ def test_model_mesh_tensorflow_with_openvino(
...
+
+        if version_match := re.search(r"Registering ModelMesh Service version as \"v(\d\.\d+\.\d+)", mm_log):
+            assert Version(version_match.group(1)) == get_product_version(admin_client=admin_client)
```
lugi0, line 83: Are you sure MM version matches RHOAI/ODH version?

tests/model_serving/model_server/model_mesh/test_model_mesh_multi_models.py
rnetser: we get here after the isvc is ready so pod is already running

tests/model_serving/model_server/model_mesh/test_model_mesh_multi_models.py
rnetser: yes

tests/model_serving/model_server/model_mesh/test_model_mesh_multi_models.py
rnetser: done

tests/model_serving/model_server/model_mesh/test_model_mesh_multi_models.py
```diff
@@ -58,3 +62,25 @@ def test_model_mesh_tensorflow_with_openvino(
             protocol=Protocols.HTTP,
             use_default_query=True,
         )
+
+    def test_product_version_in_model_mesh_container(
+        self,
+        admin_client,
+        http_s3_ovms_model_mesh_serving_runtime,
+        http_s3_openvino_model_mesh_inference_service,
+        http_s3_openvino_second_model_mesh_inference_service,
```
adolfo-ab, line 71: Why do you need the second model? As far as I can see it's not used within this test, right?

tests/model_serving/model_server/model_mesh/test_model_mesh_multi_models.py
adolfo-ab, line 80: Not a blocker but `mm_logs` is only used once, so no need to store in a variable unless you consider the code is easier to read by doing it

tests/model_serving/model_server/model_mesh/test_model_mesh_multi_models.py
lugi0: +1

tests/model_serving/model_server/model_mesh/test_model_mesh_multi_models.py
lugi0: Would it make sense to dump the logs when raising the ValueError exception?

tests/model_serving/model_server/model_mesh/test_model_mesh_multi_models.py
rnetser: as this was parametrized, needed to call them but yes, it made no sense so moved to another test with one model and moved it to smoke

tests/model_serving/model_server/model_mesh/test_model_mesh_multi_models.py
rnetser: added part of the log to the assert

lugi0 (approved):

PR #151, opened 2025-02-19
[model server] hard-code multi-node servingRuntime name

rnetser:
because of https://github.com/opendatahub-io/odh-model-controller/blob/incubating/internal/controller/serving/servingruntime_controller.go#L407-L412; servingruntime name for multi-node must be hard-coded to `vllm-multinode-runtime`
## Merge criteria:


lugi0 (approved):
LGTM, would it be possible to use the Jira plugin to remind us of the TODO?

tarukumar (approved):
/lgtm

rnetser:
> LGTM, would it be possible to use the Jira plugin to remind us of the TODO?

for this specific use case i prefer not to use it as it uses a generic runtime fixture and i do not want to call jira for just this use case

PR #153, opened 2025-02-19
Approve a PR by leaving "/lgtm" comment

lugi0:
## Description

This add a GHA to approve a PR when a user leaves a "/lgtm" comment

## Merge criteria:


lugi0 (commented):
testing /lgtm

.github/workflows/add-remove-labels.yml
```diff
@@ -3,7 +3,8 @@ on:
   pull_request_target:
     types: [synchronize]
 
-  pull_request_review:
+  pull_request_review_comment:
+    types: [created]
```
rnetser, line 7: we also need to address it when coming from `issue_comment`; needs to be done in the python code

rnetser (changes requested):

.github/workflows/add-remove-labels.yml
lugi0: `issue_comment` is handled by #L9, isn't this breaking only when leaving a comment on the diff review?

lugi0:
see https://github.com/lugi0/opendatahub-tests/pull/2 for testing of the workflow

.github/workflows/add-remove-labels.yml
rnetser: missed that this is another workflow

.github/workflows/lgtm-approval.yml
```diff
@@ -0,0 +1,22 @@
+name: Approve PR on /lgtm comment
+
+on:
+  workflow_dispatch: #manual testing
+  issue_comment:
+    types: [created]
```
rnetser, line 6: please add condition to trigger only if "lgtm in comment"
e/g
```
    if: |
      contains(github.event.comment.body, '/lgtm')
```

also should be triggered on `edited`

lugi0: The actions already checks if /lgtm is int he comment: https://github.com/opendatahub-io/opendatahub-tests/pull/153/files#diff-8adeafbcbe79cc05a3ae753fd95a88451697a3356e53b468f8a7f565c5c737d1R13
Why would we edit a comment to add a lgtm? that seems unlikely to be honest

rnetser: 1. i think adding the check on the content of the comment at the top makes it more clear
2. a user can edit a comment and add "lgtm"

lugi0: I've added the edit event, I still think it is not a likely scenario though :) 
The `if` conditional is not supported on the definition of the event as far as I can see.

lugi0:
From my testing, now all of the expected functionality of the two actions is present: https://github.com/lugi0/opendatahub-tests/pull/2
The only bit I haven't been able to figure out is how to leave an emoji reaction to a review's main comment, but I think it is a nice to have that can be fixed down the line

.github/workflows/lgtm-approval.yml
```diff
@@ -0,0 +1,34 @@
...
+  pull_request_review_comment:
+    types: [created, edited]
+
+jobs:
+  approve-pr:
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+    if: >
+      (github.event.issue.pull_request && contains(github.event.comment.body, '/lgtm')) ||
```
rnetser, line 17: what if i use `/lgtm cancel` or delete my comment? add and remove pr approval should be done in our python code

.github/workflows/scripts/pr_workflow.py
```diff
@@ -88,6 +88,8 @@ def __init__(self) -> None:
         self.user_login = os.getenv("GITHUB_USER_LOGIN")
         self.review_state = os.getenv("GITHUB_EVENT_REVIEW_STATE")
         self.comment_body = os.getenv("COMMENT_BODY", "")
+        if self.comment_body == "":
```
rnetser, line 91: please split; each action should have its relevant attributes and checks. 
so here, for example add `self.review_comment_body` etc

this helps to understand what is needed for each action, and it is easier to maintain different flows.
it is ok for  some flows share some things in common

rnetser (changes requested):

.github/workflows/lgtm-approval.yml
lugi0: Done in python code, now both options work

.github/workflows/scripts/pr_workflow.py
lugi0: This was discussed further by me and Ruth; we agreed to keep overriding this variable but adding a comment explaining what's happening. Done now

.github/workflows/add-remove-labels.yml
```diff
@@ -31,19 +36,42 @@ jobs:
...
+
       - uses: actions/checkout@v4
 
       - name: Install uv
         uses: astral-sh/setup-uv@v5
 
       - name: Run add remove labels
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          # GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
rnetser, line 67: please remove if not used

.github/workflows/scripts/pr_workflow.py
```diff
@@ -297,6 +305,19 @@ def issue_comment_label_actions(
     def add_welcome_comment(self) -> None:
         self.pr.create_issue_comment(body=WELCOME_COMMENT)
 
+    def approve_pr(self) -> None:
+        self.pr.create_review(event="APPROVE")
+
+    def dismiss_pr_approval(self) -> None:
+        all_reviews = self.pr.get_reviews()
+        current_user = self.gh_client.get_user().login
```
rnetser, line 313: isn't this `self.user_login`?

.github/workflows/scripts/pr_workflow.py
```diff
@@ -282,12 +288,14 @@ def issue_comment_label_actions(
...
 
                 if action[CANCEL_ACTION] or self.event_action == "deleted":
+                    self.dismiss_pr_approval()
                     if label_in_pr:
                         LOGGER.info(f"Removing label {label}")
                         self.pr.remove_from_labels(label=label)
 
                 elif not label_in_pr:
                     self.add_pr_label(label=label)
+                    self.approve_pr()
```
rnetser, line 298: wouldn't this approave the pr for all user labels?
pr should be approved only if self.user_login != pr owner

rnetser (changes requested):

.github/workflows/scripts/pr_workflow.py
lugi0: No, the approval through the action is left by the bot, so we need to dismiss the bot's review.
We could expand it to also remove for the user who has left the comment if an approval by them exists, but we run into the issue that we don't have a token for each user and we won't be able to dismiss it using the bot's token

.github/workflows/scripts/pr_workflow.py
lugi0: >wouldn't this approave the pr for all user labels?
shoot, yeah, I was sure it was indented under the first `if` block, will fix tomorrow

>pr should be approved only if self.user_login != pr owner
let's discuss about this, I think there are more scenarios to consider

.github/workflows/add-remove-labels.yml
lugi0: if we agree on using the bot account rather than the actions bot yeah, I can remove it

lugi0: Removed

.github/workflows/scripts/pr_workflow.py
lugi0: Checking for team membership now. I've done a workaround until we can get the bot into the org but the logic is there.

adolfo-ab (dismissed):

.github/workflows/add-remove-labels.yml
```diff
@@ -31,19 +36,41 @@ jobs:
           comment-id: ${{ github.event.comment.id }}
           reactions: '+1'
 
+      # This currently fails with either the bot PAT or the standard github token secret
```
dbasunag, line 39: Do you want to remove these?

.github/workflows/scripts/pr_workflow.py
```diff
@@ -77,22 +78,45 @@ def verify_base_config(self) -> None:
...
+        users = [member.login for member in members]
+        # TODO: replace once bot user is part of the org and team
+        # users = ["lugi0", "rnetser", "adolfo-ab", "tarukumar", "dbasunag", "mwaykole"]
+        return users
+
+    def verify_allowed_user(self) -> None:
+        allowed_users = self.get_allowed_users()
+        if self.user_login not in allowed_users:
+            LOGGER.info(f"User {self.user_login} is not allowed for this action. Exiting.")
+            sys.exit(0)
```
dbasunag, line 117: Should you exit or add a comment? Exiting would just fail the action.

.github/workflows/add-remove-labels.yml
lugi0: Not really? I'm hoping this actually gets fixed (https://github.com/github/docs/issues/36899) and we can begin using it

.github/workflows/scripts/pr_workflow.py
dbasunag: In that case how would the user know that the action was not allowed? If the workflow passed, no one is going to check for the logger message. May be you should consider adding a comment?

lugi0: why should the user know they were not allowed? I think it would be worse to see a failure every time an unallowed user tried to launch the action to be honest

mwaykole (dismissed):

.github/workflows/scripts/pr_workflow.py
```diff
@@ -101,11 +125,11 @@ def verify_labeler_config(self) -> None:
             if not self.user_login:
                 sys.exit("`GITHUB_USER_LOGIN` is not set")
 
-            if self.event_name == "issue_comment" and not self.comment_body:
-                sys.exit("`COMMENT_BODY` is not set")
-
-            if self.event_name == "pull_request_review" and not self.review_state:
-                sys.exit("`GITHUB_EVENT_REVIEW_STATE` is not set")
+            if (
+                self.event_name == "issue_comment" or self.event_name == "pull_request_review"
```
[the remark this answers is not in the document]
rnetser: why change the logic? the flow needs to fail  if:
            if self.event_name == "issue_comment" and not self.comment_body:
                sys.exit("`COMMENT_BODY` is not set")

of
            if self.event_name == "pull_request_review" and not self.review_state:
                sys.exit("`GITHUB_EVENT_REVIEW_STATE` is not set")

.github/workflows/scripts/pr_workflow.py
```diff
@@ -276,12 +301,20 @@ def issue_comment_label_actions(
             LOGGER.info(f"Processing labels: {labels}")
             for label, action in labels.items():
                 if label == LGTM_LABEL_STR:
-                    label = f"{LGTM_BY_LABEL_PREFIX}{self.user_login}"
+                    if self.user_login == self.pr.user.login:
+                        LOGGER.info("PR submitter cannot approve for their own PR")
```
rnetser, line 305:
```suggestion
                        LOGGER.error("PR submitter cannot approve for their own PR")
```

.github/workflows/scripts/pr_workflow.py
```diff
@@ -276,12 +301,20 @@ def issue_comment_label_actions(
...
+                        continue
+                    else:
+                        label = f"{LGTM_BY_LABEL_PREFIX}{self.user_login}"
+                        if not action[CANCEL_ACTION] or self.event_action == "deleted":
+                            self.approve_pr()
```
rnetser, line 310: should move to be part of the `if action[CANCEL_ACTION] or self.event_action == "deleted":` block

adolfo-ab (dismissed):

.github/workflows/scripts/pr_workflow.py
```diff
@@ -297,6 +330,19 @@ def issue_comment_label_actions(
...
+    def approve_pr(self) -> None:
+        self.pr.create_review(event="APPROVE")
+
+    def dismiss_pr_approval(self) -> None:
+        all_reviews = self.pr.get_reviews()
+        current_user = self.gh_client.get_user().login
+        LOGGER.info(f"Looking for approving review by user {current_user}")
+        # The reviews are paginated in chronological order. We need to get the newest by our account
+        for review in all_reviews.reversed:
+            if review.user.login == current_user and review.state == "APPROVED":
```
rnetser, line 342: please save `APPROVED` as constant and re-use

rnetser (changes requested):

.github/workflows/scripts/pr_workflow.py
```diff
@@ -77,22 +78,45 @@ def verify_base_config(self) -> None:
...
+        members: PaginatedList[github.NamedUser] = team.get_members()
+        users = [member.login for member in members]
+        # TODO: replace once bot user is part of the org and team
+        # users = ["lugi0", "rnetser", "adolfo-ab", "tarukumar", "dbasunag", "mwaykole"]
+        return users
+
+    def verify_allowed_user(self) -> None:
+        allowed_users = self.get_allowed_users()
+        if self.user_login not in allowed_users:
+            LOGGER.info(f"User {self.user_login} is not allowed for this action. Exiting.")
```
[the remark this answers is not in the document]
lugi0: This is not an error imho, it's just a user who commented and is not in the allowed list for this action to trigger. You'd see a failed github action every time a user not part of the admin group comments otherwise

.github/workflows/scripts/pr_workflow.py
lugi0: see comment above, I don't believe this is an error

.github/workflows/scripts/pr_workflow.py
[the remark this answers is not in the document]
lugi0: we already override comment body, to me it makes more sense to have a single line of log here

.github/workflows/scripts/pr_workflow.py
lugi0: See first comment, IMHO this should not be treated as an error, it is information but the action itself ran fine

.github/workflows/scripts/pr_workflow.py
lugi0: wouldn't work, this is checking the opposite (`if not`)

.github/workflows/scripts/pr_workflow.py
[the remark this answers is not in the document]
rnetser: it should start out in the logs; it is an error or at least a warning

.github/workflows/scripts/pr_workflow.py
rnetser: yes; error flows should exit on should exit(1). unauthorized user is an error flow

.github/workflows/scripts/pr_workflow.py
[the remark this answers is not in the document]
rnetser: you can come from `pull_request_review` and not have a body, right?

these are 2 different flows - one is an error and the other is not.

why was
`sys.exit("`GITHUB_EVENT_REVIEW_STATE` is not set")` removed? it is still in the code

.github/workflows/scripts/pr_workflow.py
rnetser: but this is an error message; it is not about the GH action it is about a user not being allowed to approve

.github/workflows/scripts/pr_workflow.py
rnetser: exactly, it is part of the same block (not condition)

.github/workflows/scripts/pr_workflow.py
rnetser: please address

rnetser (changes requested):

.github/workflows/scripts/pr_workflow.py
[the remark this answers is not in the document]
lugi0: the body of pull_request_review would still be in comment_body because we override it in the init if the "discussion page" body was empty

.github/workflows/scripts/pr_workflow.py
lugi0: I disagree, it is informational, the user is not allowed to kick off our internal automation but there's no failure in that. If someone from a different community commented on our PRs I wouldn't call that an error, we just don't want to grant them permission to perform the automated steps of the action.

.github/workflows/scripts/pr_workflow.py
lugi0: It is not the same condition, if you expand this whole block of logic you're checking that
```
if label == LGTM_LABEL_STR and  self.user_login != self.pr.user.login and  not (action[CANCEL_ACTION] or self.event_action == "deleted")
```
The other block is 
```
action[CANCEL_ACTION] or self.event_action == "deleted"
```

Even if you were to implement it in the else of the second block (not(action...)), you'd still be missing the condition on label and user. Why replicate the lgtm/user condition in the action block when we already have it in the first? The (second) action block is a catch-all for removing labels in case we are asking for delete/cancel, while the first block is already implementing the positive logic of adding the lgtm label, so to me it makes the most logical sense to perform the approval in the positive scenario

rnetser:
```
                if action[CANCEL_ACTION] or self.event_action == "deleted":
                    if label == LGTM_LABEL_STR:
                        self.dismiss_pr_approval()
                    if label_in_pr:
                        LOGGER.info(f"Removing label {label}")
                        self.pr.remove_from_labels(label=label)
                else:
                      if LGTM_LABEL_STR in label:
                               self.approve_pr()
                       if not label_in_pr:
                                 self.add_pr_label(label=label)
```

rnetser: this way all the logic is done in one place and it is clear what is the flow

.github/workflows/scripts/pr_workflow.py
lugi0: I completely disagree, I don't see the benefit of having a failed github action in a PR if a user left a comment in the PR, especially since the action did not fail at all but simply did not proceed to perform further actions given our own permission enforcement.

.github/workflows/scripts/pr_workflow.py
lugi0: See prior comment, completely disagree

.github/workflows/scripts/pr_workflow.py
lugi0: I find my own formatting better and more readable. Since it is just a disagreement on formatting I believe we won't come to an agreement and I don't see the point on discussing it further given the logic is correct.

dbasunag (approved):

PR #159, opened 2025-02-25
[docs[ Add style guide and enhance contibution guides

rnetser:
## Merge criteria:


docs/DEVELOPER_GUIDE.md
```diff
@@ -0,0 +1,135 @@
...
+- PRs that are not ready for review (but needed to be pushed for any reason) should have [WIP] in the title and labelled as "wip".
+  - When a PR is ready for review, remove the [WIP] from the title and remove the "wip" label.
+- PRs should be relatively small; if needed the PRs should be split and depended on each other.
+  - Small PRs will get quicker review.
+  - Small PRs comments will be fixed quicker and merged quicker.
+  - Both the reviewer and the committer will benefit from this.
+- When a refactor is needed as part of the current PR, the refactor should be done in another PR and the current PR should be rebased on it.
+- Please address each comment in code review
+  - If a comment was addressed and accepted, please comment as done and resolve.
+  - If a comment was addressed and rejected or additional discussion is needed, add your input and do not resolve the comment.
```
adolfo-ab, line 34: I think we need to be a bit more flexible with this. For example:
- A reviewer approves a PR, but leaves a minor comment that is not a blocker.
- A PR gets 2 approvals, while still having open comments from a third reviewer.
- A reviewer leaves a comment just before going on holidays.
- A reviewer leaves a comment that doesn't make sense, and is not willing to discuss it.
- etc

In these and other similar scenarios, IMO the committer or a designated maintainer should be able to resolve open comments and merge the PR. That's why I advocate for having at least one designated maintainer, to decide when it makes sense to "bypass" these rules, or when it makes sense to resolve a comment that don't make sense, while still ensuring high code quality. Obviously we should try to communicate and resolve comments the right way, but I feel that we should try to strike a balance where we have proper reviews while not causing unnecessary frustration and without causing PRs to sit for weeks when they could have been merged in days.

docs/STYLE_GUIDE.md
```diff
@@ -0,0 +1,19 @@
+## General
+
+- Remember [The Zen of Python](https://www.python.org/dev/peps/pep-0020/)
+- The repository styleguide is based on the [Google Python Style Guide](https://google.github.io/styleguide/pyguide.html).
+- The repository uses [pre-commit](https://pre-commit.com/) to enforce the styleguide.
+
+## Naming
+- Use descriptive names for tests, variables, functions, classes, etc.
```
adolfo-ab, line 8: - Avoid using single letters for variable names

docs/STYLE_GUIDE.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Documentation
+- Use [Google-format](https://google.github.io/styleguide/pyguide.html#381-docstrings) for docstrings.
+- Add docstrings to document functions, classes, and modules.
+- Avoid inline comments; the code should be self-explanatory.
```
adolfo-ab, line 13: Although I kind of agree with this one, I would rephrase it to something like: Write self-explanatory code that can be easily understood, and only use inline comments to supplement it when really necessary.

docs/DEVELOPER_GUIDE.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Conftest
+- Top level [conftest.py](../conftest.py) contains pytest native fixtures.
+- General tests [conftest.py](../tests/conftest.py) contains fixtures that are used in multiple tests by multiple teams.
+- If needed, create new `conftest.py` files in the relevant directories.
+
+
+## Fixtures
+- Ordering: Always call pytest native fixtures first, then session-scoped fixtures and then any other fixtures.
+- Fixtures should handle the setup and the teardown.
```
adolfo-ab, line 70: Maybe something like "Fixtures should handle their own setup and teardown, including any resources they create".

docs/DEVELOPER_GUIDE.md
```diff
@@ -0,0 +1,135 @@
...
+- Pytest reports failures in fixtures as ERROR
+- A fixture name should be a noun that describes what the fixture produces, rather than a verb.
```
adolfo-ab, line 72: We should try to explain why a fixture name should be a noun and not a verb. Maybe something like: "A fixture represents an object that has been created or modified to be used in a test, and therefore, a fixture name should be a noun that describes what the fixture produces, rather than a verb."

docs/DEVELOPER_GUIDE.md
```diff
@@ -0,0 +1,135 @@
...
+- If needed, create new `conftest.py` files in the relevant directories.
+
+
+## Fixtures
+- Ordering: Always call pytest native fixtures first, then session-scoped fixtures and then any other fixtures.
+- Fixtures should handle the setup and the teardown.
+- Pytest reports failures in fixtures as ERROR
+- A fixture name should be a noun that describes what the fixture produces, rather than a verb.  
+For example: If a test needs a storage secret, the fixture should be called 'storage_secret' and not 'create_secret'.
+- Fixtures should do one thing only.
```
adolfo-ab, line 74: This is a tricky one, and strictly speaking sometimes even our own fixtures do more than one thing. I would say something like: "Fixtures should have a single, clear responsibility."

docs/DEVELOPER_GUIDE.md
```diff
@@ -0,0 +1,135 @@
+# Project Structure
```
adolfo-ab, line 1: I think that either in `DEVELOPER_GUIDE.md` or in `README.md` we should include a brief section mentioning the wrapper with a link to its repo, since it's the main library that powers this testing suite besides pytest itself.

docs/STYLE_GUIDE.md
```diff
@@ -0,0 +1,19 @@
...
+## Naming
+- Use descriptive names for tests, variables, functions, classes, etc.
+
+## Documentation
+- Use [Google-format](https://google.github.io/styleguide/pyguide.html#381-docstrings) for docstrings.
+- Add docstrings to document functions, classes, and modules.
+- Avoid inline comments; the code should be self-explanatory.  
+Add comments only when necessary. For example, when using complex regex.
+
+## Typing
```
adolfo-ab, line 16: Should we mention using union operator `|` as opposed to `Optional`?

docs/DEVELOPER_GUIDE.md
rnetser: > at least one designated maintainer 
i completely agree
my concern is that to have quick merges these rules be skipped.
can we try it this way and if this does not work, we will revisit?

docs/DEVELOPER_GUIDE.md
rnetser: see new phrasing

docs/DEVELOPER_GUIDE.md
rnetser: updated, let me know what you think

docs/DEVELOPER_GUIDE.md
rnetser: not sure what "Fixtures should have a single, clear responsibility." means. 
this could end up with  a fixture creating all resources

docs/STYLE_GUIDE.md
rnetser: done

docs/STYLE_GUIDE.md
rnetser: the doc is pointing to https://docs.python.org/3/library/typing.html so i would refrain from mentioning specific typing, wdyt?

docs/DEVELOPER_GUIDE.md
adolfo-ab: Yes, I think that is a good approach, we can always change if it doesn't work

docs/DEVELOPER_GUIDE.md
adolfo-ab: Then maybe we can give a couple examples of good fixtures vs bad fixtures to make sure everyone understands

adolfo-ab (commented):
Just one little comment

docs/DEVELOPER_GUIDE.md
rnetser: done

lugi0: I think we could expand this to something like 
```
- Fixtures should be as limited as possible in their scope. A fixture should ideally handle only one responsibility.

### Example

What not to do:
@pytest.fixture()
def setup_openshift_app():
    my_app_cm = Configmap()
    my_app_deployment = Deployment()
    my_app_service = Service()
    return my_app_cm,my_app_deployment,my_app_service

What you should do instead:
@pytest.fixture()
def setup_configmap_for_openshift_app():
    return Configmap()

@pytest.fixture()
def setup_deployment_for_openshift_app():
    return Deployment()

@pytest.fixture()
def setup_service_for_openshift_app():
    return Service()

In general, if what your fixture is doing could be broken down in two (or more) different fixtures, you should do it.
```

lugi0 (approved):
small suggestion, lgtm otherwise

adolfo-ab (approved):

lugi0 (approved):

adolfo-ab (approved):

PR #162, opened 2025-02-27
[runtime] move ods-ci test

tarukumar:


tests/model_serving/model_runtime/vllm/utils.py
```diff
@@ -188,3 +189,31 @@ def validate_raw_tgis_inference_request(
...
+    tool_calling: dict[Any, Any] | None = None,
+) -> None:
+    model_info, chat_responses, completion_responses = fetch_openai_response(
+        url=url,
+        model_name=model_name,
+        chat_query=chat_query,
+        completion_query=completion_query,
+        tool_calling=tool_calling,
+    )
+    validate_inference_output(
```
adolfo-ab, line 209: Please call function with argument names

tarukumar: The function isn’t designed for that. In older tests, I will am sending multiple arguments, not just limited to those. That's why it's not a key argument; instead, I'm sending it as a positional argument. I have opened a bug in the rpeo to refactor the some older tests and will consider to  correct that and will change the in future

PR #174, opened 2025-03-11
run build container when PR is created, edited, reopened and rebased

dbasunag:
Add a workflow to run build container on PR create and rebase
## Description

Running build container on rebase and creation ensures, change does not break the test image
## How Has This Been Tested?

If the current PR has build container run and successfully completed, then the PR can be deemed verified/
## Merge criteria:


adolfo-ab (dismissed):

.github/workflows/verify_build_container.yml
```diff
@@ -0,0 +1,20 @@
+name: Build container using the PR
+
+on:
+  pull_request:
+    types: [ opened, synchronize, edited, reopened ]
```
lugi0, line 5: Do you know how long the build takes? It might be too much to have this run on every event, perhaps having this triggerable by a comment would be the better option

dbasunag: Looking at https://github.com/opendatahub-io/opendatahub-tests/actions/runs/13811294988/job/38633315625?pr=174, it is 1 minute currently. We should validate building the image in pr. It should never break and when it breaks you don't want to find out after merging a PR.

rnetser: please remove `edited`

lugi0: @rnetser I think removing edited defeats the purpose of this, as you'd have a build when the PR is opened but then (potentially) no other builds before it gets merged. With edited at least we'd have a check every time a commit is pushed

dbasunag: I agree with @lugi0. Webhook server that we use in many repos, runs this step on every edit. It may seem to be too much, but you don't want to find that build container does not work, after a PR is merged.

rnetser: `edited` also means pr title/description are updated, no? why build on those?

rnetser: btw, the webhook does not build on `edited`, what am i missing?

lugi0: @dbasunag I think @rnetser is right, https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=synchronize#pull_request
```
A pull request's head branch was updated. For example, the head branch was updated from the base branch or new commits were pushed to the head branch.
```
https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=edited#pull_request
```
The title or body of a pull request was edited, or the base branch of a pull request was changed.
```

I think we can remove `edited` and keep only `synchronize`

dbasunag: webhook, updates anytime the PR is updated (synchronize). I will remove edited.

dbasunag: @rnetser and @lugi0 removed `edited`.

lugi0 (approved):

rnetser (approved):

rnetser:
@dbasunag did you verify this PR (if needed do it in a test repo); if so, please mark as `verified`

dbasunag:
> @dbasunag did you verify this PR (if needed do it in a test repo); if so, please mark as `verified`

@rnetser since the PR workflow is showing it running, it is indirectly verified.

PR #189, opened 2025-03-19
On PR merge update the quay image

dbasunag:
## Merge criteria:

### DIFF
--- a/.github/workflows/build-push-container-on-merge.yml
+++ b/.github/workflows/build-push-container-on-merge.yml
@@ -1,0 +1,56 @@
+name: Build and Push Container Image On PR Merge
+
+on:
+  pull_request_target:
+    types: [closed]
+
+permissions:
+  pull-requests: write
+  contents: write
+  issues: write
+
+jobs:
+  if_merged:
+    if: github.event.pull_request.merged == true
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+      - name: Set env TAG
+        run: |
+          if [ ${{ github.event.pull_request.base.ref }} == "main" ]; then
+              echo "TAG=latest" >> $GITHUB_ENV
+          else
+              echo "TAG=${{ github.event.pull_request.base.ref }}" >> "$GITHUB_ENV"
+          fi
+      - name: Build Image
+        id: build-image
+        uses: redhat-actions/buildah-build@v2
+        with:
+          image: opendatahub-tests
+          tags: ${{ env.TAG }}
+          containerfiles: |
+            ./Dockerfile
+
+      - name: Push To Image Registry
+        id: push-to-registry
+        uses: redhat-actions/push-to-registry@v2
+        with:
+          image: ${{ steps.build-image.outputs.image }}
+          tags: ${{ steps.build-image.outputs.tags }}
+          registry: quay.io/opendatahub
+          username: ${{ secrets.QUAY_USERNAME }}
+          password: ${{ secrets.QUAY_PASSWORD }}
+
+      - name: Add comment to PR
+        if: always()
+        env:
+          URL: ${{ github.event.pull_request.comments_url }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          curl \
+            -X POST \
+            $URL \
+            -H "Content-Type: application/json" \
+            -H "Authorization: token $GITHUB_TOKEN" \
+            --data '{ "body": "Status of building tag ${{ env.TAG }}: ${{ steps.build-image.outcome }}. \nStatus of pushing tag ${{ env.TAG }} to image registry: ${{ steps.push-to-registry.outcome }}." }'


adolfo-ab (dismissed):

.github/workflows/build-push-container-on-merge.yml
```diff
@@ -0,0 +1,59 @@
...
+        with:
+          image: ${{ steps.build-image.outputs.image }}
+          tags: ${{ steps.build-image.outputs.tags }}
+          registry: ${{ secrets.IMAGE_REGISTRY }}/${{ secrets.IMAGE_NAMESPACE }}
+          username: ${{ secrets.QUAY_USERNAME }}
+          password: ${{ secrets.QUAY_PASSWORD }}
+
+      - name: Set env PR_COMMENT
+        run: |
+          echo "PR_COMMENT=Successfully updated image tag: ${{ env.TAG }}" >> $GITHUB_ENV
```
myakove, line 47: Successfully build and push ..... ?

.github/workflows/build-push-container-on-merge.yml
```diff
@@ -0,0 +1,59 @@
...
+          if [ ${{ github.event.pull_request.base.ref }} == "main" ]; then
+              echo "TAG=latest" >> $GITHUB_ENV
+          else
+              echo "TAG=${{ github.event.pull_request.base.ref }}" >> "$GITHUB_ENV"
+          fi
+      - name: Build Image
+        id: build-image
+        uses: redhat-actions/buildah-build@v2
+        with:
+          image: ${{ secrets.IMAGE_NAME }}
```
rnetser, line 30: why is the image name in a secret? the image is public

.github/workflows/build-push-container-on-merge.yml
```diff
@@ -0,0 +1,59 @@
...
+          containerfiles: |
+            ./Dockerfile
+
+      - name: Push To Image Registry
+        id: push-to-registry
+        uses: redhat-actions/push-to-registry@v2
+        with:
+          image: ${{ steps.build-image.outputs.image }}
+          tags: ${{ steps.build-image.outputs.tags }}
+          registry: ${{ secrets.IMAGE_REGISTRY }}/${{ secrets.IMAGE_NAMESPACE }}
```
rnetser, line 41: ditto

.github/workflows/build-push-container-on-merge.yml
```diff
@@ -0,0 +1,59 @@
...
+          username: ${{ secrets.QUAY_USERNAME }}
+          password: ${{ secrets.QUAY_PASSWORD }}
+
+      - name: Set env PR_COMMENT
+        run: |
+          echo "PR_COMMENT=Successfully built and pushed tag: ${{ env.TAG }}" >> $GITHUB_ENV
```
rnetser, line 47: what if it failed?

.github/workflows/build-push-container-on-merge.yml
dbasunag: If we want to share this information with public, I have no problem with that.

adolfo-ab (dismissed):

.github/workflows/build-push-container-on-merge.yml
rnetser: 1. so the same goes for a successful one, why add only for one flow?
2. how does the user know if `build` failed or `push` failed; should add clear comments on each (so there will be no need to check the logs)

rnetser (changes requested):

.github/workflows/build-push-container-on-merge.yml
dbasunag: Any step that fails, would cause the workflow fail and exit as we are not using continue-on-error. If the workflow fails, we already know it at which step it failed, no?

dbasunag: Updated the workflow to add both build and push step status as a comment to the pr.

rnetser (approved):

adolfo-ab (approved):

lugi0 (approved):

.github/workflows/build-push-container-on-merge.yml
```diff
@@ -0,0 +1,56 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+      - name: Set env TAG
+        run: |
+          if [ ${{ github.event.pull_request.base.ref }} == "main" ]; then
+              echo "TAG=latest" >> $GITHUB_ENV
+          else
+              echo "TAG=${{ github.event.pull_request.base.ref }}" >> "$GITHUB_ENV"
```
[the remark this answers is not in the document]
dbasunag: the tag should be like this:
```
latest: main branch
2.19: 2.19 branch and so on.
```
Otherwise, you would end up with lots of tags without any means to differentiate them or even understand quickly which one belongs to which branch (without having to compare the sha).

rnetser (approved):

adolfo-ab (approved):

PR #193, opened 2025-03-21
Support cherry-picking prs with /cherry-pick <branch_name> comment

dbasunag:
## Merge criteria:


.github/workflows/add-remove-labels.yml
```diff
@@ -21,6 +16,12 @@ permissions:
 
 jobs:
   add-remove-labels:
+    if: |
+      contains(github.event.comment.body, '/wip') ||
+      contains(github.event.comment.body, '/verified') ||
+      contains(github.event.comment.body, '/lgtm') ||
+      contains(github.event.comment.body, '/hold') ||
+      contains(github.event.comment.body, '/cherry-pick')
```
rnetser, line 24: please add to the welcome message

rnetser: what if i want to cherry-pick only to a specific branch?

.github/workflows/cherry_pick_comment.yml
```diff
@@ -0,0 +1,19 @@
+name: Cherry Pick On Comment
```
rnetser, line 1: the expected flow is when a pr is merged, look for the label and then cherry pick to the relevant branch(es).
cherry-pick should be blocked on an open pr

rnetser (changes requested):

.github/workflows/add-remove-labels.yml
dbasunag: I will update the welcome message. Your comment should indicate which branch you want to cherry-pick to `/cherry-pick <your branch>`

rnetser: >This ensures, comments added by cherry-pick are not getting 
this is very confusing - marking the comment as packed, is for the user to know the comment is/will be addressed and we should not rely on it for internal functionality

.github/workflows/cherry_pick_comment.yml
rnetser: >Current code flow is. Once a PR is merged, add a comment with /cherry-pick <branch_name>

this should be enabled not only when a pr is merged; as a user i would like to add the comment even when i just open the pr as the merge may not be done by me.

rnetser (changes requested):

.github/workflows/add-remove-labels.yml
dbasunag: Why? If cherry-pick action adds comments like ` ‼️ PR can't be cherry-picked, please merge it first.` or `Cherry pick action created PR $CHERRY_PICK_NUM successfully 🎉!`, why should they get a reaction? Only actions that are going to trigger a workflow should be getting reaction, no?

.github/workflows/cherry_pick_comment.yml
dbasunag: Workflow does not run that way. It is triggered on `types: [created]` of issue_comment. It would run when a user creates a comment. As far as I know there is no queue mechanism here. If we want that, then tide is the best way to proceed.

rnetser: this is why i think it should be in our python

or

on pr merge and check existing labels

rnetser (changes requested):

.github/workflows/cherry_pick_comment.yml
dbasunag: This should not be in python code (in the same repo). Workflow would need to check out the repo and cherry-pick would again need to checkout a different branch in order to do a cherry-pick. This would be a bad idea to add python code in this repository to handle that.

dbasunag: > on pr merge and check existing labels

This does not seem like a good idea to me. If I create a patch today and it gets merged say after a month, should my cherry-pick comment be considered relevant? I feel, it should be user's responsibility to add this cherry-pick comment explicitly when it is needed.

rnetser: > If I create a patch today and it gets merged say after a month, should my cherry-pick comment be considered relevant?

any reason why not? if the code should be cherry-picked. and this is a label so the user sees this label and can remove if they decide this is not needed.
i think it is safer than relying on adding the comment after merge

adolfo-ab (approved):

.github/workflows/cherry_pick_comment.yml
```diff
@@ -0,0 +1,19 @@
...
+on:
+  issue_comment:
+    types: [created]
+jobs:
+  cherry-pick:
+    name: Cherry Pick
+    if: github.event.issue.pull_request != '' && contains(github.event.comment.body, '/cherry-pick')
```
rnetser, line 8: iiuc this will allow anyone adding the comment to trigger cherry-pick right?
saw in other places this check added as well:
```
github.event.comment.author_association == 'CONTRIBUTOR'
```

so we need to adapt something like this here

dbasunag: Correct. This is a good idea.

adolfo-ab (dismissed):

.github/workflows/cherry-pick-comment.yml
```diff
@@ -0,0 +1,25 @@
...
+jobs:
+  cherry-pick:
+    name: Cherry Pick
+    if: |
+      github.event.issue.pull_request != '' &&
+      contains(github.event.comment.body, '/cherry-pick') &&
+      ((github.event.pull_request.author_association != 'CONTRIBUTOR') ||
+            (github.event.pull_request.author_association != 'MEMBER') ||
+            (github.event.pull_request.author_association != 'OWNER') ||
+             (github.event.pull_request.author_association != 'COLLABORATOR'))
```
rnetser, line 14: does it mean that other user associations like `FIRST_TIMER` work? 
why isn't `github.event.pull_request.author_association == 'MEMBER'` enough?

i.e anyone in the org can add or `CONTRIBUTOR` because it on a pr that was merged by a contributing user?

rnetser (approved):

adolfo-ab (approved):

PR #200, opened 2025-03-26
Wait for serverless pod deletion before ns deletion

rnetser:
This is a workaround for [RHOAIENG-19969](https://issues.redhat.com/browse/RHOAIENG-19969).
    When Serverless ISVC is deleted and the namespace is deleted, namespace "SomeResourcesRemain" is True.
    This is because the serverless pods are not immediately deleted, resulting in prolonged namespace deletion.


utilities/infra.py
```diff
@@ -674,3 +705,26 @@ def get_operator_distribution(client: DynamicClient, dsc_name: str = "default-ds
...
+    """
+    client = admin_client or get_client()
+    for pod in Pod.get(dyn_client=client, namespace=resource.name):
+        if (
+            pod.exists
+            and pod.instance.metadata.annotations.get(Annotations.KserveIo.DEPLOYMENT_MODE)
+            == KServeDeploymentType.SERVERLESS
+        ):
+            LOGGER.info(f"Waiting for {KServeDeploymentType.SERVERLESS} pod {pod.name} to be deleted")
+            pod.wait_deleted(timeout=Timeout.TIMEOUT_1MIN)
```
dbasunag, line 730: Instead of sequentially waiting for each of them to be deleted, why not check the list of pods that matches this annotation, till the list becomes empty list?

rnetser: because the other pods are being deleted at the same time as the first pod so once we wait for the 1st one to be deleted, the other ones are either deleted or will be deleted soon after.
saving API calls.

dbasunag (approved):

mwaykole (approved):

adolfo-ab (approved):
Just a question, feel free to merge if it doesn't make sense

utilities/infra.py
```diff
@@ -674,3 +705,26 @@ def get_operator_distribution(client: DynamicClient, dsc_name: str = "default-ds
             raise ValueError("DSC release name not found in {dsc_name}")
 
     raise MissingResourceError(f"DSC {dsc_name} not found")
+
+
+def wait_for_serverless_pods_deletion(resource: Project | Namespace, admin_client: DynamicClient | None) -> None:
```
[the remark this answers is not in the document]
rnetser: this is currently a w/a that should be deleted when we find the cause for the delayed deletion of the pod

dbasunag (approved):

mwaykole (approved):

adolfo-ab (approved):

PR #213, opened 2025-04-02
Test checking out base repo in gha

adolfo-ab:
TEST


rnetser (commented):
test

dbasunag:
> Test if checking out base repo instead of fork fixes the add-remove-labels workflow

Which part is problematic? With the if statement moving, with this PR https://github.com/opendatahub-io/opendatahub-tests/pull/193, I don't see any issues with reactions to comments that needs it and no-reactions to comments that does not need it.

PR #214, opened 2025-04-02
[wip] add failure on Image wrong name and fail fast if no inference pods

rnetser:
## Merge criteria:


utilities/infra.py
```diff
@@ -583,10 +584,31 @@ def verify_no_failed_pods(
...
+
     Raises:
-            FailedPodsError: If any pod is in failed state
+        FailedPodsError: If any pod is in failed state
+        TimeoutExpiredError: If no pods were created within `timeout_wait_for_pods` seconds.
 
     """
+    LOGGER.info("Wait for pods creation")
+
+    try:
```
dbasunag, line 596: This is a great candidate for retry decorator :)

dbasunag, line 596: May be this should be part of another function and this function should be called only when the previous one passes? I don't really like the idea of a function doing many things. Triage becomes messy as it grows.

utilities/infra.py
```diff
@@ -583,10 +584,31 @@ def verify_no_failed_pods(
...
+        for pods in TimeoutSampler(
+            wait_timeout=timeout_wait_for_pods,
+            sleep=10,
+            func=get_pods_by_isvc_label,
+            client=client,
+            isvc=isvc,
+            runtime_name=runtime_name,
+        ):
+            if pods:
```
dbasunag, line 605: Should it matter how many pods you are expecting or as long as some pods exists, we can proceed with next step?

utilities/infra.py
rnetser: this pr is marked as  `wip`, please do not review

rnetser:
/wip cancel

now ready for review

utilities/infra.py
rnetser: done

rnetser: done

utilities/infra.py
rnetser: no; at this stage have at least one is enough

adolfo-ab (approved):

dbasunag (approved):

PR #219, opened 2025-04-04
change: Update required python version

fege:
## Description
Update required python version, when using 3.9 version there are some import that are not available for example
```
    from typing import Self
E   ImportError: cannot import name 'Self' from 'typing' 
``` 
This was added only in 3.11 [typing.Self](https://docs.python.org/3/library/typing.html#typing.Self) 
Also `|` for `Union` was added in 3.10 [typing.Union](https://docs.python.org/3/library/typing.html#typing.Union)

## Merge criteria:


lugi0 (approved):

pyproject.toml
```diff
@@ -27,7 +27,7 @@ dev-dependencies = [
 [tool.uv.sources]
 
 [project]
-requires-python = ">=3.9"
+requires-python = ">=3.11"
```
dbasunag, line 30: Can we go with 3.12 or 3.13 instead? 3.11 is now security status.

fege: changed to 3.12

dbasunag (dismissed):

rnetser (changes requested):
python 3.9 is supported till 20/2025.
the repo uses
`from __future__ import annotations`

use `typing_extensions` for `Self`

rnetser (changes requested):
python 3.9 is supported till 20/2025.
the repo uses
from __future__ import annotations

use typing_extensions for Self

rnetser:
adding as a comment so it is visible in labels:

python 3.9 is supported till 20/2025.
the repo uses
from __future__ import annotations

use typing_extensions for Self

dbasunag:
> Title
@rnetser I personally feel the test repo should be using newer python versions, when possible. 
`from future import annotations` should be fine with 3.12. openshift-virtualization uses 3.12 and uses the same import and did not encounter issues.

For `use typing_extensions for Self`, not sure if we would need it. https://pypi.org/project/typing-extensions/ says it is for older python versions to enable newer typing features and I see Self is here https://github.com/python/cpython/blob/3.12/Lib/typing.py. So I might be missing something, but it does not seem like `from typing import Self` should be problematic in 3.12

rnetser:
> > > > adding as a comment so it is visible in labels:
> > > > python 3.9 is supported till 20/2025.
> > > > the repo uses
> > > > from future import annotations
> > > > use typing_extensions for Self
> > > 
> > > 
> > > @rnetser I personally feel the test repo should be using newer python versions, when possible. `from future import annotations` should be fine with 3.12. openshift-virtualization uses 3.12 and uses the same import and did not encounter issues.
> > 
> > 
> > if there's no reason to restrict a version i do not see any reason why limit
> > > For `use typing_extensions for Self`, not sure if we would need it. https://pypi.org/project/typing-extensions/ says it is for older python versions to enable newer typing features and I see Self is here https://github.com/python/cpython/blob/3.12/Lib/typing.py. So I might be missing something, but it does not seem like `from typing import Self` should be problematic in 3.12
> 
> @rnetser are you suggesting to remove completely `requires-python` ? I personally think that there is no harm to update to 3.12 given also that this week ended the active support on it https://endoflife.date/python

I meant leave it on 3.9

dbasunag:
> > > > > adding as a comment so it is visible in labels:
> > > > > python 3.9 is supported till 20/2025.
> > > > > the repo uses
> > > > > from future import annotations
> > > > > use typing_extensions for Self
> > > > 
> > > > 
> > > > @rnetser I personally feel the test repo should be using newer python versions, when possible. `from future import annotations` should be fine with 3.12. openshift-virtualization uses 3.12 and uses the same import and did not encounter issues.
> > > 
> > > 
> > > if there's no reason to restrict a version i do not see any reason why limit
> > > > For `use typing_extensions for Self`, not sure if we would need it. https://pypi.org/project/typing-extensions/ says it is for older python versions to enable newer typing features and I see Self is here https://github.com/python/cpython/blob/3.12/Lib/typing.py. So I might be missing something, but it does not seem like `from typing import Self` should be problematic in 3.12
> > 
> > 
> > @rnetser are you suggesting to remove completely `requires-python` ? I personally think that there is no harm to update to 3.12 given also that this week ended the active support on it https://endoflife.date/python
> 
> I meant leave it on 3.9

@rnetser can you elaborate why you recommend staying with 3.9? 3.9 active support ended in May 22 and eol is on Oct 2025.

pyproject.toml
```diff
@@ -27,7 +27,7 @@ dev-dependencies = [
 [tool.uv.sources]
 
 [project]
-requires-python = ">=3.9"
+requires-python = ">=3.12"
```
rnetser, line 30: did you check 3.13? I think there are packages that we use that will fail with it

fege: modified to `requires-python = "==3.12.*"`

PR #221, opened 2025-04-07
feat: add test for incorrect DB TLS config in Trusty AI

sheltoncyril:
Add test to check if TrustyAI service is unhealthy after invalid database TLS configuration

## Description
The idea behind this test is to check if the TrustyAI service fails if it is  unable to communicate with the MariaDB service when an invalid TLS certificate is set as it's secret. I've introduced a constant which has a fake certificate and is base64 encoded. As of now, I'm waiting for deployment to timeout after 2 minutes as the service will never be healthy.

## How Has This Been Tested?
All tests were run and no side-effects were observed

## Merge criteria:

## Summary by Sourcery

Add a test to validate TrustyAI service behavior when configured with an invalid database TLS certificate.

Tests:
- Add a test case to verify that the TrustyAI service pod fails with a specific TLS certificate validation error when an invalid database certificate secret is provided.
- Introduce test fixtures to set up the TrustyAI service with an invalid database certificate.
- Add a validation function to check for the expected pod failure state and error message related to DB connection issues.
- Refactor TrustyAIService creation into a reusable test helper function.

## Summary by CodeRabbit

- **New Features**
  - Added test coverage for scenarios where the TrustyAIService fails to connect due to an invalid database TLS certificate.
  - Introduced new fixtures to enable testing with invalid database certificates.

- **Bug Fixes**
  - Improved fixture behavior to yield actual Secret objects, enhancing test reliability.

- **Refactor**
  - Centralized and streamlined TrustyAIService test setup with reusable helper functions and configuration constants.

- **Tests**
  - Expanded test suite to validate database connection failures and improved test modularity.

- **Chores**
  - Added custom exceptions for clearer error handling in test utilities.


tests/model_explainability/trustyai_service/conftest.py
```diff
@@ -238,17 +268,32 @@ def mariadb(
         yield mariadb
 
 
-@pytest.fixture(scope="class")
-def trustyai_db_ca_secret(
-    admin_client: DynamicClient, model_namespace: Namespace, mariadb: MariaDB
+def set_trustyai_db_ca_secret(
+    admin_client: DynamicClient, mariadb_ca_cert: str, model_namespace: Namespace
```
adolfo-ab, line 272: Just a nitpick but I would rename arguments, they don't need to have the same name as the fixtures:
`namespace` instead of `model_namespace`
`client` instead of `admin_client`

and put `mariadb_ca_cert` last

tests/model_explainability/trustyai_service/conftest.py
```diff
@@ -238,17 +268,32 @@ def mariadb(
         yield mariadb
 
 
-@pytest.fixture(scope="class")
-def trustyai_db_ca_secret(
-    admin_client: DynamicClient, model_namespace: Namespace, mariadb: MariaDB
+def set_trustyai_db_ca_secret(
```
adolfo-ab, line 271: Make it a contextmanager: `@contextmanager`

adolfo-ab, line 271: Rename to `create_trustyai_db_ca_secret`

tests/model_explainability/trustyai_service/conftest.py
```diff
@@ -238,17 +268,32 @@ def mariadb(
...
+
+
+@pytest.fixture(scope="class")
+def trustyai_db_ca_secret(
+    admin_client: DynamicClient, model_namespace: Namespace, mariadb: MariaDB
+) -> Generator[None, Any, None]:
+    mariadb_ca_secret = Secret(
+        client=admin_client, name=f"{mariadb.name}-ca", namespace=model_namespace.name, ensure_exists=True
+    )
+    mariadb_ca_cert = mariadb_ca_secret.instance.data["ca.crt"]
```
adolfo-ab, line 290: No need to save this variable, just use `mariadb_ca_secret.instance.data["ca.crt"]` directly

tests/model_explainability/trustyai_service/conftest.py
```diff
@@ -238,17 +268,32 @@ def mariadb(
...
+    yield from set_trustyai_db_ca_secret(admin_client, mariadb_ca_cert, model_namespace)
+
+
+@pytest.fixture(scope="class")
+def trustyai_invalid_db_ca_secret(
+    admin_client: DynamicClient, model_namespace: Namespace, mariadb: MariaDB
+) -> Generator[None, Any, None]:
+    mariadb_ca_cert = FAKE_TLS_CERTIFICATE
```
adolfo-ab, line 298: Ditto

tests/model_explainability/trustyai_service/conftest.py
```diff
@@ -35,6 +37,11 @@
 DB_NAME: str = "trustyai_db"
 DB_USERNAME: str = "trustyai_user"
 DB_PASSWORD: str = "trustyai_password"
+FAKE_TLS_CERTIFICATE: str = (
```
adolfo-ab, line 40: Rename to `INVALID_TLS_CERTIFICATE`

tests/model_explainability/trustyai_service/fairness/test_fairness.py
```diff
@@ -156,6 +156,16 @@ def test_fairness_send_inference_and_verify_trustyai_service_with_db_storage(
...
 
+    def test_invalid_db_certificate_trustyai_service_with_db_storage(
+        self,
+        admin_client,
+        current_client_token,
+        model_namespace,
+        trustyai_service_with_invalid_cert_db_storage,
+        minio_data_connection,
+    ):
+        return
```
adolfo-ab, line 167: You are not testing anything here, as you tried to do the testing part in the fixtures. Fixtures are to set up the resources, not to test. Here you would want to assert that the TrustyAIService deployment (or the deployment pods, I'm not sure), fail.

tests/model_explainability/trustyai_service/fairness/test_fairness.py
```diff
@@ -156,6 +156,16 @@ def test_fairness_send_inference_and_verify_trustyai_service_with_db_storage(
             inference_config=OPENVINO_KSERVE_INFERENCE_CONFIG,
         )
 
+    def test_invalid_db_certificate_trustyai_service_with_db_storage(
```
adolfo-ab, line 159: Rename to `test_trustyai_service_with_db_storage_invalid_cert`

adolfo-ab, line 159: Add docstring explaining what are we testing and the expected result.

tests/model_explainability/trustyai_service/conftest.py
adolfo-ab, line 271: Also, move to utils

adolfo-ab (changes requested):

tests/model_explainability/trustyai_service/utils.py
```diff
@@ -44,27 +49,91 @@ def wait_for_mariadb_operator_deployments(mariadb_operator: MariadbOperator) ->
...
+                label_selector=label_selector,
             )
         ]
         return pods
 
-    sampler = TimeoutSampler(wait_timeout=timeout, sleep=1, func=lambda: bool(_get_mariadb_pods()))
+    return _get_pods
+
+
+def wait_for_pods(
```
adolfo-ab, line 67: Ditto, there's no point in replacing `wait_for_mariadb_pods` with this function, since we still only use it in the `mariadb` fixture. Just leave the `wait_for_mariadb_pods` function as it was before.

tests/model_explainability/trustyai_service/utils.py
```diff
@@ -44,27 +49,91 @@ def wait_for_mariadb_operator_deployments(mariadb_operator: MariadbOperator) ->
         deployment.wait_for_replicas()
 
 
-def wait_for_mariadb_pods(client: DynamicClient, mariadb: MariaDB, timeout: int = Timeout.TIMEOUT_5MIN) -> None:
-    def _get_mariadb_pods() -> list[Pod]:
+def get_pods_func(client: DynamicClient, namespace: str, label_selector: Optional[str]) -> Callable[[], list[Pod]]:
+    def _get_pods() -> List[Pod]:
```
adolfo-ab, line 53: Just leave `wait_for_mariadb_pods` as it is right now, no need to create `wait_for_pods` or `get_pods_func`

tests/model_explainability/trustyai_service/utils.py
```diff
@@ -44,27 +49,91 @@ def wait_for_mariadb_operator_deployments(mariadb_operator: MariadbOperator) ->
...
+            timeout=timeout,  # IDEA: using two different timeouts might be worth considering
         )
+    return pods
+
+
+def wait_for_trustyai_container_failure_and_get_status(
+    client: DynamicClient, namespace: str, label_selector: Optional[str] = None, timeout: int = Timeout.TIMEOUT_5MIN
+) -> Any | None:
+    get_pods = get_pods_func(client, namespace=namespace, label_selector=label_selector)
+    sampler = TimeoutSampler(wait_timeout=timeout, sleep=1, func=get_pods)
```
adolfo-ab, line 96: No need to create and use `get_pods_func` just do `func=lambda: Pods.get(...)`

tests/model_explainability/trustyai_service/utils.py
```diff
@@ -44,27 +49,91 @@ def wait_for_mariadb_operator_deployments(mariadb_operator: MariadbOperator) ->
...
+    for pods in sampler:
+        for pod in pods:
```
adolfo-ab, line 98: No need to loop over pods, as we should have only 1 (and we should fail if we have more than 1)

tests/model_explainability/trustyai_service/utils.py
```diff
@@ -44,27 +49,91 @@ def wait_for_mariadb_operator_deployments(mariadb_operator: MariadbOperator) ->
...
+            for container_status in pod.instance.status.containerStatuses:
+                if (terminate_state := container_status.lastState.terminated) and terminate_state.reason in (
+                    pod.Status.ERROR,
+                    pod.Status.CRASH_LOOPBACK_OFF,
+                ):
+                    return terminate_state
+    return None
```
adolfo-ab, line 105: If we reach the last sample we will fail with `TimeoutError`, so I think we will never reach this point

tests/model_explainability/trustyai_service/utils.py
```diff
@@ -44,27 +49,91 @@ def wait_for_mariadb_operator_deployments(mariadb_operator: MariadbOperator) ->
...
+    get_pods = get_pods_func(client, namespace=namespace, label_selector=label_selector)
+    sampler = TimeoutSampler(wait_timeout=timeout, sleep=1, func=get_pods)
+    for pods in sampler:
+        for pod in pods:
+            for container_status in pod.instance.status.containerStatuses:
+                if (terminate_state := container_status.lastState.terminated) and terminate_state.reason in (
+                    pod.Status.ERROR,
+                    pod.Status.CRASH_LOOPBACK_OFF,
+                ):
+                    return terminate_state
```
adolfo-ab, line 104: Instead of returning I think you should assert the messages from here, then the only thing you do from the test is calling this function and call it a day

adolfo-ab (changes requested):

tests/model_explainability/trustyai_service/test_trustyai_service.py
```diff
@@ -0,0 +1,33 @@
...
+):
+    """
+    Makes sure TrustyAIService pod fails when incorrect database TLS certificate is used.
+    """
+    if terminate_state := wait_for_trustyai_container_terminal_state(
+        client=admin_client,
+        namespace=model_namespace.name,
+        label_selector=f"app.kubernetes.io/instance={trustyai_service_with_invalid_db_cert.name}",
+    ):
+        assert "Could not connect to mariadb:3306" in terminate_state.message
```
dbasunag, line 30: Let's not use assert in tests. Instead let's use if and fail

tests/model_explainability/trustyai_service/utils.py
```diff
@@ -44,27 +49,91 @@ def wait_for_mariadb_operator_deployments(mariadb_operator: MariadbOperator) ->
...
+def wait_for_pods(
+    client: DynamicClient,
+    namespace: str,
+    label_selector: Optional[str] = None,
+    timeout: int = Timeout.TIMEOUT_5MIN,
+    wait_for_condition: str = Pod.Condition.READY,
+    condition_status: str = Pod.Condition.Status.TRUE,
+) -> List[Pod]:
+    get_pods = get_pods_func(client=client, namespace=namespace, label_selector=label_selector)
+    sampler = TimeoutSampler(wait_timeout=timeout, sleep=1, func=lambda: bool(get_pods()))
```
dbasunag, line 76: Polling every seconds for 5 minutes seems a bit too much. Can we poll every 5 seconds may be?

tests/model_explainability/trustyai_service/utils.py
dbasunag: +1

tests/model_explainability/trustyai_service/utils.py
```diff
@@ -44,27 +49,91 @@ def wait_for_mariadb_operator_deployments(mariadb_operator: MariadbOperator) ->
...
-            condition=Pod.Condition.READY,
-            status="True",
+            condition=wait_for_condition,
+            status=condition_status,
+            timeout=timeout,  # IDEA: using two different timeouts might be worth considering
         )
+    return pods
+
+
+def wait_for_trustyai_container_terminal_state(
```
dbasunag, line 92: instead of timeout sampler like this, please use @retry decorator.

tests/model_explainability/trustyai_service/conftest.py
```diff
@@ -198,7 +215,9 @@ def mariadb(
     mariadb_dict["spec"]["rootPasswordSecretKeyRef"] = password_secret_key_ref
     mariadb_dict["spec"]["passwordSecretKeyRef"] = password_secret_key_ref
     with MariaDB(kind_dict=mariadb_dict) as mariadb:
-        wait_for_mariadb_pods(client=admin_client, mariadb=mariadb)
+        wait_for_pods(
+            client=admin_client, namespace=mariadb.namespace, label_selector="app.kubernetes.io/instance=mariadb"
```
rnetser, line 219:
```suggestion
            client=admin_client, namespace=mariadb.namespace, label_selector=f"{mariadb.ApiGroup.APP_KUBERNETES_IO}/instance={MARIADB}"
```

tests/model_explainability/trustyai_service/test_trustyai_service.py
```diff
@@ -0,0 +1,33 @@
...
+from ocp_resources.namespace import Namespace
+
+from tests.model_explainability.trustyai_service.utils import wait_for_trustyai_container_terminal_state
+
+
+@pytest.mark.parametrize(
+    "model_namespace",
+    [
+        pytest.param(
+            {"name": "test-trustyai-service"},
```
rnetser, line 11: please give a meaningful name (related to what is tested) to a namespace so if there's a failure it is easy to debug

tests/model_explainability/trustyai_service/utils.py
```diff
@@ -44,27 +49,91 @@ def wait_for_mariadb_operator_deployments(mariadb_operator: MariadbOperator) ->
...
+            for container_status in pod.instance.status.containerStatuses:
+                if (terminate_state := container_status.lastState.terminated) and terminate_state.reason in (
+                    pod.Status.ERROR,
+                    pod.Status.CRASH_LOOPBACK_OFF,
+                ):
+                    return terminate_state
+    return None
+
+
+def create_trustyai_db_ca_secret(
```
rnetser, line 108: this is just a wrapper for creating a secret, call directly where needed

tests/model_explainability/trustyai_service/utils.py
rnetser, line 92: in all functions please add proper docstrings

rnetser (changes requested):

tests/model_explainability/trustyai_service/utils.py
```diff
@@ -68,3 +74,67 @@ def _get_mariadb_pods() -> list[Pod]:
...
+    Args:
+        client: DynamicClient
+        namespace: Namespace
+        label_selector: Optional(str)
+    Returns: bool
+    """
+    pods = list(Pod.get(dyn_client=client, namespace=namespace, label_selector=label_selector))
+    if len(pods) != 0:
+        if len(pods) > 1:
+            pytest.fail("More than one pod found in TrustyAIService.")
```
dbasunag, line 97: What if the pod is getting deleted, just needs a few more seconds to clean up. Should that trigger a failure?

tests/model_explainability/trustyai_service/utils.py
```diff
@@ -68,3 +74,67 @@ def _get_mariadb_pods() -> list[Pod]:
...
+@retry(wait_timeout=Timeout.TIMEOUT_2MIN, sleep=5)
+def validate_trustyai_service_db_conn_failure(
+    client: DynamicClient, namespace: str, label_selector: Optional[str]
+) -> bool:
+    """
+    Waits for TrustyAIService pod to fail and checks if the pod is:-
+    * in a CrashLoopBackOff state
+    * The LastState is in terminated state and the cause was a MariaDB TLS certificate exception
+    Also checks if there are more than one pod for the service.
+    Args:
```
dbasunag, line 88: Please add lines between docs sections. e.g. before line 88 and 92

tests/model_explainability/trustyai_service/utils.py
sheltoncyril: Hey, I think by design TrustyAIService only creates one pod and in case of this test that pod goes into CrashLoopBackOff, It does not create another pod to replace it. Maybe @adolfo-ab can chime in about this as he suggested that I check if there are more than one pods

tests/model_explainability/trustyai_service/conftest.py
```diff
@@ -41,19 +47,14 @@ def trustyai_service_with_pvc_storage(
...
 ) -> Generator[TrustyAIService, Any, Any]:
-    with TrustyAIService(
+    yield from create_trustyai_service(
         client=admin_client,
-        name=TRUSTYAI_SERVICE_NAME,
-        namespace=model_namespace.name,
+        namespace=model_namespace,
         storage={"format": "PVC", "folder": "/inputs", "size": "1Gi"},
         data={"filename": "data.csv", "format": "CSV"},
         metrics={"schedule": "5s"},
```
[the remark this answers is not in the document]
adolfo-ab: We use the same `metrics` everywhere, so we can save them in a constant. Same with `storage` in the db fixtures

tests/model_explainability/trustyai_service/utils.py
adolfo-ab: Yes, a new pod should not be created. I don't think we should worry about this case as of now.

tests/model_explainability/trustyai_service/conftest.py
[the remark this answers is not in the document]
sheltoncyril: Refactored metrics and storage as constants

adolfo-ab (dismissed):

dbasunag (dismissed):

tests/model_explainability/trustyai_service/utils.py
```diff
@@ -68,3 +74,69 @@ def _get_mariadb_pods() -> list[Pod]:
...
+        for container_status in pods[0].instance.status.containerStatuses:
+            if (terminate_state := container_status.lastState.terminated) and terminate_state.reason in (
+                pods[0].Status.ERROR,
+                pods[0].Status.CRASH_LOOPBACK_OFF,
+            ):
+                if (
+                    "Could not connect to mariadb:" not in terminate_state.message
+                    and "java.security.cert.CertPathValidatorException: signature check failed"
+                    not in terminate_state.message
+                ):
```
rnetser, line 109: please use a regex to check `terminate_state.message` instead of 2 separate checks

tests/model_explainability/trustyai_service/utils.py
```diff
@@ -68,3 +74,69 @@ def _get_mariadb_pods() -> list[Pod]:
...
+                    pytest.fail(
+                        f"Service {TRUSTYAI_SERVICE_NAME} did not fail with a mariadb connection failure as expected."
```
rnetser, line 111: please expected text in the error

rnetser (changes requested):

adolfo-ab (dismissed):

tests/model_explainability/trustyai_service/utils.py
```diff
@@ -68,3 +75,69 @@ def _get_mariadb_pods() -> list[Pod]:
...
+                    )
+                return True
+    return False
+
+
+def create_trustyai_service(
+    client: DynamicClient,
+    namespace: Namespace,
+    storage: dict[Any, Any],
+    metrics: dict[Any, Any],
```
dbasunag, line 123: I don't see a need to use Any here. str, str, should be good enough.

tests/model_explainability/trustyai_service/utils.py
```diff
@@ -68,3 +75,69 @@ def _get_mariadb_pods() -> list[Pod]:
...
+
+
+@retry(wait_timeout=Timeout.TIMEOUT_2MIN, sleep=5)
+def validate_trustyai_service_db_conn_failure(
+    client: DynamicClient, namespace: str, label_selector: Optional[str]
+) -> bool:
+    """
+    Waits for TrustyAIService pod to fail and checks if the pod is in a CrashLoopBackOff state and
+    the LastState is in terminated state and the cause was a MariaDB TLS certificate exception.
+    Also checks if there are more than one pod for the service.
```
rnetser, line 87: please check https://google.github.io/styleguide/pyguide.html#381-docstrings

```
A docstring should be organized as a summary line (one physical line not exceeding 80 characters) terminated by a period, question mark, or exclamation point. When writing more (encouraged), this must be followed by a blank line, followed by the rest of the docstring starting at the same cursor position as the first quote of the first line
```

tests/model_explainability/trustyai_service/utils.py
```diff
@@ -68,3 +75,69 @@ def _get_mariadb_pods() -> list[Pod]:
...
+        client: DynamicClient
+        namespace: Namespace
+        label_selector: Optional(str)
+
+    Returns: bool
+    """
+    pods = list(Pod.get(dyn_client=client, namespace=namespace, label_selector=label_selector))
+    mariadb_conn_failure_regex = (
+        r"^.+ERROR.+Could\snot\sconnect\sto\smariadb:.+\sPKIX\spath\svalidation\sfailed:\s"
+        r"java\.security.cert.CertPathValidatorException:\ssignature\scheck\sfailed"
```
rnetser, line 99: are the explicit `\s` needed or can you just use whitespace?
e.g 
`Could\snot\sconnect` -> `Could not connect`

rnetser (changes requested):

tests/model_explainability/trustyai_service/utils.py
```diff
@@ -68,3 +75,78 @@ def _get_mariadb_pods() -> list[Pod]:
...
+    Waits for TrustyAIService pod to fail and checks if the pod is in a CrashLoopBackOff state and
+    the LastState is in terminated state and the cause was a MariaDB TLS certificate exception.
+    Also checks if there are more than one pod for the service.
+
+    Args:
+        client: The OpenShift client.
+        namespace: Namespace under which the pod is created.
+        label_selector: The label selector used to select the correct pod(s) to monitor.
+
+    Returns: True if pod failure is of expected state else False.
```
rnetser, line 95: Missing `Raises: ...` on timeout

tests/model_explainability/trustyai_service/utils.py
```diff
@@ -68,3 +75,78 @@ def _get_mariadb_pods() -> list[Pod]:
...
+    """
+    pods = list(Pod.get(dyn_client=client, namespace=namespace.name, label_selector=label_selector))
+    mariadb_conn_failure_regex = (
+        r"^.+ERROR.+Could not connect to mariadb:.+ PKIX path validation failed: "
+        r"java\.security\.cert\.CertPathValidatorException: signature check failed"
+    )
+    if pods:
+        if len(pods) > 1:
+            pytest.fail("More than one pod found in TrustyAIService.")
```
rnetser, line 104: this is within a timeout sampler; will not be raise, right?

tests/model_explainability/trustyai_service/utils.py
```diff
@@ -68,3 +75,78 @@ def _get_mariadb_pods() -> list[Pod]:
...
+        for container_status in pods[0].instance.status.containerStatuses:
+            if (terminate_state := container_status.lastState.terminated) and terminate_state.reason in (
+                pods[0].Status.ERROR,
+                pods[0].Status.CRASH_LOOPBACK_OFF,
+            ):
+                if not re.search(mariadb_conn_failure_regex, terminate_state.message):
+                    pytest.fail(
```
rnetser, line 111: ditto

rnetser (changes requested):

rnetser (approved):

PR #226, opened 2025-04-07
[ModelRegistry] ensure RunAsUser and RunAsGroup are not set explicitly

dbasunag:
## Merge criteria:


tests/model_registry/test_model_registry_scc.py
```diff
@@ -0,0 +1,149 @@
...
+
+KEYS_TO_VALIDATE = ["runAsGroup", "runAsUser"]
+LOGGER = get_logger(name=__name__)
+
+
+@pytest.fixture(scope="class")
+def model_registry_scc_namespace(model_registry_namespace: Namespace):
+    mr_annotations = model_registry_namespace.instance.metadata.annotations
+    return {
+        "selinuxOtions": mr_annotations.get("openshift.io/sa.scc.mcs"),
```
adolfo-ab, line 24: Is this a typo? `selinuxOtions` vs `selinuxOptions`

dbasunag: Yes, typo :)

tests/model_registry/test_model_registry_scc.py
```diff
@@ -0,0 +1,149 @@
...
+                        "registriesNamespace": MR_NAMESPACE,
+                    },
+                },
+            },
+            MODEL_DICT,
+        ),
+    ],
+    indirect=True,
+)
+@pytest.mark.usefixtures("model_registry_namespace", "updated_dsc_component_state_scope_class", "registered_model")
```
rnetser, line 66: `model_registry_resource` is using `model_registry_namespace`, why is this needed?

tests/model_registry/test_model_registry_scc.py
```diff
@@ -0,0 +1,149 @@
...
+@pytest.mark.parametrize(
+    "model_registry_namespace, updated_dsc_component_state_scope_class, registered_model",
+    [
+        pytest.param(
+            {"namespace_name": MR_NAMESPACE},
+            {
+                "component_patch": {
+                    DscComponents.MODELREGISTRY: {
+                        "managementState": DscComponents.ManagementState.MANAGED,
+                        "registriesNamespace": MR_NAMESPACE,
```
rnetser, line 57: this is repeated in almost all tests, save and re-use

tests/model_registry/test_model_registry_scc.py
```diff
@@ -0,0 +1,149 @@
...
+        "model_registry_resource",
+        [
+            pytest.param({"kind": Deployment}),
+        ],
+        indirect=["model_registry_resource"],
+    )
+    def test_model_registry_deployment_security_context_validation(
+        self: Self,
+        model_registry_resource: NamespacedResource,
+        model_registry_namespace,
```
rnetser, line 78: typing is missing (tbh, i do not think we should add typing to tests)

tests/model_registry/test_model_registry_scc.py
```diff
@@ -0,0 +1,149 @@
...
+    ):
+        """
+        Validate that model registry deployment does not set runAsUser/runAsGroup
+        """
+        error = []
+        for container in model_registry_resource.instance.spec.template.spec.containers:
+            if not all([True for key in KEYS_TO_VALIDATE if not container.get(key)]):
```
rnetser, line 85:
```suggestion
            if not all([not container.get(key) for key in KEYS_TO_VALIDATE ]):
```

tests/model_registry/test_model_registry_scc.py
```diff
@@ -0,0 +1,149 @@
...
+    ):
+        """
+        Validate that model registry pod gets runAsUser/runAsGroup from openshift and the values matches namespace
+        annotations
+        """
+        errors = []
+        uid_ranges = model_registry_scc_namespace["uid-range"].split("/")
+        ns_uid = uid_ranges[0]
+        pod_spec = model_registry_resource.instance.spec
+        pod_selinux_option = pod_spec.securityContext.get("seLinuxOptions", {}).get("level")
```
rnetser, line 114: please split to functions by tested functionality; it is hard to follow what is being tested

tests/model_registry/test_model_registry_scc.py
```diff
@@ -0,0 +1,149 @@
...
+        pods = get_pods_by_name_prefix(
+            client=admin_client, pod_prefix=MR_INSTANCE_NAME, namespace=model_registry_namespace.name
+        )
+        if len(pods) != 1:
+            pytest.fail(
+                "Expected one model registry pod. Found: {[{pod.name: pod.status} for pod in pods] if pods else None}"
+            )
+        return pods[0]
+    else:
+        raise AssertionError(f"Invalid resouce requests: {request.param['kind']}")
```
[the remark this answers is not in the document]
dbasunag: And I just noticed a typo too `resouce`.

tests/model_registry/test_model_registry_scc.py
dbasunag: I will add it for now. But I too was wondering why should we have typing in tests. :)

tests/model_registry/test_model_registry_scc.py
dbasunag: I like this suggestion, so will update.

tests/model_registry/test_model_registry_scc.py
```diff
@@ -0,0 +1,149 @@
...
+
+from kubernetes.dynamic import DynamicClient
+from ocp_utilities.infra import get_pods_by_name_prefix
+
+KEYS_TO_VALIDATE = ["runAsGroup", "runAsUser"]
+LOGGER = get_logger(name=__name__)
+
+
+@pytest.fixture(scope="class")
+def model_registry_scc_namespace(model_registry_namespace: Namespace):
```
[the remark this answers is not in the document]
dbasunag: I plan to do that soon. May be I should just start with scc one.

tests/model_registry/test_model_registry_scc.py
dbasunag: the order of namespace, updating dsc has to be specific. model_registry_resource needs it to pick specific kinds out of the namespace. This should stay.

tests/model_registry/test_model_registry_scc.py
dbasunag: I feel constants makes triaging a bit more time consuming. I would use them when I must. In this case, I would prefer to keep it this way.

tests/model_registry/test_model_registry_scc.py
```diff
@@ -0,0 +1,149 @@
+import pytest
+from typing import Self
```
fege, line 2: if we do not reach an agreement on bumping the python version we should use 
```
from __future__ import annotations 
from typing_extensions import Self
```

dbasunag: `from __future__ import annotations ` is not needed for 3.11 onwards, from what I remember. But I will add it, till version is bumped.

tests/model_registry/scc/test_model_registry_scc.py
```diff
@@ -0,0 +1,144 @@
...
+                    LOGGER.info(
+                        f"For container: {container.name}, {key} validation: {expected_value[key]} completed "
+                        "successfully"
+                    )
+                else:
+                    errors.append(
+                        f"For {container.name}, expected key {key} value: {expected_value[key]},"
+                        f" actual: {container.securityContext.get(key)}"
+                    )
+
```
rnetser, line 139: for clarity, this should also be extracted

tests/model_registry/scc/utils.py
```diff
@@ -0,0 +1,29 @@
...
+from typing import Any
+
+from ocp_resources.resource import NamespacedResource
+
+
+def get_uid_from_namespace(namespace_scc: dict[str, str]) -> str:
+    return namespace_scc["uid-range"].split("/")[0]
+
+
+def validate_pod_security_context(
```
mwaykole, line 11: add doc

dbasunag: I will add one liners.

tests/model_registry/scc/test_model_registry_scc.py
```diff
@@ -0,0 +1,144 @@
...
+            namespace_scc=model_registry_scc_namespace,
+            model_registry_resource=model_registry_resource,
+            ns_uid=ns_uid,
+        )
+        containers = model_registry_resource.instance.spec.containers
+        for container in containers:
+            expected_value = {
+                "runAsUser": int(ns_uid) + 1 if "sidecar" in container.args else int(ns_uid),
+                "runAsGroup": int(ns_uid) + 1 if "sidecar" in container.args else None,
+            }
```
lugi0, line 126: Are we sure these will always be the expected values? I thought it might be better to check that whatever value is present sits in the range defined in the namespace annotation

tests/model_registry/scc/utils.py
```diff
@@ -0,0 +1,29 @@
...
+    ns_uid: str,
+) -> list[str]:
+    errors = []
+    pod_selinux_option = pod_security_context.get("seLinuxOptions", {}).get("level")
+    if pod_selinux_option != namespace_scc["seLinuxOptions"]:
+        errors.append(
+            f"selinux option from pod {model_registry_resource.name} {pod_selinux_option},"
+            f" namespace: {namespace_scc['seLinuxOptions']}"
+        )
+    if pod_security_context.get("fsGroup") != int(ns_uid):
```
lugi0, line 24: Same as the other comment, uid is defined as a range, I would check that whatever id is assigned here falls within the range rather than expecting a specific one

lugi0 (commented):
In general LGTM, but I would rather check that the uid(s) fall within the given range rather than expect a specific uid, as that might lead to flakiness if openshift changes its behaviour under the hood

tests/model_registry/scc/test_model_registry_scc.py
dbasunag: OSSM explicitly sets runAsUser to be app_UID+1, and runAsGroup app_UID+1, https://github.com/maistra/istio/pull/1057. For other container, openshift sets it to app_UID. So this should not be flaky.

tests/model_registry/scc/test_model_registry_scc.py
```diff
@@ -0,0 +1,131 @@
+from __future__ import annotations
```
fege, line 1: can be removed now

tests/model_registry/scc/utils.py
```diff
@@ -0,0 +1,62 @@
+from __future__ import annotations
```
fege, line 1: same as above

tests/model_registry/test_model_registry_creation.py
```diff
@@ -1,3 +1,4 @@
+from __future__ import annotations
```
fege, line 1: same as above

tests/model_registry/scc/test_model_registry_scc.py
lugi0: Are you sure our uids are set by OSSM and not by OCP directly?

dbasunag: yes. Proxy one comes from OSSM.

fege (approved):
/lgtm

lugi0 (approved):

PR #229, opened 2025-04-08
feat: move to marking test branches

rnetser:
Instead of branching in the repo, tests can now be:
- marked as "upsteam", for upstream-only features
- marked as "odh", for mistream features
- marked with rhoai version on which they were developed, e.g `pytest.mark.rhoai_2_16`; only tests which meet the installed product version will be selected.


conftest.py
```diff
@@ -154,13 +158,22 @@ def _add_upgrade_test(_item: Item, _upgrade_deployment_modes: list[str]) -> bool
...
     upgrade_deployment_modes: list[str] = []
+    formatted_product_version: Version | None = None
 
     run_pre_upgrade_tests: str | None = config.getoption(name="pre_upgrade")
     run_post_upgrade_tests: str | None = config.getoption(name="post_upgrade")
     if config_upgrade_deployment_modes := config.getoption(name="upgrade_deployment_modes"):
         upgrade_deployment_modes = config_upgrade_deployment_modes.split(",")
 
+    if not session.config.option.setupplan and not session.config.option.collectonly:
+        if product_version := get_product_version(admin_client=get_client(), raise_on_missing_csv=False):
```
lugi0, line 169: Why don't you want to raise on missing csv?

lugi0 (commented):
this looks ok to me, but as far as I can see we're only filtering tests based on the version marker; how do we want to handle the midstream/upstream markers?

conftest.py
rnetser: because on upstream you will not have it;
i have another pr that i am working on for cluster sanity - there i will check if this is an odh/rhoai deployment and raise

pytest.ini
```diff
@@ -7,6 +7,14 @@ markers =
     polarion: Store polarion test ID
     jira: Store jira bug ID
 
+    # Distributions and supported downstream markers
+    upstream: Mark tests as ones suitable to run only upstream.
```
dbasunag, line 11: How would we handle when a feature behavior changes between releases and when test step changes between releases?

rnetser: how about adding the option to add an indicator like
`up_to_rhoai_2_16` and there will be another test with the new functionality which will have `rhoai_2_17` ?

adolfo-ab (dismissed):

pytest.ini
```diff
@@ -7,6 +7,14 @@ markers =
...
+    odh: Mark tests as ones suitable to run in ODH deployment (AKA `midstream`.
```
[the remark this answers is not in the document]
rnetser: done

pytest.ini
dbasunag: This would add duplicate tests and the repository can potentially get messy and hard to maintain. I don't like this idea.

rnetser:
addressed in https://github.com/opendatahub-io/opendatahub-tests/pull/231

PR #233, opened 2025-04-10
Add basic InferenceGraph deployment check

israel-hdez:
## Description
This adds a test that deploys an InferenceGraph (IG), sends an inference request to the IG and verifies that the request succeeds.

The deployed InferenceGraph is based on the example on the KServe documentation available in the following URL: https://kserve.github.io/website/0.15/modelserving/inference_graph/image_pipeline/. The example was adapted to run in openvino (which is a supported server in ODH), rather than TorchServe.

Related to: https://issues.redhat.com/browse/RHOAIENG-17958

## How Has This Been Tested?
Running `uv run pytest tests/model_serving/model_server/inference_graph/test_inference_graph_deployment.py`

## Merge criteria:


israel-hdez:
@rnetser So, I finally have this working, but it is unstable because of the modelcar. May you assist me on uploading the model to cloud storage?

tests/model_serving/model_server/inference_graph/conftest.py
```diff
@@ -0,0 +1,90 @@
...
+            ],
+        }
+    }
+    with InferenceGraph(
+        client=admin_client,
+        name="dog-breed-pipeline",
+        namespace=model_namespace.name,
+        nodes=nodes,
+    ) as inference_graph:
+        inference_graph.wait_for_condition(inference_graph.Condition.READY, "True")
```
rnetser, line 44: please use arg=val

tests/model_serving/model_server/inference_graph/conftest.py
```diff
@@ -0,0 +1,90 @@
...
+        time.sleep(5)  # This wait helps OpenShift to fully enable the Route, if the InferenceGraph is exposed
```
rnetser, line 45: iiuc the wait here for a route to be created? if so, please add wait for route using TimeoutSampler
using `sleep` is risky as different envs behave differently; if the route is not ready in expected time, this should fail.

utilities/inference_utils.py
```diff
@@ -71,7 +74,10 @@ def get_deployment_type(self) -> str:
         ):
             return deployment_type
 
-        return self.inference_service.instance.status.deploymentMode
+        if isinstance(self.inference_service, InferenceService):
+            return self.inference_service.instance.status.deploymentMode
+
+        return KServeDeploymentType.SERVERLESS
```
rnetser, line 80: can you please add here a check if `InferenceGraph` so it is clear that this is returned on InferenceGraph?
btw - why not get it from the InferenceGraph first isvc?

utilities/manifests/onnx.py
```diff
@@ -2,22 +2,26 @@
...
+                "use_regex": True
+            },
+            "infer-mnist": {
+                "query_input": "@utilities/manifests/openvino/mnist-input.json",
+                "query_output": r'{"model_name":"mnist-model__isvc-[0-9a-z]+","model_version":"1","outputs":\[{"name":"Plus214_Output_0","datatype":"FP32","shape":\[1,10\],"data":\[.*\]}\]}',
+                "use_regex": True
+            },
+            "graph": {
+                "query_input": "@utilities/manifests/openvino/dog-input-tensor.json",
+                "query_output": r'{"model_name":"dog-breed-classifier","model_version":"1","outputs":\[{"name":"495","shape":\[1,133\],"datatype":"FP32","data":\[.*\]}\]}',
```
rnetser, line 22: `dog-breed-classifier` should be replaced with $model_name

rnetser (changes requested):

tests/model_serving/model_server/inference_graph/conftest.py
israel-hdez: IIRC, the `Route` is created quickly, even before the previous line `wait_for_condition` would return.

The problem that I found was that after the IG is flagged as `Ready`, immediately sending a request to the IG endpoint results in the _Application not available_ page being returned. This small sleep is allowing the Route to settle down. 

Maybe, a better approach can be to probe the endpoint until the  _Application not available_ page is no longer returned???

tests/model_serving/model_server/inference_graph/conftest.py
israel-hdez: Fixed

utilities/inference_utils.py
israel-hdez: > btw - why not get it from the InferenceGraph first isvc?

The InferenceGraph has its own deployment mode. So, for example, an ISVC being in Serverless mode doesn't mean that the IG is also Serverless.

israel-hdez: > can you please add here a check if `InferenceGraph` so it is clear that this is returned on InferenceGraph?

Done!

utilities/manifests/onnx.py
israel-hdez: I changed it, but I haven't tried the change. Let me confirm it works as intended.

tests/model_serving/model_server/inference_graph/conftest.py
rnetser: we have a retry on `run_inference` but saw yesterday that sometimes the response is not an err (i.e in res), so i will work on fixing it.
better to have the check in one place, so the sleep here should be removed

utilities/inference_utils.py
rnetser: > The InferenceGraph has its own deployment mode
So we should get it from the InferenceGraph, right?

tests/model_serving/model_server/inference_graph/conftest.py
israel-hdez: Oh, I now see the `@retry` decorator on `run_inference`.

> but saw yesterday that sometimes the response is not an err (i.e in res)

I think the _Application is not available_ page would fall on this case. I tried curl on a dummy Route and the exit status of the `curl` command is zero. So, it won' be catched in `res`.

> better to have the check in one place, so the sleep here should be removed

OK, I support this.
I've removed the sleep and and modified run_inference to retry if the route is not ready yet. I hope the changes are acceptable.

utilities/inference_utils.py
israel-hdez: > > The InferenceGraph has its own deployment mode
> So we should get it from the InferenceGraph, right?

Yes. I think this is covered by lines 70-73, no? These `if` statements are a fallback in case the annotation is missing.

In contrast to ISVCs, we still don't have the deployment mode in the status of the InferenceGraph. So we can only rely in the annotation and fallback to some default if the annotation is not present.

Or am I missing something?

utilities/manifests/onnx.py
israel-hdez: I can confirm it works.

utilities/inference_utils.py
rnetser: so line 81 (default return) should be removed. we do not want to assume any deployment mode if we do not hit any of the conditions.
another option (and i can do it in a following pr) is to return the default mode from the card if we do not find any annotation/status. wdyt?

rnetser (approved):

mwaykole (approved):

PR #235, opened 2025-04-10
Add Cluster sanity checks before test execution

rnetser:
## Merge criteria:


utilities/infra.py
```diff
@@ -809,3 +824,66 @@ def wait_for_isvc_pods(client: DynamicClient, isvc: InferenceService, runtime_na
...
+        if request.session.config.getoption(skip_rhoai_check):
+            LOGGER.warning(f"Skipping RHOAI resource checks, got {skip_rhoai_check}")
+
+        else:
+            assert_dsci_ready(dsci_resource=dsci_resource)
+            assert_dsc_ready(dsc_resource=dsc_resource)
+
+    except (ResourceNotReadyError, NodeUnschedulableError, NodeNotReadyError) as ex:
+        error_msg = f"Cluster sanity check failed: {str(ex)}"
+        return_code = 99
```
lugi0, line 881: Why 99? 😄

utilities/infra.py
```diff
@@ -809,3 +824,66 @@ def wait_for_isvc_pods(client: DynamicClient, isvc: InferenceService, runtime_na
     """
     LOGGER.info("Waiting for pods to be created")
     return get_pods_by_isvc_label(client=client, isvc=isvc, runtime_name=runtime_name)
+
+
+def assert_dsci_ready(dsci_resource: DSCInitialization) -> None:
```
dbasunag, line 829: Can we rename? We are not asserting here.

utilities/infra.py
```diff
@@ -809,3 +824,66 @@ def wait_for_isvc_pods(client: DynamicClient, isvc: InferenceService, runtime_na
...
+    LOGGER.info(f"Verify DSCI {dsci_resource.name} are {dsci_resource.Status.READY}.")
+    if dsci_resource.status != dsci_resource.Status.READY:
+        raise ResourceNotReadyError(f"DSCI {dsci_resource.name} is not ready.\nStatus: {dsci_resource.instance.status}")
+
+
+def assert_dsc_ready(dsc_resource: DataScienceCluster) -> None:
+    LOGGER.info(f"Verify DSC {dsc_resource.name} are {dsc_resource.Status.READY}.")
+    if dsc_resource.status != dsc_resource.Status.READY:
+        raise ResourceNotReadyError(f"DSCI {dsc_resource.name} is not ready.\nStatus: {dsc_resource.instance.status}")
```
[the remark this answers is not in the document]
rnetser: fixed

tests/conftest.py
```diff
@@ -446,3 +459,33 @@ def minio_data_connection(
...
+
+
+@pytest.fixture(scope="session")
+def junitxml_plugin(
+    request: FixtureRequest, record_testsuite_property: Callable[[str, object], None]
+) -> Callable[[str, object], None] | None:
+    return record_testsuite_property if request.config.pluginmanager.has_plugin("junitxml") else None
+
+
+@pytest.fixture(scope="session", autouse=True)
```
dbasunag, line 476: Can we please add a module scoped one too?

utilities/infra.py
rnetser: to be far from https://docs.pytest.org/en/stable/reference/exit-codes.html :)

utilities/infra.py
```diff
@@ -809,3 +824,66 @@ def wait_for_isvc_pods(client: DynamicClient, isvc: InferenceService, runtime_na
...
+
+def assert_dsci_ready(dsci_resource: DSCInitialization) -> None:
+    LOGGER.info(f"Verify DSCI {dsci_resource.name} are {dsci_resource.Status.READY}.")
+    if dsci_resource.status != dsci_resource.Status.READY:
+        raise ResourceNotReadyError(f"DSCI {dsci_resource.name} is not ready.\nStatus: {dsci_resource.instance.status}")
+
+
+def assert_dsc_ready(dsc_resource: DataScienceCluster) -> None:
+    LOGGER.info(f"Verify DSC {dsc_resource.name} are {dsc_resource.Status.READY}.")
+    if dsc_resource.status != dsc_resource.Status.READY:
```
dbasunag, line 837: Instead of comparing and failing, we should consider waiting for the status to be ready for the time being.

utilities/infra.py
```diff
@@ -809,3 +824,66 @@ def wait_for_isvc_pods(client: DynamicClient, isvc: InferenceService, runtime_na
...
+
+    if request.session.config.getoption(skip_cluster_sanity_check):
+        LOGGER.warning(f"Skipping cluster sanity check, got {skip_cluster_sanity_check}")
+        return
+
+    try:
+        LOGGER.info("Check cluster sanity.")
+
+        assert_nodes_in_healthy_condition(nodes=nodes, healthy_node_condition_type={"KubeletReady": "True"})
+        assert_nodes_schedulable(nodes=nodes)
```
dbasunag, line 870: How about operator pods status check (applications_namespace)? Can be a follow up PR.

tests/conftest.py
```diff
@@ -279,11 +282,21 @@ def unprivileged_client(
             raise ClusterLoginError(user=non_admin_user_name)
 
 
+@pytest.fixture(scope="session")
+def dsci_resource(admin_client: DynamicClient) -> DSCInitialization:
+    name = py_config["dsci_name"]
+    for dsci in DSCInitialization.get(dyn_client=admin_client, name=name):
```
dbasunag, line 288: If we already have the name, then why do we need to make get() call. We can create and use the python object here.

dbasunag (commented):
A few open comments are present.

tests/conftest.py
rnetser: done

tests/conftest.py
rnetser: why?

utilities/infra.py
rnetser: the cluster is expected to be installed with the required product. the installation is not controlled by the test.

utilities/infra.py
rnetser: if dsc is ready it means the relevant pods are ready

tests/conftest.py
dbasunag: We update dsc between tests, we should have a sanity check between modules, to ensure the tests are not leaving the clusters unstable.

utilities/infra.py
dbasunag: The sanity would need to be extended for module scope at some point. For session scoped check, cluster would be installed with required product, but between tests we modify dsc and after teardown, it would need to get back to ready state.

utilities/infra.py
dbasunag: This would always account for any pods that are supposed to be cleaned up too?

rnetser: > any pods that are supposed to be cleaned up too
can you please explain?

tests/conftest.py
rnetser: so it should be part of restore

utilities/infra.py
dbasunag: If I enable MR, model registry operator pods would be created, and when I turn it off, they should be be cleaned up too. Would dsc not be ready till MR operator pod is cleaned up in this case?

utilities/infra.py
rnetser: see comment above

utilities/infra.py
rnetser: for clean up, not sure. this should be tested by the platform team and not verified in cluster sanity.
when enabling component then yes

tests/conftest.py
dbasunag: between tests -no, but between module runs, it might be a good idea to do it. Restore would fail only if the operation itself failed. Some basic checks are needed between module runs.

utilities/infra.py
dbasunag: Same response as above. We should have a module scoped cluster sanity, with all the edits we do in tests/modules.

utilities/infra.py
dbasunag: Components should do basic validations from component's perspective. Otherwise the node checks does not make sense either. Those are responsibilities of OCP team :)

tests/conftest.py
```diff
@@ -279,11 +282,23 @@ def unprivileged_client(
             raise ClusterLoginError(user=non_admin_user_name)
 
 
+@pytest.fixture(scope="session")
+def dsci_resource(admin_client: DynamicClient) -> DSCInitialization:
+    name = py_config["dsci_name"]
+    dsci = DSCInitialization(client=admin_client, name=name)
```
adolfo-ab, line 288: Can we do this instead?
`return DSCInitialization(client=admin_client, name=name, ensure_exists=True)`

`ensure_exists` will raise for us

tests/conftest.py
```diff
@@ -279,11 +282,23 @@ def unprivileged_client(
...
+
 @pytest.fixture(scope="session")
 def dsc_resource(admin_client: DynamicClient) -> DataScienceCluster:
     name = py_config["dsc_name"]
-    for dsc in DataScienceCluster.get(dyn_client=admin_client, name=name):
+    dsc = DataScienceCluster(client=admin_client, name=name)
+    if dsc.exists:
         return dsc
+
     raise ResourceNotFoundError(f"DSC resource {name} not found")
```
adolfo-ab, line 302: Same as above

utilities/infra.py
adolfo-ab: I think I agree with @rnetser on this, I think that most modules will only work on specific namespaces and won't affect the cluster resources, DSC or DSCI, so running sanity for every module might add constant overhead for a scenario that is not so frequent.

utilities/infra.py
```diff
@@ -809,3 +824,66 @@ def wait_for_isvc_pods(client: DynamicClient, isvc: InferenceService, runtime_na
...
+        raise ResourceNotReadyError(f"DSCI {dsci_resource.name} is not ready.\nStatus: {dsci_resource.instance.status}")
+
+
+def verify_dsc_status_ready(dsc_resource: DataScienceCluster) -> None:
+    LOGGER.info(f"Verify DSC {dsc_resource.name} are {dsc_resource.Status.READY}.")
+    if dsc_resource.status != dsc_resource.Status.READY:
+        raise ResourceNotReadyError(f"DSC {dsc_resource.name} is not ready.\nStatus: {dsc_resource.instance.status}")
+
+
+def cluster_sanity(
```
adolfo-ab, line 841: Maybe `check_cluster_sanity`

utilities/infra.py
```diff
@@ -809,3 +824,66 @@ def wait_for_isvc_pods(client: DynamicClient, isvc: InferenceService, runtime_na
...
+    request: FixtureRequest,
+    nodes: list[Node],
+    dsci_resource: DSCInitialization,
+    dsc_resource: DataScienceCluster,
+    junitxml_property: Callable[[str, object], None] | None = None,
+) -> None:
+    """
+    Check cluster resources: nodes, DSCI, DSC and exits pytest execution on failure.
```
adolfo-ab, line 849: Maybe `Check that cluster resources (Nodes, DSCI, DSC) are healthy and exists pytest execution on failure.`

utilities/infra.py
adolfo-ab: I think this is a good start, we can add more checks in the future if we find out we are missing something

tests/conftest.py
rnetser: done

tests/conftest.py
rnetser: done

utilities/infra.py
rnetser: renamed to `verify_cluster_sanity`

utilities/infra.py
rnetser: done

utilities/infra.py
rnetser: @dbasunag these are 2 different things.
cluster sanity - make sure resources report that they are healthy. 
if you remove a component from dsc -> this is not part of cluster sanity.
you need to make sure it is restored correctly in teardown.

utilities/infra.py
rnetser: done

utilities/infra.py
rnetser: resolving for now

adolfo-ab (approved):

dbasunag (dismissed):

PR #237, opened 2025-04-10
Fix verify_no_inference_pods Utility Function

brettmthompson:
## Description
In its current state the `verify_no_inference_service` function will return None every time it executes without the `TimeSampler` ever calling the `get_pods_by_isvc_label` function.

Listed below are the issues identified with this function:
1. The `get_pods_by_isvc_label` method will actually never be triggered. This is because the `TimeSampler` is a python asynchronous iterator, and if we don't iterate over it then it will actually never yield a result by calling the assigned func.
2. If we address [1] above and `get_pods_by_isvc_label` actually does get called, the check `if not pods:` will never be true because the `get_pods_by_isvc_label` function does not return an empty list when no pods are found, it instead raises a `ResourceNotFoundError `. To properly handle this we should not include the `ResourceNotFound` error in the `TimeoutSampler`'s exception_dict, this will cause the `TimeSampler` to immediately raise a `TimeoutExpiredError` if the `get_pods_by_isvc_label` function raises a `ResourceNotFoundError`. We can then check for the `TimeoutExpiredError`'s last exception type in the except block, to determine if we are in a scenario where no pods exist.

## How Has This Been Tested?
Execute `serverless-scale-zero` e2e test
## Merge criteria:


tests/model_serving/model_server/serverless/utils.py
```diff
@@ -32,17 +34,20 @@ def verify_no_inference_pods(client: DynamicClient, isvc: InferenceService) -> N
...
-        )
-        if not pods:
-            return
+        ):
+            if not pods:
+                return
 
-    except TimeoutError:
+    except TimeoutExpiredError as e:
+        if isinstance(e.last_exp, ResourceNotFoundError):
```
dbasunag, line 49: Why not add ResourceNotFoundError to exceptions_dict?

rnetser (approved):

tests/model_serving/model_server/serverless/utils.py
brettmthompson: The `TimeoutSampler` exception_dict includes exceptions that will be [ignored](https://github.com/RedHatQE/timeout-sampler/blob/main/timeout_sampler/__init__.py#L80) if raised. In this case, we do not want to ignore the ResourceNotFoundError, but rather handle it gracefully when it is raised to verify that no pods exist.

lugi0 (approved):

dbasunag (approved):

PR #250, opened 2025-04-15
feat: Add upgrade tests for TrustyAIService

adolfo-ab:
## Description
In these tests we set up a model and a TrustyAIService with inference and training data, and schedule a drift metric before upgrade. After upgrade, we retrieve the pre-upgrade metric and delete them.

## How Has This Been Tested?
Running the tests in a working cluster.

## Merge criteria:


tests/conftest.py
```diff
@@ -83,13 +83,29 @@ def current_client_token(admin_client: DynamicClient) -> str:
...
 
-    with create_ns(admin_client=admin_client, pytest_request=request) as ns:
+    ns = Namespace(client=admin_client, name=request.param["name"])
+
+    if pytestconfig.option.post_upgrade:
         yield ns
+        ns.clean_up()
+    else:
+        with create_ns(admin_client=admin_client, pytest_request=request, teardown=teardown_resources) as ns:
+            yield ns
```
lugi0, line 108: isn't `ns` already defined in L 101? does this work?

adolfo-ab: Yes, the flow is:

1. If we are in post-upgrade, we get a reference to the already existing namespace and yield it
2. If we are not in post-upgrade, we override ns (`with create_ns(...) as ns:`) and yield it as before

tests/model_explainability/trustyai_service/conftest.py
```diff
@@ -32,28 +40,45 @@
...
+            storage={"format": "PVC", "folder": "/inputs", "size": "1Gi"},
+            data={"filename": "data.csv", "format": "CSV"},
+            metrics={"schedule": "5s"},
+            teardown=teardown_resources,
+        ) as trustyai_service:
+            trustyai_deployment = Deployment(
+                namespace=model_namespace.name, name=TRUSTYAI_SERVICE_NAME, wait_for_resource=True
+            )
+            trustyai_deployment.wait_for_replicas()
+            yield trustyai_service
```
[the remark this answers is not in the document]
adolfo-ab: No, because I am using the `teardown_resources` fixture (L59), which bypasses teardown when we are running pre-upgrade specifically. In all other scenarios we will do teardown as before.

tests/conftest.py
```diff
@@ -83,13 +83,29 @@ def current_client_token(admin_client: DynamicClient) -> str:
...
 @pytest.fixture(scope="class")
-def model_namespace(request: FixtureRequest, admin_client: DynamicClient) -> Generator[Namespace, Any, Any]:
+def model_namespace(
+    request: FixtureRequest, pytestconfig: pytest.Config, admin_client: DynamicClient, teardown_resources: bool
+) -> Generator[Namespace, Any, Any]:
     if request.param.get("modelmesh-enabled"):
         request.getfixturevalue(argname="enabled_modelmesh_in_dsc")
 
-    with create_ns(admin_client=admin_client, pytest_request=request) as ns:
+    ns = Namespace(client=admin_client, name=request.param["name"])
```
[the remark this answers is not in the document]
rnetser: can you please explain? resources will be in a component upgrade-dedicated ns

tests/model_explainability/trustyai_service/conftest.py
```diff
@@ -216,3 +240,105 @@ def trustyai_db_ca_secret(
...
+@pytest.fixture(scope="class")
+def mlserver_runtime(
+    pytestconfig: pytest.Config,
+    admin_client: DynamicClient,
+    minio_data_connection: Secret,
+    model_namespace: Namespace,
+    teardown_resources: bool,
+) -> Generator[ServingRuntime, Any, Any]:
+    name = "kserve-mlserver"
+    serving_runtime = ServingRuntime(client=admin_client, namespace=model_namespace.name, name=name)
```
rnetser, line 254: can be saved as kwargs and re-used

tests/model_explainability/trustyai_service/conftest.py
```diff
@@ -216,3 +240,105 @@ def trustyai_db_ca_secret(
...
+    model_namespace: Namespace,
+    minio_pod: Pod,
+    minio_service: Service,
+    minio_data_connection: Secret,
+    mlserver_runtime: ServingRuntime,
+    trustyai_service_with_pvc_storage: TrustyAIService,
+    teardown_resources: bool,
+) -> Generator[InferenceService, Any, Any]:
+    name = ("gaussian-credit-model",)
+    isvc = InferenceService(client=admin_client, namespace=model_namespace.name, name=name)
```
rnetser, line 318: ditto

tests/model_explainability/trustyai_service/trustyai_service_utils.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
rnetser, line 1: should be removed, we now use python3.12

tests/model_explainability/trustyai_service/upgrade/test_trustyai_service_upgrade.py
```diff
@@ -0,0 +1,109 @@
...
+
+    @pytest.mark.pre_upgrade
+    def test_trustyai_service_pre_upgrade_data_upload(
+        self,
+        admin_client,
+        minio_data_connection,
+        current_client_token,
+        trustyai_service_with_pvc_storage,
+    ) -> None:
+        """"""
```
rnetser, line 55: missing docstring

tests/model_explainability/trustyai_service/upgrade/test_trustyai_service_upgrade.py
```diff
@@ -0,0 +1,109 @@
...
+
+    @pytest.mark.pre_upgrade
+    def test_drift_metric_schedule_meanshift(
+        self,
+        admin_client,
+        current_client_token,
+        trustyai_service_with_pvc_storage,
+        gaussian_credit_model,
+    ):
+        verify_trustyai_service_metric_scheduling_request(
```
rnetser, line 71: ditto

tests/model_explainability/trustyai_service/upgrade/test_trustyai_service_upgrade.py
```diff
@@ -0,0 +1,109 @@
...
+            MinIo.PodConfig.MODEL_MESH_MINIO_CONFIG,
+            {"bucket": MinIo.Buckets.MODELMESH_EXAMPLE_MODELS},
+        )
+    ],
+    indirect=True,
+)
+@pytest.mark.usefixtures("minio_pod")
+class TestPostUpgradeTrustyAIService:
+    @pytest.mark.post_upgrade
+    def test_drift_metric_delete(
```
rnetser, line 97: shouldn't there be a matching post-upgrade test for each one of the tests above? how about tests which verify resources exist after upgrade?

rnetser (changes requested):

tests/model_explainability/trustyai_service/upgrade/test_trustyai_service_upgrade.py
adolfo-ab: The flow I'm interested in here is the following:
Pre-upgrade -> Set up model and TrustyAIService, and schedule a metric
Post-upgrade -> Check that the metric is still there untouched and can be used (in this case by deleting it).

So in this case we don't need to do the same things before and after upgrade, and that could be achieved by running smoke before and after upgrade anyways.

tests/conftest.py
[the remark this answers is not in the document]
adolfo-ab: Hi @dbasunag, I don't see the point of creating new fixtures that would do exactly the same. As @rnetser says, everything will be put into a different, upgrade-specific namespace, so there shouldn't be any problem.

tests/model_explainability/trustyai_service/conftest.py
```diff
@@ -32,28 +40,44 @@
...
-            namespace=model_namespace.name, name=TRUSTYAI_SERVICE_NAME, wait_for_resource=True
-        )
-        trustyai_deployment.wait_for_replicas()
+    trustyai_service_kwargs = {"client": admin_client, "namespace": model_namespace.name, "name": TRUSTYAI_SERVICE_NAME}
+    trustyai_service = TrustyAIService(**trustyai_service_kwargs)
+
+    if pytestconfig.option.post_upgrade:
         yield trustyai_service
+        trustyai_service.clean_up()
+
```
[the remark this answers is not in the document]
adolfo-ab: @dbasunag , as mentioned in the other comment, I don't see how since upgrade tests will happen on its own ns.

tests/model_explainability/trustyai_service/conftest.py
adolfo-ab: done

tests/model_explainability/trustyai_service/conftest.py
adolfo-ab: done

tests/model_explainability/trustyai_service/trustyai_service_utils.py
adolfo-ab: done

tests/model_explainability/trustyai_service/upgrade/test_trustyai_service_upgrade.py
adolfo-ab: done

tests/conftest.py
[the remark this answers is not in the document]
dbasunag: Doing different things based on different conditions in a fixture creates messy and hard to maintain code. The resources would be created in the same named namespace. So if you are running upgrade and then say running smoke on an upgraded cluster, you are potentially going to encounter issues if resources conflicts. Also If you are using the same fixtures in upgrade and non-upgrade scenarios, changes to them would needs to be tested with and without upgrade depending upon the complexity of changes. Instead, separating them creates less messy code with clear boundaries. Fixture codes are ok to be not reused and are not considered duplicate codes.

tests/model_explainability/trustyai_service/conftest.py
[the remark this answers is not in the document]
dbasunag: If we ever plan to expand and run post-upgrade tests, this can lead to conflicts with resources. For ease of triage, we should separate resource namespaces.

tests/conftest.py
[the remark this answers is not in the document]
dbasunag: Namespace and anything created in it would be resources. If upgrade and non-upgrade flow uses the same resources name, subsequent post upgrade runs can encounter conflicts due to failing to clean up and make triage difficult, no? However, this is @adolfo-ab 's team code and they should be able to make this call about their design. I will not hold this PR, if they find this better.

dbasunag (approved):

PR #248, opened 2025-04-14
add labeler to add labels to prs based on areas impacted

dbasunag:
## Merge criteria:

- [ ] The commits are squashed in a cohesive manner and have meaningful messages.
- [ ] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [ ] The developer has manually tested the changes and verified that the changes work

PR #251, opened 2025-04-15
on rebase clean commented-by- labels

dbasunag:
## Merge criteria:


lugi0:
why do we want to remove the commented-by labels?

dbasunag:
> why do we want to remove the commented-by labels?
Synchronize event is triggered when a new commit is pushed. 
```
A pull request's head branch was updated. For example, the head branch was updated from the base branch or new commits were pushed to the head branch.
```
When this happened, a comment could be resolved. So keeping the label would be not needed. If a comment is not addressed, a reviewer should then indicate that in further review. e.g. https://github.com/opendatahub-io/opendatahub-tests/pull/248. The comment was addressed, but commented-by label is still present, which is wrong.

lugi0 (approved):

PR #257, opened 2025-04-17
feat: fail on missing operators

rnetser:
## Merge criteria:


lugi0 (dismissed):

jiridanek (approved):

mwaykole (approved):

docs/GETTING_STARTED.md
```diff
@@ -67,6 +67,13 @@ By default, cluster sanity checks are run to make cluster ready for tests.
 To skip cluster sanity checks, pass `--cluster-sanity-skip-check` to skip all tests.
 To skip RHOAI/ODH-related tests (for example when running in upstream), pass `--cluster-sanity-skip-rhoai-check`.
 
+### Check dependent operators
+By default, `Serveless` (serverless-operator), `Authorino` (authorino-operator) and `Service Mesh` (servicemeshoperator) operators are checked to be installed when  
+running model registry or model server Serverless tests.
+To check only specific operator, pass `--tc=dependent_operators:<operator_name>` to pytest.
+For example, to check only `Serveless` and `Service Mesh` operators, pass `--tc=dependent_operators:serverless-operator,servicemeshoperator`.
```
mwaykole, line 74: nice so it wil not affect Raw kserve 
if no dependent operator is installed

lugi0 (dismissed):

conftest.py
```diff
@@ -223,17 +223,17 @@ def pytest_runtest_setup(item: Item) -> None:
     """
     Performs the following actions:
     1. Updates global config (`updated_global_config`)
-    2. Adds skip fixture for kserve if serverless or authorino operators are not installed.
-    3. Adds skip fixture for serverless if authorino/serverless/service mesh are not deployed.
+    2. Adds `fail_if_missing_dependant_operators` fixture for Serverless and model registry tests.
```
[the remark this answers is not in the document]
lugi0: both spellings are correct FWIW :)

docs/GETTING_STARTED.md
rnetser: yes

tests/conftest.py
```diff
@@ -478,3 +479,27 @@ def cluster_sanity_scope_session(
...
     )
+
+
+@pytest.fixture(scope="session")
+def fail_if_missing_dependent_operators(admin_client: DynamicClient) -> None:
+    missing_operators: list[str] = []
+
+    for operator_name in py_config.get("dependent_operators", []).split(","):
+        LOGGER.info(f"Verifying if {operator_name} is installed")
+        for csv in ClusterServiceVersion.get(
```
dbasunag, line 490: you are making the same api call in the for loop. You might want to save the CSVs outside the for loop and then traverse through the list inside.

tests/conftest.py
```diff
@@ -478,3 +479,27 @@ def cluster_sanity_scope_session(
         dsci_resource=dsci_resource,
         junitxml_property=junitxml_plugin,
     )
+
+
+@pytest.fixture(scope="session")
+def fail_if_missing_dependent_operators(admin_client: DynamicClient) -> None:
+    missing_operators: list[str] = []
+
+    for operator_name in py_config.get("dependent_operators", []).split(","):
```
dbasunag, line 488: I think this can be simplified with all() and comprehension. Wdyt?

rnetser: what do you mean?

tests/conftest.py
rnetser: done

tests/conftest.py
dbasunag: Actually instead of all(). something like this would be simpler. in global config you can have a list instead of a string:
```
 operators = ["servicemeshoperator","authorino-operator","serverless-operator"]
```
then this fixture can check:
```
csvs = ClusterServiceVersion.get(dyn_client= admin_client, namespace=py_config["applications_namespace"])
matched_csvs = [csv.name for csv in csvs if csv.name.startswith(tuple(operators))]
if len(matched_csvs) != len(operators):
     pytest.fail(...)
```
What do you think?

rnetser: it is a string in global_config so it is easy to overwrite from command line.

rnetser: added a more explicit failure to each kind of failure

dbasunag: you can take a string and convert it here to a list and use here. The number of lines can be then significantly reduced 4 vs roughly 25? Also, here our goal should be checking that the csv exists. CSV can momentarily be in not SUCCEEDED state (Installing etc. and if we just wait, we would recover), that does not necessarily mean we can't proceed further. So we should not fail on that here.

rnetser: i disagree; these are dependent operators that we depend on; cluster should be in a state where they are all ready.
even if you do wait, you'd want a different error if not installed or not ready

dbasunag: Then we must wait, otherwise, this would cause flakiness without any product bug.

rnetser: wait for the dependent operators are not ready - this is not part of setup; this is part of cluster setup.
if they are not ready, setup will fail.

dbasunag: You can't predict when the CSV would be in succeeded state. This may not necessarily be a bug if it is in even `failing` state momentarily after deployment. https://downey.io/blog/desired-state-vs-actual-state-in-kubernetes/#:~:text=Eventual%20consistency%20means%20that%20if,current%20update%20and%20be%20Consistent. Our concern should be 1) if the operator is installed (then CSV would be present) 2) if the CSV reaches consistent state eventually. So checking for SUCCEEDED state and failing if we find something else for a moment is not a good idea. If you want to check the status add a wait for call. I have worked on something similar: https://github.com/RedHatQE/openshift-virtualization-tests/blob/main/tests/after_cluster_deploy_sanity/test_after_cluster_deploy_sanity.py#L162, so, telling you this based on my findings at that point.

rnetser: https://github.com/RedHatQE/openshift-virtualization-tests/blob/main/tests/after_cluster_deploy_sanity/test_after_cluster_deploy_sanity.py#L162 - is thesting the product under the test in the repo and not other operators.

dbasunag: True. But the logic for CSV remains the same. CSV not being in SUCCEEDED state could be transient. We should not fail.

mwaykole (approved):
/lgtm

tests/conftest.py
rnetser: if they are `progressing/degraded/failed` then the tests should fail.
i am strongly against adding wait; I can remove the status check but i think this is not the right approach.

rnetser: As mentioned, i do not think this is the right approach.
I am moving the check to model server.
MR can decide to have different checks and address in a different pr

mwaykole (approved):

adolfo-ab (approved):

PR #269, opened 2025-04-25
Fixing TestServerlessScaleToZero test

brettmthompson:
## Description
Currently the `TestServerlessScaleToZero` test will fail at the final step `test_serverless_pods_after_scale_to_one_replica`. The reason for this failure is because this step immediately retrieves the deployments after the infernce service is edited, not giving enough time for the new deployment to be created. So when the deployments are retrieved only the first 2 will be returned, causing the label check for `serving.knative.dev/configurationGeneration=3` to never return true.

To resolve this issue, I have updated the  `wait_for_inference_deployments_replicas` function in the following ways:

1. Wrapping call to Deployment.Get() with `TimeoutSampler`. This allows time for deployments to be created. The same `ResourceNotUniqueError` or `ResourceNotFoundError` exceptions will be raised if an unexpected number of deployments is returned after timeout, with a bit more detail in the error message.
2. Using the `TimeoutWatch` class to distribute the timeout across the sub functions called.
3. Previously when looping over the deployments returned, if any of the deployments have Raw deployment mode, the `wait_for_replicas_in_deployment` method would be called, but only the first element from the list of deployments would ever be passed to this function, regardless of which deployment we were currently iterating over in the list. I have updated this to pass the current deployment being iterated over.
4. Introduced a `labels` string input parameter that allows for deployments to be filtered on custom labels in addition to the inference service label.
5. Introduced a `deployed` boolean input parameter. This allows for deployments that are expected to have 0 replicas to be checked using this method.

**Note**
I am using the `TimeoutSampler` class here and not the `retry` decorator for the following reasons:

1. Allow for custom timeout value to be passed rather than hard coding it in the decorator. 
2. Allow for the consistency of timeout by distrubuting the remaining timeout window across any sub functions called using the `TimeoutWatch` class
3. Allow for custom top level exceptions to be thrown by this class (`ResourceNotUniqueError` and `ResourceNotFoundError`). If a retry decorator is used, the top level exception raised will always be of type `TimeoutExpiredError`

## How Has This Been Tested?

Ran the `TestServerlessScaleToZero` locally and it is now successful.

## Merge criteria:

## Summary by Sourcery

Fix flaky `TestServerlessScaleToZero` test by waiting for the deployment to be ready after scaling.

Tests:
- Introduce `wait_for_inference_deployments_replicas` utility function to poll for expected deployments within a timeout.
- Update `test_serverless_pods_after_scale_to_one_replica`wait_for_inference_deployments_replicas` before checking deployment state.

## Summary by CodeRabbit

- **Tests**
	- Improved reliability of serverless scaling tests by centralizing deployment waiting and validation logic.
- **Refactor**
	- Enhanced deployment waiting utilities with configurable timeouts, flexible label filtering, and improved error handling for deployment readiness checks.
- **Chores**
	- Added a new exception for handling unexpected resource counts in API responses.


tests/model_serving/model_server/serverless/utils.py
```diff
@@ -50,6 +51,56 @@ def verify_no_inference_pods(client: DynamicClient, isvc: InferenceService) -> N
         raise
 
 
+def wait_for_deployments(
```
dbasunag, line 54: you can use `@retry` decorator from [timeout sampler](https://github.com/RedHatQE/timeout-sampler) here. It would be simpler.

rnetser, line 54: why can't `utilities.infra.wait_for_inference_deployment_replicas` be used?

brettmthompson: Updated the PR description to describe the new changes.

tests/model_serving/model_server/serverless/test_scale_to_zero.py
```diff
@@ -95,14 +94,9 @@ def test_no_serverless_pods_when_no_traffic(self, admin_client, inference_servic
...
-            if deployment.labels["serving.knative.dev/configurationGeneration"] == "3":
-                deployment.wait_for_replicas()
-                return
-
-        raise DeploymentValidationError(
-            f"Inference Service {inference_service_patched_replicas.name} new deployment not found"
+            isvc=inference_service_patched_replicas,
+            expected_num_deployments=1,
+            labels="serving.knative.dev/configurationGeneration=3",
         )
```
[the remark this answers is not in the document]
brettmthompson: I am not sure how I feel about this suggested change. I do not see a problem hard coding the generation given that this is a test so we are checking for the desired state of a resource based on the behavior as defined in the test. 

In the test we have updated the inference service twice, so we should be checking for the generation to be 3. This hard coded check is a condition for the test to pass successfully, i.e. if the generation is not 3 then the inference service is in an unexpected state.

Please let me know if others think differently.

tests/model_serving/model_server/serverless/test_scale_to_zero.py
```diff
@@ -95,14 +94,9 @@ def test_no_serverless_pods_when_no_traffic(self, unprivileged_client, inference
     @pytest.mark.order(5)
     def test_serverless_pods_after_scale_to_one_replica(self, unprivileged_client, inference_service_patched_replicas):
         """Verify pod is running after scaling to 1 replica"""
-        for deployment in Deployment.get(
+        wait_for_inference_deployment_replicas(
```
mwaykole, line 97: hey @brettmthompson why not  using `timeout` of func wait_for_inference_deployment_replicas?

brettmthompson: Hi Milind, if you are referring to why I am not passing a `timeout` value when calling the `wait_for_inference_deployment_replicas` here, it is because I am just using the default value of 5 minutes.

tests/model_serving/model_server/serverless/test_scale_to_zero.py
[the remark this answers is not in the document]
Jooho: >In the test we have updated the inference service twice, so we should be checking for the generation to be 3. This hard coded check is a condition for the test to pass successfully,

Thanks for the explanation! I reviewed the code and had a couple of clarifying questions:

You mentioned that the label is added to help locate the correct deployment. I was wondering—since there's only one deployment, is the label strictly necessary in this case? I understand that there may be multiple ReplicaSets over time, but the lookup seems to focus on the deployment itself. [code](https://github.com/opendatahub-io/opendatahub-tests/pull/269/files#diff-e2ff6db074d9c48c5dff7b9542d476ac8e50d300dd5614fc2abd7cc5d09e7435R232)

> i.e. if the generation is not 3 then the inference service is in an unexpected state.

so I don't get this comment either.

tests/model_serving/model_server/serverless/test_scale_to_zero.py
Jooho: the default value is 2 mins so I think it will use 2 min default timeout.

tests/model_serving/model_server/serverless/test_scale_to_zero.py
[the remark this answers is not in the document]
brettmthompson: @Jooho when running the test, we can see that a new deployment is created for each update made to the ISVC
Screenshot From 2025-04-29 10-37-36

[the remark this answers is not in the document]
Jooho: Oh got it. It is serverless and ksvc(revision) creates each deployment. ok sound good to me

mwaykole (dismissed):

utilities/infra.py
```diff
@@ -196,49 +200,75 @@ def wait_for_inference_deployment_replicas(
...
+        ResourceNotFoundError: If a less number of deployments exist than expected after timeout.
+
     """
+    timeout_watcher = TimeoutWatch(timeout=timeout)
     ns = isvc.namespace
     label_selector = utilities.general.create_isvc_label_selector_str(
         isvc=isvc, resource_type="deployment", runtime_name=runtime_name
     )
+    if labels:
+        label_selector += f",{labels}"
```
dbasunag, line 223: Not needed, you can directly pass the string to line 232 as `label_selector=labels if labels else None`

utilities/infra.py
```diff
@@ -196,49 +200,75 @@ def wait_for_inference_deployment_replicas(
...
 
-    deployments = list(
-        Deployment.get(
+    deployment_list = []
+    try:
+        for deployments in TimeoutSampler(
+            wait_timeout=timeout_watcher.remaining_time(),
```
dbasunag, line 228: I don't believe you need this. You instead need to pass timeout here.

utilities/infra.py
```diff
@@ -196,49 +200,75 @@ def wait_for_inference_deployment_replicas(
...
-                    wait_for_replicas_in_deployment(
-                        deployment=deployments[0],
-                        replicas=isvc.instance.spec.predictor.get("minReplicas", 1),
-                    )
-
-                deployment.wait_for_replicas(timeout=timeout)
-
-        return deployments
+    for deployment in deployment_list:
+        if deployment.exists:
```
dbasunag, line 256: if the deployment does not exist at this point, you should fail, right? Because you counted it towards expected number of deployment in sampler.

utilities/infra.py
brettmthompson: Here I am combining the `labels` passed as input with the result of the `utilities.general.create_isvc_label_selector_str` function. If we just used `label_selector=labels if labels else None` then we would not include the result of that function call in the label selectors. 

@dbasunag if you think it is cleaner I can remove lines 222-223 and update line 232 to be:
`lael_selector=f"{label_selector},{labels}" if labels else label_selector`

utilities/infra.py
brettmthompson: You are correct that the result of `timeout_watcher.remaining_time()` will be equal to `timeout` at this step. I used `timeout_watcher.remaining_time()` here to be consistent with how the timeout value is passed to the other function calls and since `.remaining_time()` is a low cost function call.

dbasunag: Timeout sampler is already handling that, no?

utilities/infra.py
dbasunag: > `f"{label_selector},{labels}" ` 

you only need to pass labels or None to label_selector. Can you please check how wrapper implemented label selector?

utilities/infra.py
brettmthompson: The previous logic did a similar check, first checking the expected number of deployments were returned, then looping through and checking that each deployment exists. I tried to keep that consistent in the new logic as well, but I may be misunderstanding what it means for a deployment to 'exist' in this context.

utilities/infra.py
brettmthompson: Within the scope of the TimeoutSampler, yes it is handling it. The `timeout_watcher` I create controls the timeout for the entire execution of the `wait_for_inference_deployment_replicas` so that the total timeout gets even distributed across the sub functions called, similar to the `@retry` decorator.

utilities/infra.py
brettmthompson: I am not sure I am understanding your comment. 

The `wait_for_inference_deployment_replicas` method will always create at least 1 label used to filter deployments by calling the `create_isvc_label_selector_str` method, so the label_selector value passed to `Deployment.get()` will never have a value of `None`. 

I am exposing a n optional `labels`  input variable that allows users to pass labels to be used **in addition to** the label returned by the `create_isvc_label_selector_str` function. If they don't need any additional labels to filter the deployments on then they can pass that input variable as None, in which case only the label returned by `create_isvc_label_selector_str`  will be used to filter deployments.

brettmthompson: Maybe the confusion here is that the variable used to capture the result of `create_isvc_label_selector_str` is also called `label_selector`? :smile:

utilities/infra.py
dbasunag: My question is what if `if deployment.exists:` is False? Your previous code (line 237) ensured the number of deployments matched expected number of deployments. But at this line if the deployment does not exist, should that cause a failure?

utilities/infra.py
dbasunag: Nevermind. I misread it before.

utilities/infra.py
dbasunag: If you are using timeout sampler, then this line is not needed. Timeout sampler already handles it.

dbasunag: @rnetser would not this be delta (i.e. only couple of seconds less from whatever the timeout value was passed to the function) from timeout value being passed to the sampler call? I don't understand why it would be needed here. Can you please elaborate? I understand the reason to use it in wait_for_replicas_in_deployment().

utilities/infra.py
```diff
@@ -196,49 +200,75 @@ def wait_for_inference_deployment_replicas(
...
+            exceptions_dict=DEFAULT_CLUSTER_RETRY_EXCEPTIONS,
+            func=Deployment.get,
             label_selector=label_selector,
             client=client,
-            namespace=isvc.namespace,
-        )
-    )
+            namespace=ns,
+        ):
+            deployment_list = list(deployments)
```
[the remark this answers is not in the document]
brettmthompson: The `Deployment.Get()` function will return an object of type `Generator`, which has no `len()` function.

utilities/infra.py
```diff
@@ -196,49 +200,75 @@ def wait_for_inference_deployment_replicas(
...
         timeout (int): Time to wait for the model deployment.
 
     Returns:
         list[Deployment]: List of Deployment objects for InferenceService.
 
+    Raises:
+        TimeoutExpiredError: If an exception is raised when retrieving deployments or
+                             timeout expires when checking replicas.
+        ResourceNotUniqueError: If a greater number of deployments exist than expected after timeout.
+        ResourceNotFoundError: If a less number of deployments exist than expected after timeout.
```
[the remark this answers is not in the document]
brettmthompson: Agreed. I think a similar case can be made for the `ResourceNotUniqueError` used here as well. I will update the logic to raise a new `UnexpectedResourceCountError`

brettmthompson:
@rnetser @mwaykole @dbasunag 

Should be good for review again.

mwaykole (approved):

dbasunag (approved):
@brettmthompson I am good with the PR, can you please see if you can get https://github.com/opendatahub-io/opendatahub-tests/pull/269/files#r2080184023 resolved. Either way, I will not be holding it for this one.

utilities/infra.py
rnetser: please see https://github.com/opendatahub-io/opendatahub-tests/pull/269/files#diff-e2ff6db074d9c48c5dff7b9542d476ac8e50d300dd5614fc2abd7cc5d09e7435R252

PR #274, opened 2025-04-26
inferece - catch when 503 error is retured

rnetser:
## Merge criteria:

### DIFF
--- a/utilities/inference_utils.py
+++ b/utilities/inference_utils.py
@@ -2,6 +2,7 @@ import json
 import re
 import shlex
 from contextlib import contextmanager
+from http import HTTPStatus
 from json import JSONDecodeError
 from string import Template
 from typing import Any, Optional, Generator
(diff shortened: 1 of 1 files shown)


utilities/inference_utils.py
```diff
@@ -433,7 +434,14 @@ def run_inference(self, cmd: str) -> str:
         else:
             res, out, err = run_command(command=shlex.split(cmd), verify_stderr=False, check=False)
 
-        if not res:
+        if res:
+            if f"http/1.0 {HTTPStatus.SERVICE_UNAVAILABLE}" in out.lower():
```
dbasunag, line 438: What if the error shows in stderr instead of stdout?

rnetser: res will be `False`
for this specific error, res is `True` and it appears in `out`

mwaykole (approved):

adolfo-ab (approved):

dbasunag (approved):

PR #283, opened 2025-05-02
test: Add test to verify model registry access for different users created from the run

fege:
## Merge criteria:

## Summary by CodeRabbit

- **New Features**
  - Added comprehensive tests for user and group permissions related to Model Registry access, including scenarios for group membership and direct user role binding.
  - Introduced utilities for managing temporary OpenShift test users and groups, enabling automated creation and cleanup of test user sessions.
  - Provided helper functions for building Model Registry client arguments and verifying user/group access in test environments.
  - Added a context manager to switch OpenShift CLI contexts temporarily during operations.
  - Enhanced test infrastructure with reusable fixtures for dynamic group and user session management.

- **Tests**
  - Implemented end-to-end RBAC tests for Model Registry, covering user, group, and service account access scenarios.
  - Enhanced test infrastructure with reusable fixtures and utility functions to streamline RBAC testing workflows.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -6,6 +6,7 @@ markers =
     # General
     polarion: Store polarion test ID
     jira: Store jira bug ID
+    no_collect: mark a test to not be collected by pytest # utilities/user_utils.py
 
     # CI
     smoke: Mark tests as smoke tests; covers core functionality of the product. Aims to ensure that the build is stable enough for further testing.
--- a/pytest.ini
+++ b/pytest.ini
@@ -6,7 +6,6 @@ markers =
     # General
     polarion: Store polarion test ID
     jira: Store jira bug ID
-    no_collect: mark a test to not be collected by pytest # utilities/user_utils.py
 
     # CI
     smoke: Mark tests as smoke tests; covers core functionality of the product. Aims to ensure that the build is stable enough for further testing.
--- a/tests/model_registry/rbac/conftest.py
+++ b/tests/model_registry/rbac/conftest.py
@@ -93,6 +93,32 @@ def sa_token(service_account: ServiceAccount) -> str:
         raise
 
 
+@pytest.fixture(scope="session", autouse=True)
+def check_env_vars() -> None:
+    """
+    Check if the required environment variables are set.
+    Fails the test session if any are missing.
+    """
+    missing: list[str] = []
+    missing.extend(var for var in ["ADMIN_PASSWORD", "NON_ADMIN_PASSWORD"] if not os.environ.get(var))
+    if missing:
+        pytest.skip(f"Required environment variables not set: {', '.join(missing)}")
+
+
+@pytest.fixture(scope="function")
+def new_group(request: pytest.FixtureRequest) -> Generator[str, None, None]:
+    """
+    Fixture to create a new OpenShift group and add a user, then delete the group after the test.
+    The parameter should be a tuple: (group_name, user_name).
+    """
+    group_name, user_name = request.param
+    run_command(command=["oc", "adm", "groups", "new", group_name, user_name])
+    try:
+        yield group_name
+    finally:
+        run_command(command=["oc", "delete", "group", group_name])
+
+
 # --- RBAC Fixtures ---
 
 
@@ -166,7 +192,7 @@ def mr_access_role_binding(
         subjects_name=f"system:serviceaccounts:{sa_namespace.name}",
         subjects_api_group="rbac.authorization.k8s.io",  # This is the default apiGroup for Group kind
         # Role reference parameters
-        role_ref_kind="Role",
+        role_ref_kind=mr_access_role.kind,
         role_ref_name=mr_access_role.name,
         label=binding_labels,
         wait_for_resource=True,
@@ -174,18 +200,6 @@ def mr_access_role_binding(
         LOGGER.info(f"RoleBinding {binding.name} created successfully.")
         yield binding
         LOGGER.info(f"RoleBinding {binding.name} deletion initiated by context manager.")
-
-
-@pytest.fixture(scope="session", autouse=True)
-def check_env_vars() -> None:
-    """
-    Check if the required environment variables are set.
-    Fails the test session if any are missing.
-    """
-    missing: list[str] = []
-    missing.extend(var for var in ["ADMIN_PASSWORD", "NON_ADMIN_PASSWORD"] if not os.environ.get(var))
-    if missing:
-        pytest.skip(f"Required environment variables not set: {', '.join(missing)}")
 
 
 @pytest.fixture(scope="function")
@@ -214,52 +228,3 @@ def user_in_group_context() -> Callable[[str], ContextManager[None]]:
             run_command(command=["oc", "adm", "groups", "remove-users", "model-registry-users", user_name])
 
     return _context
-
-
-@pytest.fixture
-def new_group(request: pytest.FixtureRequest) -> Generator[str, None, None]:
-    """
-    Fixture to create a new OpenShift group and add a user, then delete the group after the test.
-    The parameter should be a tuple: (group_name, user_name).
-    """
-    group_name, user_name = request.param
-    run_command(command=["oc", "adm", "groups", "new", group_name, user_name])
-    try:
-        yield group_name
-    finally:
-        run_command(command=["oc", "delete", "group", group_name])
-
-
-@pytest.fixture(scope="function")
-def model_registry_role(
-    admin_client: DynamicClient,
-    model_registry_namespace: str,
-) -> Generator[Role, None, None]:
-    """
-    Fixture to create a Role for Model Registry access.
-    """
-    with Role(
-        client=admin_client,
-        namespace=model_registry_namespace,
-        name="test-model-registry-get",
-        rules=[
-            {
-                "apiGroups": [""],
-                "resources": ["services"],
-                "resourceNames": [MR_INSTANCE_NAME],
-                "verbs": ["get"],
-            },
-        ],
-        label={
-            "app.kubernetes.io/component": "model-registry-test-rbac",
-            "test.opendatahub.io/namespace": model_registry_namespace,
-        },
-        wait_for_resource=True,
-    ) as role:
-        try:
-            LOGGER.info(f"Role {role.name} created successfully.")
-            yield role
-            LOGGER.info(f"Role {role.name} deleted successfully.")
-        except Exception as e:  # Catch other potential errors during Role instantiation or wait
-            LOGGER.error(f"Error during Role {role.name} creation or wait: {e}", exc_info=True)
-            pytest.fail(f"Failed during Role {role.name} creation: {e}")
--- a/tests/model_registry/rbac/conftest.py
+++ b/tests/model_registry/rbac/conftest.py
@@ -112,11 +112,17 @@ def new_group(request: pytest.FixtureRequest) -> Generator[str, None, None]:
     The parameter should be a tuple: (group_name, user_name).
     """
     group_name, user_name = request.param
-    run_command(command=["oc", "adm", "groups", "new", group_name, user_name])
+    try:
+        run_command(command=["oc", "adm", "groups", "new", group_name, user_name])
+    except Exception as e:
+        pytest.skip(f"Failed to create group {group_name} and add user {user_name}: {e}")
     try:
         yield group_name
     finally:
-        run_command(command=["oc", "delete", "group", group_name])
+        try:
+            run_command(command=["oc", "delete", "group", group_name])
+        except Exception as e:
+            LOGGER.error(f"Failed to delete group {group_name}: {e}")
 
 
 # --- RBAC Fixtures ---
--- a/tests/model_registry/rbac/conftest.py
+++ b/tests/model_registry/rbac/conftest.py
@@ -94,11 +94,13 @@ def sa_token(service_account: ServiceAccount) -> str:
 
 @pytest.fixture(scope="function")
 def new_group(
-    request: pytest.FixtureRequest, test_idp_user_session: TestUserSession, admin_client: DynamicClient
+    request: pytest.FixtureRequest,
+    admin_client: DynamicClient,
+    test_idp_user_session: TestUserSession,
 ) -> Generator[str, None, None]:
     """
     Fixture to create a new OpenShift group and add a user, then delete the group after the test.
-    The parameter should be a tuple: (group_name, user_name).
+    The group name is passed as a parameter to the fixture and the user is taken from test_idp_user_session.
     """
 
     group_name = request.param
@@ -120,20 +122,20 @@ def test_idp_user_session() -> Generator[TestUserSession, None, None]:
     Returns a TestUserSession object that contains all necessary credentials and contexts.
     """
     # Create the test IDP and user
-    session = create_test_idp()
-    LOGGER.info(f"Created session test IDP user: {session.username}")
+    idp_session = create_test_idp()
+    LOGGER.info(f"Created session test IDP user: {idp_session.username}")
 
     try:
         # Yield the session object to all tests in the session
-        yield session
+        yield idp_session
     finally:
         # Clean up after all tests are done
-        LOGGER.info(f"Cleaning up session test IDP user: {session.username}")
+        LOGGER.info(f"Cleaning up session test IDP user: {idp_session.username}")
         cleanup_test_idp(
-            idp_name=session.idp_name,
-            secret_name=session.secret_name,
-            original_context=session.original_context,
-            user_context=session.user_context,
+            idp_name=idp_session.idp_name,
+            secret_name=idp_session.secret_name,
+            original_context=idp_session.original_context,
+            user_context=idp_session.user_context,
         )
 
 
--- a/tests/model_registry/rbac/conftest.py
+++ b/tests/model_registry/rbac/conftest.py
@@ -1,0 +1,49 @@
+import pytest
+import os
+from typing import Callable, ContextManager, Generator
+from contextlib import contextmanager
+from pyhelper_utils.shell import run_command
+
+
+@pytest.fixture(scope="session", autouse=True)
+def check_env_vars() -> None:
+    """
+    Check if the required environment variables are set.
+    Fails the test session if any are missing.
+    """
+    missing: list[str] = []
+    missing.extend(var for var in ["ADMIN_PASSWORD", "NON_ADMIN_PASSWORD"] if not os.environ.get(var))
+    if missing:
+        pytest.skip(f"Required environment variables not set: {', '.join(missing)}")
+
+
+@pytest.fixture
+def user_in_group_context() -> Callable[[str], ContextManager[None]]:
+    """
+    Fixture to add and remove a user from the model-registry-users group.
+    Returns a context manager that adds the user on entry and removes on exit.
+    """
+
+    @contextmanager
+    def _context(user_name: str) -> Generator[None, None, None]:
+        run_command(command=["oc", "adm", "groups", "add-users", "model-registry-users", user_name])
+        try:
+            yield
+        finally:
+            run_command(command=["oc", "adm", "groups", "remove-users", "model-registry-users", user_name])
+
+    return _context
+
+
+@pytest.fixture
+def new_group(request: pytest.FixtureRequest) -> Generator[str, None, None]:
+    """
+    Fixture to create a new OpenShift group and add a user, then delete the group after the test.
+    The parameter should be a tuple: (group_name, user_name).
+    """
+    group_name, user_name = request.param
+    run_command(command=["oc", "adm", "groups", "new", group_name, user_name])
+    try:
+        yield group_name
+    finally:
+        run_command(command=["oc", "delete", "group", group_name])
--- a/tests/model_registry/rbac/conftest.py
+++ b/tests/model_registry/rbac/conftest.py
@@ -100,6 +100,18 @@ def add_user_to_group(
     admin_client: DynamicClient,
     test_idp_user_session: UserTestSession,
 ) -> Generator[str, None, None]:
+    """
+    Fixture to create a group and add a test user to it.
+    Uses create_group context manager to ensure proper cleanup.
+
+    Args:
+        request: The pytest request object containing the group name parameter
+        admin_client: The admin client for accessing the cluster
+        test_idp_user_session: The test user session containing user information
+
+    Yields:
+        str: The name of the created group
+    """
     group_name = request.param
     with create_group(
         admin_client=admin_client,
@@ -126,9 +138,6 @@ def model_registry_group_with_user(
 
     Yields:
         Group: The group with the test user added
-
-    Note:
-        The group name should be passed as a parameter to the fixture using pytest.mark.parametrize
     """
     group_name = request.param
     group = Group(
--- a/tests/model_registry/rbac/conftest.py
+++ b/tests/model_registry/rbac/conftest.py
@@ -13,7 +13,7 @@ from ocp_resources.group import Group
 from kubernetes.dynamic import DynamicClient
 from pyhelper_utils.shell import run_command
 from tests.model_registry.utils import generate_random_name, generate_namespace_name
-from utilities.user_utils import create_test_idp, cleanup_test_idp, TestUserSession
+from utilities.user_utils import create_test_idp, cleanup_test_idp, UserTestSession
 from tests.model_registry.constants import MR_INSTANCE_NAME
 
 
@@ -96,7 +96,7 @@ def sa_token(service_account: ServiceAccount) -> str:
 def new_group(
     request: pytest.FixtureRequest,
     admin_client: DynamicClient,
-    test_idp_user_session: TestUserSession,
+    test_idp_user_session: UserTestSession,
 ) -> Generator[str, None, None]:
     """
     Fixture to create a new OpenShift group and add a user, then delete the group after the test.
@@ -116,10 +116,10 @@ def new_group(
 
 
 @pytest.fixture(scope="session")
-def test_idp_user_session() -> Generator[TestUserSession, None, None]:
+def test_idp_user_session() -> Generator[UserTestSession, None, None]:
     """
     Session-scoped fixture that creates a test IDP user and cleans it up after all tests.
-    Returns a TestUserSession object that contains all necessary credentials and contexts.
+    Returns a UserTestSession object that contains all necessary credentials and contexts.
     """
     # Create the test IDP and user
     idp_session = create_test_idp()
--- a/tests/model_registry/rbac/conftest.py
+++ b/tests/model_registry/rbac/conftest.py
@@ -4,6 +4,13 @@ from typing import Callable, ContextManager, Generator
 from contextlib import contextmanager
 from pyhelper_utils.shell import run_command
 
+from tests.model_registry.constants import MR_INSTANCE_NAME
+from kubernetes.dynamic import DynamicClient
+from ocp_resources.role import Role
+from simple_logger.logger import get_logger
+
+LOGGER = get_logger(name=__name__)
+
 
 @pytest.fixture(scope="session", autouse=True)
 def check_env_vars() -> None:
@@ -17,7 +24,7 @@ def check_env_vars() -> None:
         pytest.skip(f"Required environment variables not set: {', '.join(missing)}")
 
 
-@pytest.fixture
+@pytest.fixture(scope="function")
 def user_in_group_context() -> Callable[[str], ContextManager[None]]:
     """
     Fixture to add and remove a user from the model-registry-users group.
@@ -47,3 +54,38 @@ def new_group(request: pytest.FixtureRequest) -> Generator[str, None, None]:
         yield group_name
     finally:
         run_command(command=["oc", "delete", "group", group_name])
+
+
+@pytest.fixture(scope="function")
+def model_registry_role(
+    admin_client: DynamicClient,
+    model_registry_namespace: str,
+) -> Generator[Role, None, None]:
+    """
+    Fixture to create a Role for Model Registry access.
+    """
+    with Role(
+        client=admin_client,
+        namespace=model_registry_namespace,
+        name="test-model-registry-get",
+        rules=[
+            {
+                "apiGroups": [""],
+                "resources": ["services"],
+                "resourceNames": [MR_INSTANCE_NAME],
+                "verbs": ["get"],
+            },
+        ],
+        label={
+            "app.kubernetes.io/component": "model-registry-test-rbac",
+            "test.opendatahub.io/namespace": model_registry_namespace,
+        },
+        wait_for_resource=True,
+    ) as role:
+        try:
+            LOGGER.info(f"Role {role.name} created successfully.")
+            yield role
+            LOGGER.info(f"Role {role.name} deleted successfully.")
+        except Exception as e:  # Catch other potential errors during Role instantiation or wait
+            LOGGER.error(f"Error during Role {role.name} creation or wait: {e}", exc_info=True)
+            pytest.fail(f"Failed during Role {role.name} creation: {e}")
--- a/tests/model_registry/rbac/conftest.py
+++ b/tests/model_registry/rbac/conftest.py
@@ -9,7 +9,10 @@ from kubernetes.dynamic import DynamicClient
 from ocp_resources.role import Role
 from simple_logger.logger import get_logger
 
+import time
+
 LOGGER = get_logger(name=__name__)
+SLEEP_TIME = 5
 
 
 @pytest.fixture(scope="session", autouse=True)
@@ -34,6 +37,16 @@ def user_in_group_context() -> Callable[[str], ContextManager[None]]:
     @contextmanager
     def _context(user_name: str) -> Generator[None, None, None]:
         run_command(command=["oc", "adm", "groups", "add-users", "model-registry-users", user_name])
+        max_attempts = 5
+        for attempt in range(max_attempts):
+            result = run_command(command=["oc", "get", "group", "model-registry-users", "-o", "jsonpath='{.users}'"])
+            if user_name in result[1]:
+                LOGGER.info(f"User {user_name} added to model-registry-users group successfully.")
+                break
+            if attempt <= max_attempts - 1:
+                time.sleep(SLEEP_TIME)  # noqa: FCN001
+        else:
+            pytest.fail(f"User {user_name} not added to model-registry-users group after {max_attempts} attempts.")
         try:
             yield
         finally:
--- a/tests/model_registry/rbac/conftest.py
+++ b/tests/model_registry/rbac/conftest.py
@@ -10,10 +10,12 @@ from ocp_resources.service_account import ServiceAccount
 from ocp_resources.role_binding import RoleBinding
 from ocp_resources.role import Role
 from ocp_resources.group import Group
+from ocp_resources.resource import ResourceEditor
 from kubernetes.dynamic import DynamicClient
 from pyhelper_utils.shell import run_command
 from tests.model_registry.utils import generate_random_name, generate_namespace_name
 from utilities.user_utils import create_test_idp, UserTestSession
+from tests.model_registry.rbac.group_utils import create_group
 from tests.model_registry.constants import MR_INSTANCE_NAME
 
 
@@ -93,63 +95,60 @@ def sa_token(service_account: ServiceAccount) -> str:
 
 
 @pytest.fixture(scope="function")
-def new_group(
+def add_user_to_group(
     request: pytest.FixtureRequest,
     admin_client: DynamicClient,
     test_idp_user_session: UserTestSession,
 ) -> Generator[str, None, None]:
-    """
-    Fixture to create a new OpenShift group and add a user, then delete the group after the test.
-    The group name is passed as a parameter to the fixture and the user is taken from test_idp_user_session.
-    """
-
     group_name = request.param
-    with Group(
-        client=admin_client,
-        name=group_name,
+    with create_group(
+        admin_client=admin_client,
+        group_name=group_name,
         users=[test_idp_user_session.username],
-        wait_for_resource=True,
-    ) as _:
-        LOGGER.info(f"Group {group_name} created successfully.")
+    ) as group_name:
         yield group_name
-        LOGGER.info(f"Group {group_name} deletion initiated by context manager.")
 
 
 @pytest.fixture(scope="function")
 def model_registry_group_with_user(
+    request: pytest.FixtureRequest,
     admin_client: DynamicClient,
     test_idp_user_session: UserTestSession,
 ) -> Generator[Group, None, None]:
     """
-    Fixture to manage a test user in the Model Registry group.
+    Fixture to manage a test user in a specified group.
     Adds the user to the group before the test, then removes them after.
 
     Args:
+        request: The pytest request object containing the group name parameter
         admin_client: The admin client for accessing the cluster
         test_idp_user_session: The test user session containing user information
 
     Yields:
-        Group: The Model Registry group with the test user added
+        Group: The group with the test user added
+
+    Note:
+        The group name should be passed as a parameter to the fixture using pytest.mark.parametrize
     """
-    model_registry_users_group = f"{MR_INSTANCE_NAME}-users"
+    group_name = request.param
     group = Group(
         client=admin_client,
-        name=model_registry_users_group,
+        name=group_name,
         wait_for_resource=True,
     )
 
     # Add user to group
-    group.update(
-        resource_dict={"metadata": {"name": model_registry_users_group}, "users": [test_idp_user_session.username]}
-    )
-    LOGGER.info(f"Added user {test_idp_user_session.username} to {model_registry_users_group} group")
-
-    try:
+    with ResourceEditor(
+        patches={
+            group: {
+                "metadata": {"name": group_name},
+                "users": [test_idp_user_session.username],
+            }
+        }
+    ) as _:
+        LOGGER.info(f"Added user {test_idp_user_session.username} to {group_name} group")
         yield group
-    finally:
-        # Remove user from group
-        group.update(resource_dict={"metadata": {"name": model_registry_users_group}, "users": []})
-        LOGGER.info(f"Removed user {test_idp_user_session.username} from {model_registry_users_group} group")
+        LOGGER.info(f"Removed user {test_idp_user_session.username} from {group_name} group")
 
 
 @pytest.fixture(scope="session")
--- a/tests/model_registry/rbac/conftest.py
+++ b/tests/model_registry/rbac/conftest.py
@@ -1,18 +1,179 @@
 import pytest
+import shlex
+import subprocess
 import os
-from typing import Callable, ContextManager, Generator
+import time
 from contextlib import contextmanager
-from pyhelper_utils.shell import run_command
 
-from tests.model_registry.constants import MR_INSTANCE_NAME
-from kubernetes.dynamic import DynamicClient
+from typing import Callable, ContextManager, Generator, List, Dict, Any
+from ocp_resources.namespace import Namespace
+from ocp_resources.service_account import ServiceAccount
+from ocp_resources.role_binding import RoleBinding
 from ocp_resources.role import Role
+from kubernetes.dynamic import DynamicClient
+from pyhelper_utils.shell import run_command
+from tests.model_registry.utils import generate_random_name, generate_namespace_name
+from tests.model_registry.constants import MR_INSTANCE_NAME
 from simple_logger.logger import get_logger
 
-import time
 
 LOGGER = get_logger(name=__name__)
+DEFAULT_TOKEN_DURATION = "10m"
 SLEEP_TIME = 5
+
+
+@pytest.fixture(scope="function")
+def sa_namespace(request: pytest.FixtureRequest, admin_client: DynamicClient) -> Generator[Namespace, None, None]:
+    """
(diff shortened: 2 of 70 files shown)


fege:
@coderabbitai review

fege:
@coderabbitat review

tests/model_registry/utils.py
```diff
@@ -163,6 +163,32 @@ def get_model_registry_db_label_dict(db_resource_name: str) -> dict[str, str]:
...
+    """
+
+    services = list(
+        Service.get(
+            dyn_client=admin_client, namespace=namespace, label_selector="app=model-registry,component=model-registry"
+        )
+    )
+    if not services:
+        raise RuntimeError(f"Model Registry service not found in namespace {namespace}")
+    return services[0]
```
lugi0, line 189: Can't you use https://github.com/opendatahub-io/opendatahub-tests/pull/283/files#diff-1d1c4a8accca671d431801d22056cfc7deee5b4bb6ff6f1f6c684a8fc17cae62L21 ?

fege: probably yes, let me try out

tests/model_registry/test_user_permission.py
```diff
@@ -0,0 +1,96 @@
...
+
+LOGGER = get_logger(name=__name__)
+TEST_NAMESPACE = "model-registry-test-ns"
+
+
+def get_token(user_name: str, password: str, admin_client: DynamicClient) -> str:
+    current_context = run_command(command=["oc", "config", "current-context"])[1].strip()
+
+    command: str = (
+        f"oc login  --insecure-skip-tls-verify=true {admin_client.configuration.host} -u {user_name} -p {password}"
```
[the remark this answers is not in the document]
fege: this runs only on test environments

tests/model_registry/test_user_permission.py
```diff
@@ -0,0 +1,137 @@
...
+        LOGGER.info("User has no acces to MR")
+        assert_mr_client(user_token=user_token, admin_client=admin_client, context=pytest.raises(ForbiddenException))
+
+        LOGGER.info("Add user hthe model registry users group")
+        run_command(command=["oc", "adm", "groups", "add-users", "model-registry-users", user_name])
+
+        user_token = get_token(user_name=user_name, password=password, admin_client=admin_client)
+
+        LOGGER.info("User has acces to MR")
+        assert_mr_client(user_token=user_token, admin_client=admin_client, context=nullcontext())
```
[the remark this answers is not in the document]
fege: added cleanup

tests/model_registry/utils.py
fege: removed function

tests/model_registry/test_user_permission.py
```diff
@@ -0,0 +1,292 @@
...
+
+
+@pytest.mark.parametrize(
+    "updated_dsc_component_state_scope_class",
+    [
+        pytest.param({
+            "component_patch": {
+                DscComponents.MODELREGISTRY: {
+                    "managementState": DscComponents.ManagementState.MANAGED,
+                    "registriesNamespace": TEST_NAMESPACE,
```
lugi0, line 119: you can use MR_NAMESPACE here

tests/model_registry/test_user_permission.py
```diff
@@ -0,0 +1,292 @@
...
+            user_token=user_token,
+            admin_client=admin_client,
+            context=pytest.raises(ForbiddenException),
+            mr_instance=model_registry_instance,
+            mr_namespace_name=model_registry_namespace,
+        )
+
+        LOGGER.info("Add user to the model registry users group")
+        with user_in_group_context(user_name):
+            user_token = get_token(user_name=user_name, password=password, admin_client=admin_client)
```
lugi0, line 200: IIRC after adding a user to a group it takes a while to get "reconciled" such that the token works as expected. Is there some way to wait until the reconciliation is successful? and/or adding a retry to the assert_mr_client call

tests/model_registry/test_user_permission.py
```diff
@@ -0,0 +1,292 @@
...
+        """
+
+        LOGGER.info("Group created and user added to it")
+
+        with RoleBinding(
+            client=admin_client,
+            namespace=model_registry_namespace,
+            name="test-model-registry-group-edit",
+            role_ref_name="edit",
+            role_ref_kind="ClusterRole",
```
lugi0, line 240: Where do you create this role?

lugi0 (changes requested):
Can you refactor this such that the test file is under a subfolder (called something like `RBAC` or similar?) and that all the fixtures and helper functions are defined in the appropriate files (conftest.py, ...)?

tests/model_registry/test_user_permission.py
fege: is it not created by default? `oc get clusterroles -o yaml | grep -A 10 modelregistry`

lugi0: The `edit` role present by default on the cluster is not the Role we want to use, since it has a ton of permissions for all kinds of different resources. The Role we want here should only give permission to get the MR resources we need, i.e.
```
rules:
 - verbs:
 	- get
   apiGroups:
 	- ''
   resources:
 	- services
   resourceNames:
 	- <MR_NAME>
 - verbs:
 	- get
   apiGroups:
 	- ''
   resources:
 	- endpoints
   resourceNames:
 	- <MR_NAME>
```

tests/model_registry/rbac/conftest.py
```diff
@@ -2,20 +2,24 @@
...
 from pyhelper_utils.shell import run_command
 from tests.model_registry.utils import generate_random_name, generate_namespace_name
-from simple_logger.logger import get_logger
 from tests.model_registry.constants import MR_INSTANCE_NAME
+from simple_logger.logger import get_logger
 
 
 LOGGER = get_logger(name=__name__)
 DEFAULT_TOKEN_DURATION = "10m"
+SLEEP_TIME = 5
```
dbasunag, line 22: if you use a constant, then please add unit in the name. Otherwise, it is ok to use 5 directly where you need to.

tests/model_registry/rbac/conftest.py
```diff
@@ -89,6 +93,38 @@ def sa_token(service_account: ServiceAccount) -> str:
...
+@pytest.fixture(scope="session", autouse=True)
+def check_env_vars() -> None:
+    """
+    Check if the required environment variables are set.
+    Fails the test session if any are missing.
+    """
+    missing: list[str] = []
+    missing.extend(var for var in ["ADMIN_PASSWORD", "NON_ADMIN_PASSWORD"] if not os.environ.get(var))
+    if missing:
+        pytest.skip(f"Required environment variables not set: {', '.join(missing)}")
```
dbasunag, line 105: Why do we need to set it? It should be randomly generated in the test session.

tests/model_registry/rbac/conftest.py
```diff
@@ -89,6 +93,38 @@ def sa_token(service_account: ServiceAccount) -> str:
...
+def new_group(request: pytest.FixtureRequest) -> Generator[str, None, None]:
+    """
+    Fixture to create a new OpenShift group and add a user, then delete the group after the test.
+    The parameter should be a tuple: (group_name, user_name).
+    """
+    group_name, user_name = request.param
+    try:
+        run_command(command=["oc", "adm", "groups", "new", group_name, user_name])
+    except Exception as e:
+        pytest.skip(f"Failed to create group {group_name} and add user {user_name}: {e}")
```
dbasunag, line 118: We should raise it. This can potentially cause silent skippping of tests and we don't want that.

tests/model_registry/rbac/conftest.py
```diff
@@ -89,6 +93,38 @@ def sa_token(service_account: ServiceAccount) -> str:
...
+    try:
```
dbasunag, line 119: Why do you need a try here?

tests/model_registry/rbac/conftest.py
```diff
@@ -89,6 +93,38 @@ def sa_token(service_account: ServiceAccount) -> str:
...
+        yield group_name
+    finally:
```
dbasunag, line 121: Do you need a finally? if the command adds the user to the group, only then we need to delete it. in case of exception to add the user, there is nothing to delete, right?

tests/model_registry/rbac/conftest.py
```diff
@@ -89,6 +93,38 @@ def sa_token(service_account: ServiceAccount) -> str:
...
+        try:
+            run_command(command=["oc", "delete", "group", group_name])
+        except Exception as e:
+            LOGGER.error(f"Failed to delete group {group_name}: {e}")
```
dbasunag, line 125: We also need to raise the exception here.

tests/model_registry/rbac/conftest.py
```diff
@@ -162,11 +198,39 @@ def mr_access_role_binding(
...
+        for attempt in range(max_attempts):
+            result = run_command(command=["oc", "get", "group", "model-registry-users", "-o", "jsonpath='{.users}'"])
+            if user_name in result[1]:
+                LOGGER.info(f"User {user_name} added to model-registry-users group successfully.")
+                break
+            if attempt <= max_attempts - 1:
+                time.sleep(SLEEP_TIME)  # noqa: FCN001
+        else:
+            pytest.fail(f"User {user_name} not added to model-registry-users group after {max_attempts} attempts.")
+        try:
```
dbasunag, line 231: I don't understand this try. Can you please elaborate why we need it?

tests/model_registry/rbac/conftest.py
```diff
@@ -162,11 +198,39 @@ def mr_access_role_binding(
...
+            yield
+        finally:
```
dbasunag, line 233: You don't need finally. If the command failed to add, then we need to remove it.

tests/model_registry/rbac/conftest.py
```diff
@@ -162,11 +198,39 @@ def mr_access_role_binding(
...
+    def _context(user_name: str) -> Generator[None, None, None]:
+        run_command(command=["oc", "adm", "groups", "add-users", "model-registry-users", user_name])
+        max_attempts = 5
+        for attempt in range(max_attempts):
+            result = run_command(command=["oc", "get", "group", "model-registry-users", "-o", "jsonpath='{.users}'"])
+            if user_name in result[1]:
+                LOGGER.info(f"User {user_name} added to model-registry-users group successfully.")
+                break
+            if attempt <= max_attempts - 1:
+                time.sleep(SLEEP_TIME)  # noqa: FCN001
```
dbasunag, line 228: Why we need to retry this? Also, if the oc adm add-user command is successful, then we don't need to validate it via get call. We are not testing oc adm command :)

tests/model_registry/rbac/test_mr_rbac.py
```diff
@@ -0,0 +1,248 @@
...
+from ocp_resources.role_binding import RoleBinding
+from utilities.constants import DscComponents, Protocols
+from mr_openapi.exceptions import ForbiddenException
+
+
+LOGGER = get_logger(name=__name__)
+NEW_GROUP_NAME = "test-model-registry-group"
+
+
+def assert_positive_mr_registry(
```
dbasunag, line 23: Let's move it to utils.py

tests/model_registry/rbac/test_mr_rbac.py
```diff
@@ -0,0 +1,248 @@
...
+    model_registry_namespace: str,
+    admin_client: DynamicClient,
+    user_name: str,
+    password: str,
+) -> None:
+    """
+    Assert that a user has access to the Model Registry.
+    """
+    user_token = get_token(user_name=user_name, password=password, admin_client=admin_client)
+    namespace_instance = admin_client.resources.get(api_version="v1", kind="Namespace").get(
```
dbasunag, line 34: We should use wrapper object. 
```
Namespace(name= model_registry_namespace)
```

tests/model_registry/rbac/test_mr_rbac.py
```diff
@@ -0,0 +1,248 @@
...
+        context_manager: ContextManager,
+    ):
+        """
+        Test that a user with permission can access the Model Registry,
+        and a user without permission receives a ForbiddenException.
+        """
+        LOGGER.info("-----Test that a user with permission can access the Model Registry-----")
+        assert model_registry_instance.name == MR_INSTANCE_NAME
+        user_token = get_token(user_name=user_name, password=password, admin_client=admin_client)
+        namespace_instance = admin_client.resources.get(api_version="v1", kind="Namespace").get(
```
dbasunag, line 90: Ditto. Please use `Namespace` class from wrapper.

tests/model_registry/rbac/test_mr_rbac.py
```diff
@@ -0,0 +1,248 @@
...
+)
+@pytest.mark.usefixtures("updated_dsc_component_state_scope_class")
+class TestUserPermission:
+    """
+    Test suite for verifying user and group permissions for the Model Registry.
+    """
+
+    @pytest.mark.smoke
+    @pytest.mark.parametrize(
+        "user_name, password, context_manager",
```
dbasunag, line 68: Let's use a different name for `context_manager`

tests/model_registry/rbac/test_mr_rbac.py
```diff
@@ -0,0 +1,248 @@
...
+    indirect=True,
+    scope="class",
+)
+@pytest.mark.usefixtures("updated_dsc_component_state_scope_class")
+class TestUserPermission:
+    """
+    Test suite for verifying user and group permissions for the Model Registry.
+    """
+
+    @pytest.mark.smoke
```
dbasunag, line 66: This does not seem like a good candidate for smoke test.

tests/model_registry/rbac/test_mr_rbac.py
```diff
@@ -0,0 +1,248 @@
...
+        if exc:
+            http_error = exc.value
+            assert http_error.body is not None, "HTTPError should have a response object"
+            LOGGER.info(f"Received expected HTTP error: Status Code {http_error.status}")
+            assert http_error.status == 403, f"Expected HTTP 403 Forbidden, but got {http_error.status}"
+            LOGGER.info("Successfully received expected HTTP 403 status code.")
+        else:  # If no exception was raised
+            LOGGER.info("Successfully created Model Registry client")
+
+    @pytest.mark.smoke
```
dbasunag, line 108: ditto

tests/model_registry/rbac/test_mr_rbac.py
```diff
@@ -0,0 +1,248 @@
...
+    """
+
+    @pytest.mark.smoke
+    @pytest.mark.parametrize(
+        "user_name, password, context_manager",
+        [
+            ("htpasswd-cluster-admin-user", os.environ.get("ADMIN_PASSWORD"), nullcontext()),
+            ("ldap-user1", os.environ.get("NON_ADMIN_PASSWORD"), pytest.raises(ForbiddenException)),
+        ],
+    )
```
dbasunag, line 73: We don't need to parameterize the tests. We only have 1 test here.

tests/model_registry/rbac/test_mr_rbac.py
```diff
@@ -0,0 +1,248 @@
...
+        with user_in_group_context(user_name):
+            assert_positive_mr_registry(
+                model_registry_instance=model_registry_instance,
+                model_registry_namespace=model_registry_namespace,
+                admin_client=admin_client,
+                user_name=user_name,
+                password=password,
+            )
+
+    @pytest.mark.smoke
```
dbasunag, line 163: ditto. Not a smoke test candidate.

tests/model_registry/rbac/test_mr_rbac.py
fege: file changed now but the concept is the same, same test with admin and non admin user to verify that the first as the access and the other one no

tests/model_registry/rbac/test_mr_rbac.py
fege: done

tests/model_registry/rbac/test_mr_rbac.py
fege: done

tests/model_registry/rbac/conftest.py
fege: code removed

tests/model_registry/rbac/conftest.py
fege: code written before the idp and user utils

tests/model_registry/rbac/conftest.py
fege: code changed

fege:
@coderabbitai full review

tests/model_registry/rbac/conftest.py
fege: code is changed

tests/model_registry/rbac/test_mr_rbac.py
```diff
@@ -0,0 +1,365 @@
...
+                DscComponents.MODELREGISTRY: {
+                    "managementState": DscComponents.ManagementState.MANAGED,
+                    "registriesNamespace": MR_NAMESPACE,
+                },
+            }
+        })
+    ],
+    indirect=True,
+    scope="class",
+)
```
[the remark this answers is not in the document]
fege: there is no usage of ADMIN_PASSWORD or NON_ADMIN_PASSWORD, so I do not get your point. Please explain

tests/model_registry/rbac/test_mr_rbac_sa.py
```diff
@@ -1,28 +1,16 @@
 # AI Disclaimer: Google Gemini 2.5 pro has been used to generate a majority of this code, with human review and editing.
 import pytest
-from typing import Self, Dict, Any
+from typing import Self
```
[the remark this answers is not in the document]
fege: [project]
requires-python = "==3.12.*" 
is defined in pyproject.toml, so I do not think that this is a problem

utilities/user_utils.py
```diff
@@ -0,0 +1,265 @@
...
+    """
+    user_context = ""
+    try:
+        # Save the current context (cluster admin)
+        _, original_context, _ = run_command(command=["oc", "config", "current-context"], check=True)
+        original_context = original_context.strip()
+        LOGGER.info(f"Original context (cluster admin): {original_context}")
+
+        # Generate a unique suffix for this test run
+        unique_suffix = "".join(random.choices(string.ascii_lowercase + string.digits, k=8))
```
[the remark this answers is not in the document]
fege: original_context always exists otherwise test are not even starting

pytest.ini
```diff
@@ -6,6 +6,7 @@ markers =
     # General
     polarion: Store polarion test ID
     jira: Store jira bug ID
+    no_collect: mark a test to not be collected by pytest # utilities/user_utils.py
```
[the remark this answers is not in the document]
fege: no_collect is used in TestUserSession that is a dataclass

utilities/user_utils.py
```diff
@@ -0,0 +1,265 @@
...
+                "generic",
+                secret_name,
+                f"--from-file=htpasswd={temp_file.name}",
+                "-n",
+                "openshift-config",
+            ],
+            check=True,
+        )
+        LOGGER.info(f"Created secret: {secret_name}")
+
```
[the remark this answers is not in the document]
fege: htpasswd is always present on the cluster

tests/model_registry/rbac/conftest.py
```diff
@@ -89,6 +92,51 @@ def sa_token(service_account: ServiceAccount) -> str:
...
+
+
+@pytest.fixture(scope="session")
+def test_idp_user_session() -> Generator[TestUserSession, None, None]:
+    """
+    Session-scoped fixture that creates a test IDP user and cleans it up after all tests.
+    Returns a TestUserSession object that contains all necessary credentials and contexts.
+    """
+    # Create the test IDP and user
+    session = create_test_idp()
```
dbasunag, line 123: I would use a different name for this variable. May be idp_session?

tests/model_registry/rbac/conftest.py
```diff
@@ -89,6 +92,51 @@ def sa_token(service_account: ServiceAccount) -> str:
         raise
 
 
+@pytest.fixture(scope="function")
+def new_group(
+    request: pytest.FixtureRequest, test_idp_user_session: TestUserSession, admin_client: DynamicClient
```
dbasunag, line 97: Whenever we use admin_client, we should use it before other session scoped fixtures. The reason being, it is used always before other session scoped fixtures.

tests/model_registry/rbac/test_mr_rbac.py
```diff
@@ -0,0 +1,299 @@
...
+            model_registry_instance: The Model Registry instance to test against
+            model_registry_namespace: The namespace where Model Registry is deployed
+            admin_client: The admin client for accessing the cluster
+            test_idp_user_session: The test user session containing both admin and user contexts
+
+        Raises:
+            AssertionError: If access permissions don't match expectations
+            ForbiddenException: Expected before group addition, unexpected after
+        """
+        LOGGER.info(
```
dbasunag, line 131: We already print start of tests, that would tell you which test is running. Do you still need it?

tests/model_registry/rbac/test_mr_rbac.py
```diff
@@ -0,0 +1,299 @@
...
+            resource_dict={"metadata": {"name": model_registry_users_group}, "users": [test_idp_user_session.username]}
+        )
+
+        # Verify group membership
+        group = Group(client=admin_client, name=model_registry_users_group)
+        users = group.instance.get("users", []) or []
+        assert test_idp_user_session.username in users, (
+            f"User {test_idp_user_session.username} not in group {model_registry_users_group}. Current users: {users}"
+        )
+        LOGGER.info(f"Added user {test_idp_user_session.username} to {model_registry_users_group} group")
```
dbasunag, line 165: I would create a utility for this. Also, add a check please to ensure group exists. Either via ensure_exists=True or using group.exists.

tests/model_registry/rbac/test_mr_rbac.py
```diff
@@ -0,0 +1,299 @@
...
+                    wait_timeout=240,
+                    sleep=5,
+                    func=assert_positive_mr_registry,
+                    model_registry_instance=model_registry_instance,
+                    model_registry_namespace=model_registry_namespace,
+                    admin_client=admin_client,
+                )
+                for _ in sampler:
+                    break  # Break after first successful iteration
+                LOGGER.info("Successfully accessed Model Registry")
```
dbasunag, line 185: Ditto. Please create an utility. Even though it is only used once, it would keep the test code small and concise :).

tests/model_registry/rbac/test_mr_rbac.py
```diff
@@ -0,0 +1,299 @@
...
+        finally:
+            group = Group(
+                client=admin_client,
+                name=model_registry_users_group,
+            )
+            group.update(resource_dict={"metadata": {"name": model_registry_users_group}, "users": []})
+            LOGGER.info(f"Removed user {test_idp_user_session.username} from {model_registry_users_group} group")
```
dbasunag, line 192: Why not create a fixture instead? That way, we don't have to handle clean up in tests.

tests/model_registry/rbac/test_mr_rbac.py
```diff
@@ -0,0 +1,299 @@
...
+
```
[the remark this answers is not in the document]
dbasunag: @fege do we need this? What do you suggest?

tests/model_registry/rbac/utils.py
```diff
@@ -0,0 +1,55 @@
...
+    Note: Uses is_secure=False for testing purposes.
+    """
+    server, port = rest_endpoint.split(":")
+    return {
+        "server_address": f"{Protocols.HTTPS}://{server}",
+        "port": port,
+        "user_token": token,
+        "is_secure": False,
+        "author": author,
+    }
```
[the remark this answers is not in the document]
dbasunag: It would be an overkill. We are fine without this check.

tests/model_registry/rbac/utils.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+@contextmanager
+def use_context(context_name: str) -> Generator[None, None, None]:
```
dbasunag, line 41: I would name it `switch_user_context` or something similar please.

tests/model_registry/rbac/test_mr_rbac.py
```diff
@@ -0,0 +1,299 @@
...
+        group = Group(
+            client=admin_client,
+            name=model_registry_users_group,
+        )
+        group.update(
+            resource_dict={"metadata": {"name": model_registry_users_group}, "users": [test_idp_user_session.username]}
+        )
+
+        # Verify group membership
+        group = Group(client=admin_client, name=model_registry_users_group)
```
lugi0, line 160: I don't think you need to do this again

tests/model_registry/rbac/test_mr_rbac.py
```diff
@@ -0,0 +1,299 @@
...
+                    func=assert_positive_mr_registry,
+                    model_registry_instance=model_registry_instance,
+                    model_registry_namespace=model_registry_namespace,
+                    admin_client=admin_client,
+                )
+                for _ in sampler:
+                    break  # Break after first successful iteration
+                LOGGER.info("Successfully accessed Model Registry")
+        finally:
+            group = Group(
```
lugi0, line 187: same, you already have the group reference in scope here

tests/model_registry/rbac/test_mr_rbac.py
fege: I think it is better to have this, it is helping in recognize faster the test and also it is targeting the test and not setup or teardown
```-----Test ..... -----```

File: tests/model_registry/rbac/conftest.py [hunk 40]
```diff
@@ -89,6 +92,51 @@ def sa_token(service_account: ServiceAccount) -> str:
...
+) -> Generator[str, None, None]:
+    """
+    Fixture to create a new OpenShift group and add a user, then delete the group after the test.
+    The parameter should be a tuple: (group_name, user_name).
```
lugi0, line 101: It doesn't look like this is true? you get the username from the `test_idp_user_session`

lugi0 (commented):
largely LGTM, a couple comments

tests/model_registry/rbac/test_mr_rbac.py
fege: I understand the previous one, but here it is used only once as you said but also it is only calling the other methods `setup_mr_client` and `TimeoutSampler`. IMHO is not worth to create a 3rd utils to just call 2 other utils and nothing more

tests/model_registry/rbac/test_mr_rbac.py
fege: if I use a fixture the user would be added before the test starts, so I could not check in the test run that the user does not have access to MR but they will gain once they are added to the correct group

tests/model_registry/rbac/utils.py
fege: switch_context

tests/model_registry/rbac/test_mr_rbac.py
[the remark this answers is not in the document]
fege: I don't think we need this

lugi0 (approved):

tests/model_registry/rbac/utils.py
```diff
@@ -0,0 +1,130 @@
...
+        as it uses the current context to get the token.
+    """
+    token = get_openshift_token()
+    namespace_instance = Namespace(client=admin_client, name=model_registry_namespace)
+    svc = get_mr_service_by_label(client=admin_client, ns=namespace_instance, mr_instance=model_registry_instance)
+    endpoint = get_endpoint_from_mr_service(svc=svc, protocol=Protocols.REST)
+    client_args = build_mr_client_args(rest_endpoint=endpoint, token=token, author="rbac-test-user-granted")
+    mr_client = ModelRegistryClient(**client_args)
+    assert mr_client is not None, "Client initialization failed after granting permissions"
+    LOGGER.info("Client instantiated successfully after granting permissions.")
```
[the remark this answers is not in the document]
dbasunag: Not needed. if we fail to assert that is a good indication.

dbasunag (changes requested):

utilities/user_utils.py
```diff
@@ -0,0 +1,265 @@
...
+from pyhelper_utils.shell import run_command
+from timeout_sampler import TimeoutSampler
+from utilities.infra import login_with_user_password
+
+
+LOGGER = logging.getLogger(__name__)
+SLEEP_TIME = 5
+
+
+@pytest.mark.no_collect
```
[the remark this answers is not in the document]
fege: removed

tests/model_registry/rbac/test_mr_rbac.py
dbasunag: We print ```--------------<test name>------------``` at the begining, then `-----SETUP-------` for fixtures, `---------CALL---------------` for test body and then `-------------TEARDOWN----------`. Is not that good enough?

tests/model_registry/rbac/test_mr_rbac.py
```diff
@@ -0,0 +1,305 @@
...
+        with switch_context(test_idp_user_session.user_context):
+            _, client_args = setup_mr_client(
+                model_registry_instance=model_registry_instance,
+                model_registry_namespace=model_registry_namespace,
+                admin_client=admin_client,
+            )
+
+            with pytest.raises(ForbiddenException) as exc_info:
+                _ = ModelRegistryClient(**client_args)
+            assert exc_info.value.status == 403
```
dbasunag, line 158: This is already covered in test_user_permission(). You can remove this section.

tests/model_registry/rbac/test_mr_rbac.py
```diff
@@ -0,0 +1,305 @@
...
+        try:
+            # Wait for access to be granted
+            with switch_context(test_idp_user_session.user_context):
+                _, client_args = setup_mr_client(
+                    model_registry_instance=model_registry_instance,
+                    model_registry_namespace=model_registry_namespace,
+                    admin_client=admin_client,
+                )
+
+                sampler = TimeoutSampler(
```
dbasunag, line 185: Should be a fixture.

tests/model_registry/rbac/utils.py
```diff
@@ -0,0 +1,149 @@
...
+        "server_address": f"{Protocols.HTTPS}://{server}",
+        "port": port,
+        "user_token": token,
+        "is_secure": False,
+        "author": author,
+    }
+
+
+@contextmanager
+def switch_context(context_name: str) -> Generator[None, None, None]:
```
dbasunag, line 42: Can you please rename to `switch_user_context` or `switch_openshift_user_context`?

tests/model_registry/rbac/utils.py
```diff
@@ -0,0 +1,149 @@
...
+    namespace_instance = Namespace(client=admin_client, name=model_registry_namespace)
+    svc = get_mr_service_by_label(client=admin_client, ns=namespace_instance, mr_instance=model_registry_instance)
+    endpoint = get_endpoint_from_mr_service(svc=svc, protocol=Protocols.REST)
+    client_args = build_mr_client_args(rest_endpoint=endpoint, token=token, author="rbac-test-user-granted")
+    mr_client = ModelRegistryClient(**client_args)
+    assert mr_client is not None, "Client initialization failed after granting permissions"
+    LOGGER.info("Client instantiated successfully after granting permissions.")
+
+
+def setup_mr_client(
```
dbasunag, line 103: This function is not setting mr client though. Instead it is setting arguments.

utilities/user_utils.py
```diff
@@ -0,0 +1,263 @@
...
+    """
+    user_context = ""
+    try:
+        # Save the current context (cluster admin)
+        _, original_context, _ = run_command(command=["oc", "config", "current-context"], check=True)
+        original_context = original_context.strip()
+        LOGGER.info(f"Original context (cluster admin): {original_context}")
+
+        # Generate a unique suffix for this test run
+        unique_suffix = "".join(random.choices(string.ascii_lowercase + string.digits, k=8))
```
dbasunag, line 51: Luca already created an util for this. generate_random_name(). Can that be used?

utilities/user_utils.py
```diff
@@ -0,0 +1,263 @@
...
+        idp_name = f"{idp_name}-{unique_suffix}"
+        secret_name = f"{secret_name}-{unique_suffix}"
+
+        # Generate username and password
+        username = f"test-user-{secrets.token_hex(4)}"
+        password = "".join(secrets.choice(string.ascii_letters + string.digits) for _ in range(12))
```
dbasunag, line 57: Ditto. Let's use generate_random_name()

utilities/user_utils.py
```diff
@@ -0,0 +1,263 @@
...
+    __test__ = False
+    idp_name: str
+    secret_name: str
+    username: str
+    password: str
+    user_context: str
+    original_context: str
+
+
+def create_test_idp(idp_name: str = "test-htpasswd-idp", secret_name: str = "test-htpasswd-secret") -> UserTestSession:
```
dbasunag, line 32: Can this use contextmanager decorator?

utilities/user_utils.py
```diff
@@ -0,0 +1,263 @@
...
+        secret_name: The name of the secret to create.
+    """
+    with tempfile.NamedTemporaryFile(mode="w+") as temp_file:
+        subprocess.run(args=["htpasswd", "-c", "-b", temp_file.name, username, password], check=True)
+        LOGGER.info(f"Creating secret {secret_name} in openshift-config namespace")
+        run_command(
+            command=[
+                "oc",
+                "create",
+                "secret",
```
dbasunag, line 193: We should be using wrapper secret object. We don't need to use run_command() for this call.

utilities/user_utils.py
```diff
@@ -0,0 +1,263 @@
...
+
+
+def _remove_idp_from_oauth(idp_name: str) -> None:
+    """
+    Removes a specified identity provider from the OpenShift OAuth configuration.
+
+    Args:
+        idp_name: The name of the identity provider to remove.
+    """
+    _, stdout, _ = run_command(command=["oc", "get", "oauth", "cluster", "-o", "json"], check=True)
```
dbasunag, line 235: Ditto. please use https://github.com/RedHatQE/openshift-python-wrapper/blob/main/ocp_resources/oauth.py

tests/model_registry/rbac/test_mr_rbac.py
fege: Screenshot 2025-05-20 at 13 57 23
in my opinion that works only if the setup is really short and close to `----CALL-----` otherwise you do not see on the terminal which test the `CALL` is referring to. Check this screenshot. But if this is blocking the PR I could remove

utilities/user_utils.py
fege: I tried and it is not really working fine with the update

utilities/user_utils.py
fege: same here as for the oauth

dbasunag: Creation should have no issues. We already use to to create model registry secret.

tests/model_registry/rbac/test_mr_rbac.py
fege: removed

tests/model_registry/rbac/test_mr_rbac.py
```diff
@@ -0,0 +1,305 @@
...
+            admin_client: The admin client for accessing the cluster
+            test_idp_user_session: The test user session containing both admin and user contexts
+            use_admin_context: Whether to use admin context (True) or user context (False)
+
+        Raises:
+            AssertionError: If access permissions don't match expectations
+            ForbiddenException: Expected for normal users, unexpected for admin users
+        """
+
+        assert model_registry_instance.name == MR_INSTANCE_NAME
```
dbasunag, line 89: We have similar validations in other tests. I don't believe you need it here.

tests/model_registry/rbac/test_mr_rbac.py
dbasunag: Not going to block the PR, but I don't see why `CALL` message is not enough to understand start of the test. Then may be we should update `CALL` to something meaningful?

tests/model_registry/rbac/test_mr_rbac.py
```diff
@@ -0,0 +1,305 @@
...
+
+        Raises:
+            AssertionError: If access permissions don't match expectations
+            ForbiddenException: Expected before group addition, unexpected after
+        """
+        LOGGER.info(
+            "-----Test that a user can access to the Model Registry once added to a "
+            "group that has the permissions to access it-----"
+        )
+        assert model_registry_instance.name == MR_INSTANCE_NAME
```
dbasunag, line 145: Not needed.

tests/model_registry/rbac/test_mr_rbac.py
dbasunag: I still see it here.

tests/model_registry/rbac/test_mr_rbac.py
dbasunag: Open comment.

tests/model_registry/rbac/utils.py
dbasunag: It is not yet addressed.

tests/model_registry/rbac/utils.py
dbasunag: Not yet addressed.

dbasunag (commented):
@fege Not sure if this PR is ready for review. I see a bunch of comments resolved, but they are not addressed. Going to hold off review till you indicate it is ready for review.

tests/model_registry/rbac/test_mr_rbac.py
fege: maybe we should add the test name ora something more in `CALL` for how is it now I found difficult to know which test is been called, especially if the `SETUP` is verbose. But I would like to not do the change in this PR

fege:
> @fege Not sure if this PR is ready for review. I see a bunch of comments resolved, but they are not addressed. Going to hold off review till you indicate it is ready for review.

they were resolved but not pushed, I resolved to keep track for myself. Now they should be addressed

tests/model_registry/rbac/conftest.py
```diff
@@ -89,6 +92,77 @@ def sa_token(service_account: ServiceAccount) -> str:
...
+    group_name = request.param
+    with Group(
+        client=admin_client,
+        name=group_name,
+        users=[test_idp_user_session.username],
+        wait_for_resource=True,
+    ) as _:
+        LOGGER.info(f"Group {group_name} created successfully.")
+        yield group_name
+        LOGGER.info(f"Group {group_name} deletion initiated by context manager.")
```
lugi0, line 115: this would be a good candidate for factory/composition, I assume we will need a fixture to create an empty group at some point, and that could be the factory/builder, while your fixture could pass in the additional user. Maybe worth keeping in mind for some refactoring.

tests/model_registry/rbac/conftest.py
```diff
@@ -89,6 +92,77 @@ def sa_token(service_account: ServiceAccount) -> str:
...
+    Adds the user to the group before the test, then removes them after.
+
+    Args:
+        admin_client: The admin client for accessing the cluster
+        test_idp_user_session: The test user session containing user information
+
+    Yields:
+        Group: The Model Registry group with the test user added
+    """
+    model_registry_users_group = f"{MR_INSTANCE_NAME}-users"
```
lugi0, line 134: might be worth having this parametrized rather than using the constant in case we want to expand this test for multi-tenancy and need to deal with the default groups of multiple MR instances

tests/model_registry/rbac/conftest.py
lugi0: I still see the try here, I think you can just yield and then have the update to revert it to an empty group after it

tests/model_registry/rbac/test_mr_rbac.py
```diff
@@ -0,0 +1,271 @@
...
+            "component_patch": {
+                DscComponents.MODELREGISTRY: {
+                    "managementState": DscComponents.ManagementState.MANAGED,
+                    "registriesNamespace": MR_NAMESPACE,
+                },
+            }
+        })
+    ],
+    indirect=True,
+    scope="class",
```
lugi0, line 40: isn't the scope already class on the fixture?

lugi0 (approved):
few suggestions but looks good in general

tests/model_registry/rbac/conftest.py
dbasunag: +1. Tests should pass the group name.

tests/model_registry/rbac/conftest.py
```diff
@@ -89,6 +92,77 @@ def sa_token(service_account: ServiceAccount) -> str:
...
+        client=admin_client,
+        name=model_registry_users_group,
+        wait_for_resource=True,
+    )
+
+    # Add user to group
+    group.update(
+        resource_dict={"metadata": {"name": model_registry_users_group}, "users": [test_idp_user_session.username]}
+    )
+    LOGGER.info(f"Added user {test_idp_user_session.username} to {model_registry_users_group} group")
```
dbasunag, line 145: You can use ResourceEditor here. That way you won't need to keep track of the clean up.

tests/model_registry/rbac/conftest.py
dbasunag: +1 no need for the try.

tests/model_registry/rbac/test_mr_rbac.py
dbasunag: +1, not needed.

tests/model_registry/rbac/test_mr_rbac.py
```diff
@@ -0,0 +1,271 @@
...
+        Raises:
+            AssertionError: If access permissions don't match expectations
+            ForbiddenException: Expected for normal users, unexpected for admin users
+        """
+
+        context_to_use = (
+            test_idp_user_session.original_context if use_admin_context else test_idp_user_session.user_context
+        )
+        LOGGER.info(
+            f"-----Testing Model Registry access for user '{test_idp_user_session.username}' "
```
dbasunag, line 93: Can you please remove the leading and trailing "----"

tests/model_registry/rbac/test_mr_rbac.py
```diff
@@ -0,0 +1,271 @@
...
+            model_registry_namespace: The namespace where Model Registry is deployed
+            admin_client: The admin client for accessing the cluster
+            test_idp_user_session: The test user session containing both admin and user contexts
+            mr_access_role: The Role that grants Model Registry access
+
+        Raises:
+            AssertionError: If group creation or access permissions don't match expectations
+        """
+        LOGGER.info(
+            "-----Test that a new group can be granted access to Model Registry and user added to it can access MR-----"
```
dbasunag, line 205: Can you please remove the leading and trailing "----" from all the log messages?

tests/model_registry/rbac/test_mr_rbac.py
```diff
@@ -0,0 +1,271 @@
...
+            admin_client: The admin client for accessing the cluster
+            test_idp_user_session: The test user session containing both admin and user contexts
+            model_registry_group_with_user: The Model Registry group with the test user added
+
+        Raises:
+            AssertionError: If access permissions don't match expectations
+            ForbiddenException: Expected before group addition, unexpected after
+        """
+        LOGGER.info(
+            "-----Test that a user can access to the Model Registry once added to a "
```
dbasunag, line 143: Ditto. Let's remove "----" from all the messages.

tests/model_registry/rbac/test_mr_rbac.py
```diff
@@ -0,0 +1,271 @@
...
+            "group that has the permissions to access it-----"
+        )
+        # Verify group membership
+        verify_group_membership(
+            group=model_registry_group_with_user,
```
dbasunag, line 148: Should be a fixture, since this is a pre-condition to your test.

tests/model_registry/rbac/test_mr_rbac.py
```diff
@@ -0,0 +1,271 @@
...
+            test_idp_user_session: The test user session containing both admin and user contexts
+            mr_access_role: The Role that grants Model Registry access
+
+        Raises:
+            AssertionError: If group creation or access permissions don't match expectations
+        """
+        LOGGER.info(
+            "-----Test that a new group can be granted access to Model Registry and user added to it can access MR-----"
+        )
+        LOGGER.info(f"Group {NEW_GROUP_NAME} created and user {test_idp_user_session.username} added to it")
```
dbasunag, line 207: You already printed similar message in new_group fixture. Do you still need it?

dbasunag (changes requested):
Couple of comments. Looks good otherwise. Thank you for addressing all the open comments.

tests/model_registry/rbac/utils.py
```diff
@@ -0,0 +1,117 @@
...
+    Returns:
+        A dictionary of arguments for ModelRegistryClient.
+
+    Note: Uses is_secure=False for testing purposes.
+    """
+    server, port = rest_endpoint.split(":")
+    return {
+        "server_address": f"{Protocols.HTTPS}://{server}",
+        "port": port,
+        "user_token": token,
```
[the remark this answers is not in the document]
fege: it is not needed

dbasunag (approved):

lugi0 (dismissed):

dbasunag (approved):

PR #289, opened 2025-05-05
feat: add option to run tests even if cluster sanity checks failed

lugi0:
## Merge criteria:

### DIFF
--- a/utilities/infra.py
+++ b/utilities/infra.py
@@ -871,7 +871,10 @@ def verify_cluster_sanity(
     junitxml_property: Callable[[str, object], None] | None = None,
 ) -> None:
     """
-    Check that cluster resources (Nodes, DSCI, DSC) are healthy and exists pytest execution on failure.
+    Always runs cluster sanity checks (Nodes, DSCI, DSC).
+     If checks fail:
+       - Default behavior: Logs error and exits.
+       - If --cluster-sanity-continue-on-failure is set: Logs warning and allows tests to run.
 
     Args:
         request (FixtureRequest): pytest request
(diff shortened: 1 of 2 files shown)


utilities/infra.py
```diff
@@ -881,35 +884,48 @@ def verify_cluster_sanity(
...
-        return
+    # Handle the option to completely skip the check first
+    if request.session.config.getoption(skip_check_opt):
+        LOGGER.warning(f"Skipping cluster sanity check entirely due to {skip_check_opt} flag.")
+        return  # Exit the function early, tests will run without checks
 
     try:
-        LOGGER.info("Check cluster sanity.")
-
+        LOGGER.info("Running cluster sanity check...")
```
dbasunag, line 900: I think we should not skip here. If the nodes are not schedulable or healthy. It is a cluster problem.

utilities/infra.py
```diff
@@ -881,35 +884,48 @@ def verify_cluster_sanity(
...
-
-        if junitxml_property:
-            junitxml_property(name="exit_code", value=return_code)  # type: ignore[call-arg]
-
-        # TODO: Write to file to easily report the failure in jenkins
-        pytest.exit(reason=error_msg, returncode=return_code)
+        # Check if the flag to continue *despite* failure is set
+        if request.session.config.getoption(continue_on_failure_opt):
+            # Flag is set: Log warning, then return normally to allow tests to run
+            LOGGER.warning(f"{error_msg}. Proceeding with tests anyway due to {continue_on_failure_opt} flag.")
```
dbasunag, line 916: We should make it error. It is still a failure and we are just choosing to continue.

utilities/infra.py
```diff
@@ -881,35 +884,48 @@ def verify_cluster_sanity(
...
+            # Optional: Add JUnit property to mark that sanity failed but tests continued
+            if junitxml_property:
+                junitxml_property(name="cluster_sanity_check_failed", value=True)  # type: ignore[call-arg]
+                junitxml_property(name="cluster_sanity_forced_continue", value=True)  # type: ignore[call-arg]
+            # Return normally from the fixture setup
```
dbasunag, line 921: I believe this is for test result reporting. If cluster sanity fails and we are continuing with tests, we should not mark the tests as failure because of sanity failure. They should be marked failed only if the test fails.

dbasunag (changes requested):

utilities/infra.py
lugi0: This is only marking those two specific properties, not the tests as failures

utilities/infra.py
lugi0: This is different from what was being discussed in slack. I do agree with you, but can we get an ack to proceed before implementing the change?

utilities/infra.py
dbasunag: Then the associated test would be marked as failed though (e.g. first test of model registry). Do you want that?

PR #296, opened 2025-05-06
Add test for Model Registry RBAC for SA token

lugi0:
## Merge criteria:

## Summary by CodeRabbit

- **New Features**
  - Added tests for Role-Based Access Control (RBAC) in the Model Registry using ServiceAccount tokens, including scenarios for both access denial and granted permissions.
  - Introduced test fixtures for automated setup and cleanup of Kubernetes namespaces, ServiceAccounts, roles, and role bindings to support RBAC testing.
  - Added utility functions for generating random names and namespace names for use in tests.

- **Bug Fixes**
  - Removed an unused import from the test setup to improve code cleanliness.

### DIFF
--- a/tests/model_registry/conftest.py
+++ b/tests/model_registry/conftest.py
@@ -1,11 +1,7 @@
 import pytest
 import re
-import random
-import string
-import subprocess
 import schemathesis
-import shlex
-from typing import Generator, Any, List, Dict
+from typing import Generator, Any
 from kubernetes.dynamic.exceptions import ResourceNotFoundError
 from ocp_resources.pod import Pod
 from ocp_resources.secret import Secret
@@ -14,12 +10,8 @@ from ocp_resources.service import Service
 from ocp_resources.persistent_volume_claim import PersistentVolumeClaim
 from ocp_resources.data_science_cluster import DataScienceCluster
 from ocp_resources.deployment import Deployment
-from ocp_resources.service_account import ServiceAccount
-from ocp_resources.role import Role
-from ocp_resources.role_binding import RoleBinding
 
 from ocp_resources.model_registry import ModelRegistry
-import schemathesis.schemas
 from schemathesis.specs.openapi.schemas import BaseOpenAPISchema
 from schemathesis.generation.stateful.state_machine import APIStateMachine
 from schemathesis.core.transport import Response
@@ -30,7 +22,6 @@ from pytest import FixtureRequest
 from simple_logger.logger import get_logger
 from kubernetes.dynamic import DynamicClient
 from pytest_testconfig import config as py_config
-from pyhelper_utils.shell import run_command
 from model_registry.types import RegisteredModel
 from tests.model_registry.constants import (
     MR_OPERATOR_NAME,
@@ -52,7 +43,6 @@ from model_registry import ModelRegistry as ModelRegistryClient
 
 
 LOGGER = get_logger(name=__name__)
-DEFAULT_TOKEN_DURATION = "10m"
 
 
 @pytest.fixture(scope="class")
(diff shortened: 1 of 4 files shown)


tests/model_registry/conftest.py
```diff
@@ -293,3 +302,194 @@ def model_registry_operator_pod(admin_client: DynamicClient) -> Pod:
...
+
+    LOGGER.info(f"Attempting to create Role: {role_name} with rules and labels.")
+    # Use context manager for creation and deletion
+    # Pass rules and labels directly
+    with Role(
+        client=admin_client,
+        name=role_name,
+        namespace=model_registry_namespace,
+        rules=role_rules,
+        label=role_labels,  # Pass labels via kwargs
```
dbasunag, line 435: ditto. please use ensure_exists=True instead.

tests/model_registry/conftest.py
```diff
@@ -293,3 +302,194 @@ def model_registry_operator_pod(admin_client: DynamicClient) -> Pod:
...
+
+    # Define labels, to be passed via **kwargs
+    role_labels = {
+        "app.kubernetes.io/component": "model-registry-test-rbac",
+        "test.opendatahub.io/namespace": sa_namespace.name,
+    }
+
+    LOGGER.info(f"Attempting to create Role: {role_name} with rules and labels.")
+    # Use context manager for creation and deletion
+    # Pass rules and labels directly
```
dbasunag, line 429: ditto

tests/model_registry/conftest.py
```diff
@@ -293,3 +302,194 @@ def model_registry_operator_pod(admin_client: DynamicClient) -> Pod:
...
+        label=role_labels,  # Pass labels via kwargs
+    ) as role:
+        try:
+            role.wait(timeout=60)  # Wait for role object to exist
+            LOGGER.info(f"Role {role.name} created successfully.")
+            yield role
+            LOGGER.info(f"Role {role.name} deletion initiated by context manager.")
+        except Exception as e:  # Catch other potential errors during Role instantiation or wait
+            LOGGER.error(f"Error during Role {role_name} creation or wait: {e}", exc_info=True)
+            pytest.fail(f"Failed during Role {role_name} creation: {e}")
```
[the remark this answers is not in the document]
dbasunag: resource.wait() exists in ocp-resources. And Role is a derived class from Resource class. While this code exists, we should be using ensure_exists=True

tests/model_registry/conftest.py
```diff
@@ -293,3 +302,194 @@ def model_registry_operator_pod(admin_client: DynamicClient) -> Pod:
...
+
+
+@pytest.fixture(scope="function")
+def mr_access_role_binding(
+    admin_client: DynamicClient,
+    model_registry_namespace: str,  # Existing fixture from main conftest
+    mr_access_role: Role,  # Depend on the role fixture to get its name
+    sa_namespace: Namespace,  # The namespace containing the test SA
```
dbasunag, line 452: You may want to clean up these comments. The names are self explanatory and comments are a bit too much.

tests/model_registry/conftest.py
```diff
@@ -293,3 +302,194 @@ def model_registry_operator_pod(admin_client: DynamicClient) -> Pod:
...
+) -> Generator[RoleBinding, None, None]:
+    """
+    Creates the MR Access RoleBinding using direct constructor parameters and a context manager.
+    """
+    binding_name = f"{mr_access_role.name}-binding"  # Simplify name slightly, role name is already unique
+    role_name_ref = mr_access_role.name  # Get the actual name from the created Role object
```
dbasunag, line 458: no need to save. Use directly.

tests/model_registry/conftest.py
```diff
@@ -293,3 +302,194 @@ def model_registry_operator_pod(admin_client: DynamicClient) -> Pod:
...
+
+    # Define labels, to be passed via **kwargs
+    binding_labels = {
+        "app.kubernetes.io/component": "model-registry-test-rbac",
+        "test.opendatahub.io/namespace": sa_namespace.name,
+    }
+
+    LOGGER.info(f"Attempting to create RoleBinding: {binding_name} with labels.")
+    # Use context manager for creation and deletion
+    # Pass subject and role_ref details directly to constructor
```
dbasunag, line 473: ditto

tests/model_registry/conftest.py
```diff
@@ -293,3 +302,194 @@ def model_registry_operator_pod(admin_client: DynamicClient) -> Pod:
...
+        label=binding_labels,  # Pass labels via kwargs
+    ) as binding:
+        try:
+            binding.wait(timeout=60)  # Wait for binding object to exist
+            LOGGER.info(f"RoleBinding {binding.name} created successfully.")
+            yield binding
+            LOGGER.info(f"RoleBinding {binding.name} deletion initiated by context manager.")
+        except Exception as e:  # Catch other potential errors
+            LOGGER.error(f"Error during RoleBinding {binding_name} creation or wait: {e}", exc_info=True)
+            pytest.fail(f"Failed during RoleBinding {binding_name} creation: {e}")
```
dbasunag, line 495: Let's move all these fixtures to rbac/conftest.py.

tests/model_registry/rbac/test_mr_rbac_sa.py
```diff
@@ -0,0 +1,145 @@
+# AI Disclaimer: Google Gemini 2.5 pro has been used to generate a majority of this code, with human review and editing.
+import pytest
+from typing import Self, Dict, Any
+from simple_logger.logger import get_logger
```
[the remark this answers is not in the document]
dbasunag: This is fine. The repo is currently pinning to 3.12

tests/model_registry/rbac/test_mr_rbac_sa.py
```diff
@@ -0,0 +1,145 @@
...
+        # Conftest example uses is_secure=False - this implies TLS verification might be off
+        # Adjust if your environment requires strict TLS verification
+        return {
+            "server_address": f"{Protocols.HTTPS}://{server}",
+            "port": port,
+            "user_token": token,
+            "is_secure": False,  # Match conftest example (disable TLS verification)
+            "author": author,
+        }
+    except Exception as e:
```
dbasunag, line 33: This seems not necessary. If you don't get rest_endpoint, may be in that fixture raise exception?

tests/model_registry/rbac/test_mr_rbac_sa.py
```diff
@@ -0,0 +1,145 @@
...
+        LOGGER.error(f"Error parsing REST endpoint '{rest_endpoint}': {e}")
+        raise ValueError(f"Could not parse REST endpoint: {rest_endpoint}") from e
+
+
+# --- Test Class ---
+
+
+# Parametrize to ensure Model Registry component is enabled via DSC fixture
```
dbasunag, line 41: We can remove these. Let's stick with existing coding style.

tests/model_registry/rbac/test_mr_rbac_sa.py
```diff
@@ -0,0 +1,145 @@
...
+    """
+    Tests RBAC for Model Registry REST endpoint using ServiceAccount tokens.
+    """
+
+    @pytest.mark.smoke
+    @pytest.mark.usefixtures("sa_namespace", "service_account")  # Need SA and NS for token
+    def test_service_account_access_denied(
+        self: Self,
+        sa_token: str,  # Get token for the test SA
+        model_registry_instance_rest_endpoint: str,  # REST endpoint fixture
```
dbasunag, line 71: Let's clean up the comments. They are distracting and you are already using really good and self explanatory names for fixtures.

tests/model_registry/rbac/test_mr_rbac_sa.py
```diff
@@ -0,0 +1,145 @@
...
+    indirect=True,
+    scope="class",
+)
+@pytest.mark.usefixtures("updated_dsc_component_state_scope_class")
+class TestModelRegistryRBAC:
+    """
+    Tests RBAC for Model Registry REST endpoint using ServiceAccount tokens.
+    """
+
+    @pytest.mark.smoke
```
dbasunag, line 66: smoke should be absolutely basic set of tests, which when broken would stop our release. This seems like a good candidate for sanity though.

tests/model_registry/rbac/test_mr_rbac_sa.py
```diff
@@ -0,0 +1,145 @@
...
+    ):
+        """
+        Verifies SA access is DENIED (403 Forbidden) by default via REST.
+        Does NOT use mr_access_role or mr_access_role_binding fixtures.
+        """
+        LOGGER.info("--- Starting RBAC Test: Access Denied ---")
+        LOGGER.info(f"Targeting Model Registry REST endpoint: {model_registry_instance_rest_endpoint}")
+        LOGGER.info("Expecting initial access DENIAL (403 Forbidden)")
+
+        try:
```
dbasunag, line 81: I don't believe you need this try/except.

tests/model_registry/rbac/test_mr_rbac_sa.py
```diff
@@ -0,0 +1,145 @@
...
+            assert http_error.body is not None, "HTTPError should have a response object"
+            LOGGER.info(f"Received expected HTTP error: Status Code {http_error.status}")
+            assert http_error.status == 403, f"Expected HTTP 403 Forbidden, but got {http_error.status}"
+            LOGGER.info("Successfully received expected HTTP 403 status code.")
+
+        except Exception as e:
+            LOGGER.error(f"Received unexpected error during 'access denied' check: {e}", exc_info=True)
+            pytest.fail(f"'Access denied' check failed unexpectedly: {e}")
+
+    @pytest.mark.smoke
```
dbasunag, line 103: ditto. This does not seem like a candidate for smoke.

tests/model_registry/rbac/test_mr_rbac_sa.py
```diff
@@ -0,0 +1,145 @@
...
+    # Use fixtures for SA/NS/Token AND the RBAC Role/Binding
+    @pytest.mark.usefixtures("sa_namespace", "service_account", "mr_access_role", "mr_access_role_binding")
+    def test_service_account_access_granted(
+        self: Self,
+        sa_token: str,  # Get token for the test SA
+        model_registry_instance_rest_endpoint: str,  # REST endpoint fixture
+        # mr_access_role and mr_access_role_binding are activated by @usefixtures
```
dbasunag, line 110: no need for this line.

tests/model_registry/rbac/test_mr_rbac_sa.py
```diff
@@ -0,0 +1,145 @@
...
+            if e.body is not None and e.status == 403:
+                pytest.fail(f"Still received HTTP 403 Forbidden after applying Role/RoleBinding: {e}")
+            else:
+                pytest.fail(f"Client interaction failed with HTTP error after granting permissions: {e}")
+
+        except Exception as e:
+            LOGGER.error(f"Received unexpected general error after granting access: {e}", exc_info=True)
+            pytest.fail(f"Client interaction failed unexpectedly after granting permissions: {e}")
+
+        LOGGER.info("--- RBAC Test Completed Successfully ---")
```
dbasunag, line 145: Do you need this part? Any exception here is bad and should be raised as they arise.

utilities/infra.py
```diff
@@ -871,7 +871,10 @@ def verify_cluster_sanity(
     junitxml_property: Callable[[str, object], None] | None = None,
 ) -> None:
     """
-    Check that cluster resources (Nodes, DSCI, DSC) are healthy and exists pytest execution on failure.
```
dbasunag, line 874: Let's keep these changes out of this PR.

dbasunag (commented):
Thanks a lot for creating these tests so fast @lugi0 ! I have a few comments. Mostly about reorganizing things, removing cluster sanity part and cleaning up some generated docs/comments.

tests/model_registry/conftest.py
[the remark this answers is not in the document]
fege: I think that `ensure_exists=True` is looking if the resource is already available , we should use `wait_for_resource=True` instead. FYI I am adding a similar/same fixture in `tests/model_registry/rbac/conftest.py` why you think is better in model_registry? do you imagine to create Roles and RoleBindings in other scenarios?

tests/model_registry/conftest.py
lugi0: I reuse it also in the logs so I think it's cleaner defining it here at the top, same with role_name

dbasunag: It is just a python object property. There is no point saving here.

lugi0: Are you talking about role_name_ref?

tests/model_registry/rbac/test_mr_rbac_sa.py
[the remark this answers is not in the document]
lugi0: This works correctly as is, and the rest_endpoint str is not going to change for the foreseeable future. In fact, if it does somehow change, I want this to fail so that I am aware.

tests/model_registry/conftest.py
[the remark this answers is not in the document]
dbasunag: @fege We should add generic fixtures that are used across multiple modules, in `tests/model_registry/conftest.py`. But RBAC ones are currently only used for RBAC tests. So they should be better suited in relevant module's conftest instead of `tests/model_registry/conftest.py`. 
Looking at the wrapper code, since we are using context manager, we are deploying the resource. Both`ensure_exists` and `wait_for_resource`, makes the same self.exist call, but one waits for it to exist and the first one simply check it exists. I would prefer your suggestion of using `wait_for_resource` for cases like this. Thank you!

tests/model_registry/rbac/conftest.py
```diff
@@ -0,0 +1,176 @@
...
+    Creates a temporary namespace using a context manager for automatic cleanup.
+    Function scope ensures a fresh namespace for each test needing it.
+    """
+    ns_name = generate_namespace_name(file_path=request.fspath.strpath.split(f"{os.path.dirname(__file__)}/")[1])
+    LOGGER.info(f"Creating temporary namespace: {ns_name}")
+    with Namespace(client=admin_client, name=ns_name) as ns:
+        ns.wait_for_status(status=Namespace.Status.ACTIVE, timeout=120)
+        LOGGER.info(f"Namespace {ns_name} is active.")
+        yield ns
+        LOGGER.info(f"Namespace {ns_name} deletion initiated by context manager.")
```
dbasunag, line 34: You don't need this. wrapper would already print the delete call.

tests/model_registry/rbac/conftest.py
```diff
@@ -0,0 +1,176 @@
...
+def sa_namespace(request: pytest.FixtureRequest, admin_client: DynamicClient) -> Generator[Namespace, None, None]:
+    """
+    Creates a temporary namespace using a context manager for automatic cleanup.
+    Function scope ensures a fresh namespace for each test needing it.
+    """
+    ns_name = generate_namespace_name(file_path=request.fspath.strpath.split(f"{os.path.dirname(__file__)}/")[1])
+    LOGGER.info(f"Creating temporary namespace: {ns_name}")
+    with Namespace(client=admin_client, name=ns_name) as ns:
+        ns.wait_for_status(status=Namespace.Status.ACTIVE, timeout=120)
+        LOGGER.info(f"Namespace {ns_name} is active.")
```
dbasunag, line 32: Same here. Sampler would print that it is waiting for active status.

tests/model_registry/rbac/conftest.py
```diff
@@ -0,0 +1,176 @@
...
+    """
+    Creates a ServiceAccount within the temporary namespace using a context manager.
+    Function scope ensures it's tied to the lifetime of sa_namespace for that test.
+    """
+    sa_name = generate_random_name(prefix="mr-test-user")
+    LOGGER.info(f"Creating ServiceAccount: {sa_name} in namespace {sa_namespace.name}")
+    with ServiceAccount(client=admin_client, name=sa_name, namespace=sa_namespace.name, wait_for_resource=True) as sa:
+        LOGGER.info(f"ServiceAccount {sa_name} created.")
+        yield sa
+        LOGGER.info(f"ServiceAccount {sa_name} deletion initiated by context manager.")
```
dbasunag, line 48: ditto.

tests/model_registry/rbac/conftest.py
```diff
@@ -0,0 +1,176 @@
...
+@pytest.fixture(scope="function")
+def service_account(admin_client: DynamicClient, sa_namespace: Namespace) -> Generator[ServiceAccount, None, None]:
+    """
+    Creates a ServiceAccount within the temporary namespace using a context manager.
+    Function scope ensures it's tied to the lifetime of sa_namespace for that test.
+    """
+    sa_name = generate_random_name(prefix="mr-test-user")
+    LOGGER.info(f"Creating ServiceAccount: {sa_name} in namespace {sa_namespace.name}")
+    with ServiceAccount(client=admin_client, name=sa_name, namespace=sa_namespace.name, wait_for_resource=True) as sa:
+        LOGGER.info(f"ServiceAccount {sa_name} created.")
```
dbasunag, line 46: Ditto.

tests/model_registry/rbac/conftest.py
```diff
@@ -0,0 +1,177 @@
...
+    Retrieves a short-lived token for the ServiceAccount using 'oc create token'.
+    Function scope because token is temporary and tied to the SA for that test.
+    """
+    sa_name = service_account.name
+    namespace = service_account.namespace
+    LOGGER.info(f"Retrieving token for ServiceAccount: {sa_name} in namespace {namespace}")
+    try:
+        cmd = f"oc create token {sa_name} -n {namespace} --duration={DEFAULT_TOKEN_DURATION}"
+        LOGGER.debug(f"Executing command: {cmd}")
+        res, out, err = run_command(command=shlex.split(cmd), verify_stderr=False, check=True, timeout=30)
```
dbasunag, line 63: I think we need to use hide_log_command option here: https://github.com/RedHatQE/pyhelper-utils/blob/main/pyhelper_utils/shell.py#L23. We don't want to display the token information in the console. Can you please double check?

lugi0: I can also confirm it does not print the token
```
2025-05-07T12:46:40.366859 conftest INFO Retrieving token for ServiceAccount: mr-test-user-b45f0c11 in namespace mr-rbac-sa
2025-05-07T12:46:40.367240 pyhelper_utils.shell INFO Running oc create token mr-test-user-b45f0c11 -n mr-rbac-sa --duration=10m command
2025-05-07T12:46:41.248802 conftest INFO Successfully retrieved token for SA 'mr-test-user-b45f0c11'
```

tests/model_registry/conftest.py
dbasunag: Right. But won't hold this PR for this comment.

tests/model_registry/rbac/conftest.py
dbasunag: yeah. if the token was displaying it could have given unauthorized user a window to access the cluster. Such data must be scrubbed from log. Thanks for confirming.

dbasunag (approved):

tests/model_registry/rbac/conftest.py
lugi0: It would give access to the MR instance, btw :)

dbasunag: For now and any unintended access is bad :)

tests/model_registry/rbac/conftest.py
```diff
@@ -0,0 +1,172 @@
...
+    """
+    role_name = f"registry-user-{MR_INSTANCE_NAME}-{sa_namespace.name[:8]}"
+    LOGGER.info(f"Defining Role: {role_name} in namespace {model_registry_namespace}")
+
+    role_rules: List[Dict[str, Any]] = [
+        {
+            "apiGroups": [""],
+            "resources": ["services"],
+            "resourceNames": [MR_INSTANCE_NAME],  # Grant access only to the specific MR service object
+            "verbs": ["get"],
```
fege, line 112: should we pass the verbs as input param so to have more flexibility?

tests/model_registry/rbac/conftest.py
```diff
@@ -0,0 +1,172 @@
...
+    with RoleBinding(
+        client=admin_client,
+        name=binding_name,
+        namespace=model_registry_namespace,
+        # Subject parameters
+        subjects_kind="Group",
+        subjects_name=f"system:serviceaccounts:{sa_namespace.name}",
+        subjects_api_group="rbac.authorization.k8s.io",  # This is the default apiGroup for Group kind
+        # Role reference parameters
+        role_ref_kind="Role",
```
fege, line 165: since we are passing already `mr_access_role` we could use it `mr_access_role.kind`

lugi0: yes, good idea. if you rebase in your PR can you add it?

tests/model_registry/rbac/conftest.py
lugi0: in our case we only really care about `get` on the MR Instance, so I don't think it's really needed. If we were to generalize this to create any Role then I'd agree with you

PR #297, opened 2025-05-06
Fix for post upgrade operator check

mwaykole:
The test logic has been updated because pure KServe raw deployments do not require dependent operators,

how to run test
`uv run pytest --pre-upgrade --upgrade-deployment-modes=rawdeployment --tc=distribution:downstream  --tc=dependent_operators:""`
`uv run pytest --post-upgrade --upgrade-deployment-modes=rawdeployment --tc=distribution:downstream  --tc=dependent_operators:""`

### DIFF
--- a/tests/model_serving/model_server/conftest.py
+++ b/tests/model_serving/model_server/conftest.py
@@ -573,28 +573,33 @@ def unprivileged_s3_caikit_serverless_inference_service(
 
 @pytest.fixture(scope="package")
 def fail_if_missing_dependent_operators(admin_client: DynamicClient) -> None:
-    missing_operators: list[str] = []
-    csvs = list(
-        ClusterServiceVersion.get(
-            dyn_client=admin_client,
-            namespace=py_config["applications_namespace"],
-        )
-    )
+    if dependent_operators := py_config.get("dependent_operators"):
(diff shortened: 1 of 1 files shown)


tests/model_serving/model_server/conftest.py
```diff
@@ -573,28 +573,33 @@ def unprivileged_s3_caikit_serverless_inference_service(
...
-                    missing_operators.append(
-                        f"Operator {operator_name} is installed but CSV is not in {csv.Status.SUCCEEDED} state"
-                    )
-
-        else:
-            missing_operators.append(f"{operator_name} is not installed")
-
-    if missing_operators:
-        pytest.fail(str(missing_operators))
+    if dependent_operators := py_config.get("dependent_operators"):
```
dbasunag, line 576: This fixture is supposed to be added only for serverless. https://github.com/opendatahub-io/opendatahub-tests/blob/main/conftest.py#L233. May be fix it there, so that it is not inserted there for not serverless upgrade cases?

mwaykole: No this is for all, raw and serverless. I have tested it

adolfo-ab (approved):

nehachopra27 (approved):

PR #301, opened 2025-05-09
feat: add TrustyAIService Prometheus integration tests

adolfo-ab:
Add tests to check the integration between TrustyAIService and Prometheus, for both drift and fairness metrics.

## How Has This Been Tested?
Running the tests against a working cluster.

## Merge criteria:

## Summary by CodeRabbit

- **Tests**
	- Added new tests to validate Prometheus metrics for drift and fairness in model explainability services.
	- Updated metric validation in model serving tests to use a more flexible validation approach.
	- Removed a redundant Prometheus fixture from model server tests.
	- Introduced a new Prometheus fixture for monitoring in test configurations.
- **Refactor**
	- Improved metric validation utilities by splitting functionality into modular functions for greater flexibility and clearer logging.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -14,6 +14,7 @@ from ocp_resources.node import Node
 from ocp_resources.pod import Pod
 from ocp_resources.secret import Secret
 from ocp_resources.service import Service
+from ocp_utilities.monitoring import Prometheus
 from pyhelper_utils.shell import run_command
 from pytest import FixtureRequest, Config
 from kubernetes.dynamic import DynamicClient
@@ -519,4 +520,14 @@ def cluster_sanity_scope_session(
         dsc_resource=dsc_resource,
         dsci_resource=dsci_resource,
         junitxml_property=junitxml_plugin,
+    )
+
+
+@pytest.fixture(scope="session")
+def prometheus(admin_client: DynamicClient) -> Prometheus:
+    return Prometheus(
+        client=admin_client,
+        resource_name="thanos-querier",
+        verify_ssl=False,
+        bearer_token=get_openshift_token(),
     )
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -24,6 +24,7 @@ from ocp_resources.resource import get_client
 from pytest_testconfig import config as py_config
 from simple_logger.logger import get_logger
 
+from utilities.certificates_utils import create_ca_bundle_file
 from utilities.data_science_cluster_utils import update_components_in_dsc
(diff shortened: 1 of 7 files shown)


utilities/monitoring.py
```diff
@@ -7,53 +7,82 @@
...
+        prometheus (Prometheus): Prometheus object
+        metrics_query (str): Metrics query string
+
+    Returns:
+        Any: Metrics value
+
+    """
+    metric_results = prometheus.query_sampler(query=metrics_query)
+    if metric_values_list := [value for metric_val in metric_results for value in metric_val.get("value")]:
+        return metric_values_list[1]
```
lugi0, line 24: why specifically [1]?

tests/conftest.py
```diff
@@ -520,3 +521,13 @@ def cluster_sanity_scope_session(
...
         junitxml_property=junitxml_plugin,
     )
+
+
+@pytest.fixture(scope="session")
+def prometheus(admin_client: DynamicClient) -> Prometheus:
+    return Prometheus(
+        client=admin_client,
+        resource_name="thanos-querier",
+        verify_ssl=False,
```
lugi0, line 531: if possible we should start implementing functions that validate SSL and grab the certs from the cluster if needed

lugi0 (dismissed):
couple questions but looks good overall

dbasunag (approved):

utilities/monitoring.py
adolfo-ab: This was implemented by @rnetser , I just moved it to the top of the file, maybe she can clarify

tests/conftest.py
mwaykole: +1

mwaykole (approved):

tests/conftest.py
adolfo-ab: I agree, but for that we need to change the `Prometheus` class in ocp_utilities, so I propose we merge this now with a TODO to handle ssl in the short term future.

adolfo-ab: NVM, I used the openshift ca bundle and it works fine, just need a change of type hints in ocp_utilities

utilities/monitoring.py
rnetser: `prometheus.query_sampler` returns the query result and there should be 1 value there

lugi0: so shouldn't it be [0]? @rnetser

dbasunag (approved):

PR #304, opened 2025-05-12
Set temp value to 0 to reduce the randomness of the output for vllm output

tarukumar:
## Summary by CodeRabbit

- **Refactor**
  - Adjusted the temperature setting for AI requests to always use a value of 0, including for embeddings, ensuring consistent behavior across all supported endpoints.


jiridanek:
> reduce the randomness

only reduce? shouldn't this completely eliminate any randomness from the output?

dbasunag (approved):

PR #306, opened 2025-05-14
fix: increase timeouts in LMEval tests and fixtures

adolfo-ab:
Increases the timeouts in LMEval tests and fixtures

## Description
The 10 minutes timeout used is sometimes not enough, therefore increasing it significantly to avoid flakiness

## How Has This Been Tested?
Running the tests in a working cluster

## Merge criteria:

## Summary by CodeRabbit

- **Chores**
  - Increased timeout durations from 10 minutes to 20 minutes for various resource readiness checks in model explainability evaluation tests.
  - Added a new 20-minute timeout constant for improved configuration consistency.

### DIFF
--- a/tests/model_explainability/lm_eval/conftest.py
+++ b/tests/model_explainability/lm_eval/conftest.py
@@ -204,7 +204,7 @@ def lmeval_data_downloader_pod(
         restart_policy="Never",
         volumes=[{"name": "pvc-volume", "persistentVolumeClaim": {"claimName": "lmeval-data"}}],
     ) as pod:
-        pod.wait_for_status(status=Pod.Status.SUCCEEDED, timeout=Timeout.TIMEOUT_10MIN)
+        pod.wait_for_status(status=Pod.Status.SUCCEEDED, timeout=Timeout.TIMEOUT_20MIN)
         yield pod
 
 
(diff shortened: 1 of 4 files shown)


tests/model_explainability/lm_eval/conftest.py
```diff
@@ -204,7 +204,7 @@ def lmeval_data_downloader_pod(
         restart_policy="Never",
         volumes=[{"name": "pvc-volume", "persistentVolumeClaim": {"claimName": "lmeval-data"}}],
     ) as pod:
-        pod.wait_for_status(status=Pod.Status.SUCCEEDED, timeout=Timeout.TIMEOUT_10MIN)
+        pod.wait_for_status(status=Pod.Status.SUCCEEDED, timeout=Timeout.TIMEOUT_20MIN)
```
dbasunag, line 207: Do we know why they are taking so long to be in succeeded state? It is going to add significant run time at your end.

adolfo-ab: In most scenarios the 10 minute timeout is more than enough, but in some of the environments during release testing we hit the timeout. When I log into these clusters and run the tests manually, they work, so my best guess is that these clusters have less resources, or pulling the image is slower for some reason. I propose we set this timeout for the next release, and then I will be able to see if we still hit the timeout, and see if it's too big (maybe 15mins is enough).

dbasunag (approved):

tests/model_explainability/lm_eval/conftest.py
mwaykole, line 207: hey @adolfo-ab  have u tried with 15 min , or 20 min is the best option ?

mwaykole (approved):

dbasunag (approved):

PR #307, opened 2025-05-15
feat: add check on authorino and its channel/version for MR tests

lugi0:
## Merge criteria:

## Summary by CodeRabbit

- **Tests**
	- Added a new automated check to ensure the Authorino operator is installed and meets version and channel requirements before running tests in the model registry module. Tests will fail if these conditions are not met, providing clearer feedback on missing or misconfigured dependencies.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -225,15 +225,10 @@ def pytest_runtest_setup(item: Item) -> None:
     1. Updates global config (`updated_global_config`)
     2. Adds `fail_if_missing_dependent_operators` fixture for Serverless tests.
     3. Adds fixtures to enable KServe/model mesh in DSC for model server tests.
-    4. Adds `fail_if_missing_authorino_operator` fixture for model registry tests.
     """
 
     BASIC_LOGGER.info(f"\n{separator(symbol_='-', val=item.name)}")
     BASIC_LOGGER.info(f"{separator(symbol_='-', val='SETUP')}")
-
-    # Add Authorino operator check for model registry tests
-    if "model_registry" in str(item.fspath):
-        item.fixturenames.insert(0, "fail_if_missing_authorino_operator")
 
     if KServeDeploymentType.SERVERLESS.lower() in item.keywords:
         item.fixturenames.insert(0, "fail_if_missing_dependent_operators")
--- a/conftest.py
+++ b/conftest.py
@@ -225,10 +225,15 @@ def pytest_runtest_setup(item: Item) -> None:
     1. Updates global config (`updated_global_config`)
     2. Adds `fail_if_missing_dependent_operators` fixture for Serverless tests.
     3. Adds fixtures to enable KServe/model mesh in DSC for model server tests.
+    4. Adds `fail_if_missing_authorino_operator` fixture for model registry tests.
     """
 
     BASIC_LOGGER.info(f"\n{separator(symbol_='-', val=item.name)}")
     BASIC_LOGGER.info(f"{separator(symbol_='-', val='SETUP')}")
+
+    # Add Authorino operator check for model registry tests
+    if "model_registry" in str(item.fspath):
+        item.fixturenames.insert(0, "fail_if_missing_authorino_operator")
 
     if KServeDeploymentType.SERVERLESS.lower() in item.keywords:
         item.fixturenames.insert(0, "fail_if_missing_dependent_operators")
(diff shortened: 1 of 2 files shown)


fege (approved):
/lgtm

tests/model_registry/conftest.py
```diff
@@ -292,3 +294,60 @@ def model_registry_operator_pod(admin_client: DynamicClient) -> Pod:
...
+
+    LOGGER.info(f"Verifying if {operator_name} is installed with required version and channel")
+
+    # Find the CSV for the operator
+    authorino_csvs = [csv for csv in csvs if csv.name.startswith(operator_name)]
+    if not authorino_csvs:
+        pytest.fail(f"{operator_name} is not installed")
+
+    authorino_csv = authorino_csvs[0]
+    if authorino_csv.status != authorino_csv.Status.SUCCEEDED:
```
dbasunag, line 322: This is a bad idea. Please add a wait_for_status. We use stable channel - which means we don't control authorino upgrades. We don't require a customer to wait for upgrade to complete, instead, upgrade is not supposed to be disruptive. Your csv can be in many states that won't mean a bad situation - Installing, Pending etc.

conftest.py
```diff
@@ -225,11 +225,16 @@ def pytest_runtest_setup(item: Item) -> None:
...
+    4. Adds `fail_if_missing_authorino_operator` fixture for model registry tests.
     """
 
     BASIC_LOGGER.info(f"\n{separator(symbol_='-', val=item.name)}")
     BASIC_LOGGER.info(f"{separator(symbol_='-', val='SETUP')}")
 
+    # Add Authorino operator check for model registry tests
+    if "model_registry" in str(item.fspath):
+        item.fixturenames.insert(0, "fail_if_missing_authorino_operator")
+
```
dbasunag, line 237: This is a hacky way. Is there any reason to not have `autouse=True`.  https://docs.pytest.org/en/stable/how-to/fixtures.html#autouse-fixtures-fixtures-you-don-t-have-to-request

lugi0: I've done the same as the other calls already here, which ensures this fixture is run for any test under model_registry but not for any other (which is what we want).

tests/model_registry/conftest.py
lugi0: I don't think this will ever be an issue, as we won't be the ones triggering the upgrade, which is almost instantaneous. If for some reason the operator is stuck upgrading once we get to this code I think it should fail, as it is in a state we cannot trust

conftest.py
dbasunag: I know model serving did it this way because of the complexity of their need, but our use case is simpler and can use autouse.

tests/model_registry/conftest.py
dbasunag: When you are using production catalog source, if the subscription defines `Automatic approval of install plan` you don't need to trigger upgrade - it would be fully automatic. This is what I see from authorino subscription:
```
apiVersion: operators.coreos.com/v1alpha1
kind: Subscription
metadata:
  creationTimestamp: "2025-04-09T13:40:21Z"
  generation: 1
  labels:
    operators.coreos.com/authorino-operator.openshift-operators: ""
  name: authorino-operator
  namespace: openshift-operators
  resourceVersion: "59192882"
  uid: e605afdf-2a43-4829-aef6-3e3d009f2fa2
spec:
  channel: stable
  installPlanApproval: Automatic
  name: authorino-operator
  source: redhat-operators
  sourceNamespace: openshift-marketplace
  startingCSV: authorino-operator.v1.2.1
status:
```
We should keep our check to reasonable level. A CSV not being in SUCCEEDED state does not necessarily mean the cluster is not usable with respect to openshift AI. Kubernetes is about eventual consistency. If we don't get to succeeded state after a period of time, then it `may` indicate bad situation. But a momentary state change does not indicate bad state. Tests should only be halted when we don't have reasonable expectation of success. I would either drop the CSV status check or add a wait for succeeded call.

dbasunag: It is not just about a production update. Authorino is not a managed operator of RHOAI. There is no requirement from our side that the associated CSV has to be in SUCCEEDED state - all the time, for MR to work. We should only fail when requirements are not met.

dbasunag (approved):

fege (approved):
/lgtm

PR #311, opened 2025-05-20
add test to validate container images for MR, improve ODH compatibility

lugi0:
## Merge criteria:

### DIFF
--- a/tests/model_registry/conftest.py
+++ b/tests/model_registry/conftest.py
@@ -32,6 +32,7 @@ from tests.model_registry.constants import (
     DB_RESOURCES_NAME,
     MODEL_REGISTRY_DB_SECRET_STR_DATA,
     MODEL_REGISTRY_DB_SECRET_ANNOTATIONS,
+    init_mr_namespace,
 )
 from tests.model_registry.utils import (
     get_endpoint_from_mr_service,
@@ -228,10 +229,25 @@ def updated_dsc_component_state_scope_class(
     request: FixtureRequest, dsc_resource: DataScienceCluster, admin_client: DynamicClient
 ) -> Generator[DataScienceCluster, Any, Any]:
     original_components = dsc_resource.instance.spec.components
-    with ResourceEditor(patches={dsc_resource: {"spec": {"components": request.param["component_patch"]}}}):
-        for component_name in request.param["component_patch"]:
+    component_patch = request.param["component_patch"]
+
+    # Check if ModelRegistry component is already Managed and patch includes registriesNamespace
+    if (
+        DscComponents.MODELREGISTRY in component_patch
+        and original_components[DscComponents.MODELREGISTRY]["managementState"] == DscComponents.ManagementState.MANAGED
+        and "registriesNamespace" in component_patch[DscComponents.MODELREGISTRY]
+    ):
+        LOGGER.warning(
+            "Cannot patch registriesNamespace when ModelRegistry component is already set to Managed state. "
+            "Current namespace will be used."
+        )
+        # Remove registriesNamespace from patch to avoid the update
+        component_patch[DscComponents.MODELREGISTRY].pop("registriesNamespace")
+
+    with ResourceEditor(patches={dsc_resource: {"spec": {"components": component_patch}}}):
+        for component_name in component_patch:
             dsc_resource.wait_for_condition(condition=DscComponents.COMPONENT_MAPPING[component_name], status="True")
-        if request.param["component_patch"].get(DscComponents.MODELREGISTRY):
+        if component_patch.get(DscComponents.MODELREGISTRY):
             namespace = Namespace(
                 name=dsc_resource.instance.spec.components.modelregistry.registriesNamespace, ensure_exists=True
             )
@@ -243,19 +259,27 @@ def updated_dsc_component_state_scope_class(
         )
         yield dsc_resource
 
-    for component_name, value in request.param["component_patch"].items():
-        LOGGER.info(f"Waiting for component {component_name} to be updated.")
-        if original_components[component_name]["managementState"] == DscComponents.ManagementState.MANAGED:
-            dsc_resource.wait_for_condition(condition=DscComponents.COMPONENT_MAPPING[component_name], status="True")
-        if (
-            component_name == DscComponents.MODELREGISTRY
-            and value.get("managementState") == DscComponents.ManagementState.MANAGED
-        ):
-            # Since namespace specified in registriesNamespace is automatically created after setting
-            # managementStateto Managed. We need to explicitly delete it on clean up.
-            namespace = Namespace(name=value["registriesNamespace"], ensure_exists=True)
-            if namespace:
-                namespace.delete(wait=True)
+    # I don't understand this code block. If the component is set to MANAGED, the namespace cannot be deleted.
+    # This should only be done if 1) the component was set to Removed originally, 2) we set it to Managed, and
+    # 3) we decide to reset it to Removed and delete the namespace. Commenting out for now otherwise the namespace
+    # deletion hangs for 4 minutes on every teardown and doesn't work.
+
+    # for component_name, value in component_patch.items():
+    #     LOGGER.info(f"Waiting for component {component_name} to be updated.")
+    #     if original_components[component_name]["managementState"] == DscComponents.ManagementState.MANAGED:
+    #         dsc_resource.wait_for_condition(condition=DscComponents.COMPONENT_MAPPING[component_name], status="True")
+    #     if (
+    #         component_name == DscComponents.MODELREGISTRY
+    #         and value.get("managementState") == DscComponents.ManagementState.MANAGED
+    #     ):
+    #         # Since namespace specified in registriesNamespace is automatically created after setting
+    #         # managementStateto Managed. We need to explicitly delete it on clean up.
+    #         namespace = Namespace(
+    #             name=dsc_resource.instance.spec.components.modelregistry.registriesNamespace,
+    #             ensure_exists=True
+    #         )
+    #         if namespace:
+    #             namespace.delete(wait=True)
 
 
 @pytest.fixture(scope="class")
@@ -378,3 +402,13 @@ def fail_if_missing_authorino_operator(admin_client: DynamicClient) -> None:
 
     if not subscription_found:
         pytest.fail(f"Could not find Subscription for {operator_name} in openshift-operators namespace")
+
+
+@pytest.fixture(scope="session", autouse=True)
+def init_model_registry_namespace(admin_client: DynamicClient) -> None:
+    """Initialize MR_NAMESPACE with the admin client.
+
+    This fixture runs automatically at session start to ensure MR_NAMESPACE
+    is properly initialized before any tests run.
+    """
+    init_mr_namespace(admin_client=admin_client)
--- a/tests/model_registry/constants.py
+++ b/tests/model_registry/constants.py
@@ -1,13 +1,15 @@
 from typing import Any
 from ocp_resources.resource import Resource
 from utilities.constants import ModelFormat
+from kubernetes.dynamic import DynamicClient
+from ocp_resources.data_science_cluster import DataScienceCluster
+from pytest_testconfig import config as py_config
 
 
 class ModelRegistryEndpoints:
     REGISTERED_MODELS: str = "/api/model_registry/v1alpha3/registered_models"
 
 
-MR_NAMESPACE: str = "rhoai-model-registries"
 MR_OPERATOR_NAME: str = "model-registry-operator"
 MODEL_NAME: str = "my-model"
 MODEL_DICT: dict[str, Any] = {
@@ -45,3 +47,61 @@ MODEL_REGISTRY_DB_SECRET_ANNOTATIONS = {
     f"{Resource.ApiGroup.TEMPLATE_OPENSHIFT_IO}/expose-password": "'{.data[''database-password'']}'",
     f"{Resource.ApiGroup.TEMPLATE_OPENSHIFT_IO}/expose-username": "'{.data[''database-user'']}'",
 }
+
+
+def get_mr_namespace(admin_client: DynamicClient) -> str:
+    """Get the Model Registry namespace from the DSC instance.
+
+    Args:
+        admin_client: The kubernetes client
+
+    Returns:
+        The namespace for Model Registry as specified in the DSC instance
+
+    Raises:
+        ValueError: If the namespace cannot be found in the DSC instance
+    """
+    dsc = DataScienceCluster(client=admin_client, name=py_config["dsc_name"], ensure_exists=True)
+    if not dsc.instance or not dsc.instance.spec.components.modelregistry:
+        raise ValueError("ModelRegistry component not found in DSC instance")
+
+    namespace = dsc.instance.spec.components.modelregistry.registriesNamespace
+    if not namespace:
+        raise ValueError("registriesNamespace not set in ModelRegistry component")
+
+    return namespace
+
+
+class LazyNamespace:
+    """Lazy evaluation wrapper for MR_NAMESPACE to avoid circular imports."""
+
+    def __init__(self) -> None:
+        self._namespace = None
+        self._admin_client = None
+
+    def __str__(self) -> str:
+        if self._namespace is None:
+            if self._admin_client is None:
+                raise ValueError("admin_client must be set before accessing MR_NAMESPACE")
+            self._namespace = get_mr_namespace(admin_client=self._admin_client)
+        return self._namespace
+
+    def set_client(self, admin_client: DynamicClient) -> None:
+        """Set the admin client to be used when resolving the namespace."""
+        self._admin_client = admin_client
+
+
+# Create a singleton instance
+_MR_NAMESPACE = LazyNamespace()
+
+# Export the singleton instance as MR_NAMESPACE
+MR_NAMESPACE = _MR_NAMESPACE
+
+
+# Function to initialize the namespace with a client
+def init_mr_namespace(admin_client: DynamicClient) -> None:
+    """Initialize MR_NAMESPACE with an admin client.
+
+    This should be called early in test setup, typically in a session-scoped fixture.
+    """
+    _MR_NAMESPACE.set_client(admin_client=admin_client)
--- a/tests/model_registry/image_validation/test_verify_rhoai_images.py
+++ b/tests/model_registry/image_validation/test_verify_rhoai_images.py
@@ -20,7 +20,7 @@ LOGGER = get_logger(name=__name__)
             "component_patch": {
                 DscComponents.MODELREGISTRY: {
                     "managementState": DscComponents.ManagementState.MANAGED,
-                    "registriesNamespace": "odh-model-registries",
+                    "registriesNamespace": MR_NAMESPACE,
                 }
             }
         }
@@ -96,6 +96,11 @@ class TestModelRegistryImages:
             if not is_valid:
                 validation_errors.append(f"Pod {instance_pod.name} image validation failed: {error_msg}")
 
+            # If it's a sidecar image defined correctly (comes from registry and uses sha256 digest)
+            # we don't need to check that it is in our relatedImages
+            if "openshift-service-mesh" in image:
+                LOGGER.warning(f"Skipping image {image} as it is a service mesh sidecar image")
+                continue
             # Check if image is in relatedImages
             if image not in related_image_refs:
                 validation_errors.append(f"Pod {instance_pod.name} image {image} is not listed in CSV's relatedImages")
--- a/tests/model_registry/image_validation/test_verify_rhoai_images.py
+++ b/tests/model_registry/image_validation/test_verify_rhoai_images.py
@@ -1,0 +1,104 @@
+import pytest
+from typing import Self
+from simple_logger.logger import get_logger
+from kubernetes.dynamic import DynamicClient
+from ocp_resources.pod import Pod
+from pytest_testconfig import config as py_config
+
+from utilities.constants import DscComponents, Labels
+from utilities.general import get_csv_related_images, get_pod_images, validate_image_format
+from tests.model_registry.constants import MR_OPERATOR_NAME, MR_INSTANCE_NAME, MR_NAMESPACE
+from ocp_resources.model_registry import ModelRegistry
+
+LOGGER = get_logger(name=__name__)
+
+
+@pytest.mark.parametrize(
+    "updated_dsc_component_state_scope_class",
+    [
+        {
+            "component_patch": {
+                DscComponents.MODELREGISTRY: {
+                    "managementState": DscComponents.ManagementState.MANAGED,
+                    "registriesNamespace": "odh-model-registries",
+                }
+            }
+        }
+    ],
+    indirect=True,
+)
+@pytest.mark.usefixtures("updated_dsc_component_state_scope_class")
+class TestModelRegistryImages:
+    """
+    Tests to verify that all Model Registry component images (operator and instance container images)
+    meet the requirements:
+    1. Images are hosted in registry.redhat.io
+    2. Images use sha256 digest instead of tags
+    3. Images are listed in the CSV's relatedImages section
+    """
+
+    @pytest.mark.smoke
+    def test_verify_model_registry_images(
+        self: Self, admin_client: DynamicClient, model_registry_instance: ModelRegistry
+    ):
+        # Get related images from CSV
+        related_images = get_csv_related_images(admin_client=admin_client)
+        related_image_refs = {img["image"] for img in related_images}
+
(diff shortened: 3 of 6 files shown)


lugi0:
This has been verified against an ODH nightly install, will verify against a RHOAI install ASAP

tests/model_registry/conftest.py
```diff
@@ -228,10 +229,25 @@ def updated_dsc_component_state_scope_class(
...
+        DscComponents.MODELREGISTRY in component_patch
+        and original_components[DscComponents.MODELREGISTRY]["managementState"] == DscComponents.ManagementState.MANAGED
+        and "registriesNamespace" in component_patch[DscComponents.MODELREGISTRY]
+    ):
+        LOGGER.warning(
+            "Cannot patch registriesNamespace when ModelRegistry component is already set to Managed state. "
+            "Current namespace will be used."
+        )
+        # Remove registriesNamespace from patch to avoid the update
+        component_patch[DscComponents.MODELREGISTRY].pop("registriesNamespace")
```
dbasunag, line 245: Why would we do this. As of now MR is in Removed state in DSC, if it is enabled, we should fail, instead of working around it. No?

tests/model_registry/conftest.py
```diff
@@ -378,3 +402,13 @@ def fail_if_missing_authorino_operator(admin_client: DynamicClient) -> None:
...
+
+
+@pytest.fixture(scope="session", autouse=True)
+def init_model_registry_namespace(admin_client: DynamicClient) -> None:
+    """Initialize MR_NAMESPACE with the admin client.
+
+    This fixture runs automatically at session start to ensure MR_NAMESPACE
+    is properly initialized before any tests run.
+    """
+    init_mr_namespace(admin_client=admin_client)
```
dbasunag, line 414: We should only do this if MR is enabled by default.

dbasunag (changes requested):
We need to discuss about this PR offline.

tests/model_registry/conftest.py
lugi0: We should not assume the state of the component, since 1) it is bound to change ~soon, and 2) in your local environment you might have it enabled already

dbasunag: Exactly. Right now MR is TP, if it is enabled by default, that should be considered a bug and your change would hide that.

tests/model_registry/conftest.py
```diff
@@ -243,19 +259,27 @@ def updated_dsc_component_state_scope_class(
...
+    #         and value.get("managementState") == DscComponents.ManagementState.MANAGED
+    #     ):
+    #         # Since namespace specified in registriesNamespace is automatically created after setting
+    #         # managementStateto Managed. We need to explicitly delete it on clean up.
+    #         namespace = Namespace(
+    #             name=dsc_resource.instance.spec.components.modelregistry.registriesNamespace,
+    #             ensure_exists=True
+    #         )
+    #         if namespace:
+    #             namespace.delete(wait=True)
```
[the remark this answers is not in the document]
lugi0: This does not work if MR is set to Managed when the test starts, it needs to be updated if we want to keep and I'm personally not even sure it is worth keeping either way

tests/model_registry/conftest.py
lugi0: registriesNamespace is defined either way

tests/model_registry/conftest.py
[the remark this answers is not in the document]
dbasunag: If MR is set to managed, when test starts, we should fail. So we need to keep it as it is, unless product behavior changes that requires removal of this code. We should discuss offline.

tests/model_registry/conftest.py
lugi0: updated to fail if we are using RHOAI and it's managed. I want to stress that I also disagree with this, we should not be testing the operator inner logic in our component-specific tests.

dbasunag: If TP feature is supposed to be disabled by default, the the component team has a responsibility to ensure it is not enabled by default. Customers needs to explicitly enable TP features. This code should be removed.

tests/model_registry/conftest.py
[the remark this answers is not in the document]
dbasunag: MR should not be managed. It is still TP, so unless someone is running tests agianst a cluster that has it manually enabled, we should not hit error at this point.

tests/model_registry/constants.py
```diff
@@ -45,3 +47,61 @@ class ModelRegistryEndpoints:
...
+    def set_client(self, admin_client: DynamicClient) -> None:
+        """Set the admin client to be used when resolving the namespace."""
+        self._admin_client = admin_client
+
+
+# Create a singleton instance
+_MR_NAMESPACE = LazyNamespace()
+
+# Export the singleton instance as MR_NAMESPACE
+MR_NAMESPACE = _MR_NAMESPACE
```
dbasunag, line 98: Should always be in a function or fixture.

tests/model_registry/image_validation/test_verify_rhoai_images.py
```diff
@@ -0,0 +1,109 @@
...
+        operator_pod = operator_pods[0]
+
+        # Get MR instance pod(s)
+        instance_pods = list(
+            Pod.get(
+                dyn_client=admin_client,
+                namespace=MR_NAMESPACE,
+                label_selector=f"app={MR_INSTANCE_NAME}",
+            )
+        )
```
dbasunag, line 69: We are testing how OLM works here.  I believe if CSV has the right images, we can assume the pods/deployments are good. Otherwise, we can write a parameterize test to check all the pods. You have quite a bit of similar code here.

utilities/general.py
```diff
@@ -171,3 +179,75 @@ def create_isvc_label_selector_str(isvc: InferenceService, resource_type: str, r
...
+    Args:
+        admin_client: The kubernetes client
+        csv_name: Optional CSV name. If not provided, will use {operator_name}.{version}
+                 where operator_name is determined by the distribution (rhods-operator for OpenShift AI,
+                 opendatahub-operator for Open Data Hub)
+
+    Returns:
+        List of related images from the CSV
+    """
+    from utilities.infra import (
```
dbasunag, line 196: You can import utilities.infra to avoid circular import

lugi0: I can't, because infra imports utilities.general

lugi0:
Updated to remove most ODH logic. Verified against ODH by manually setting the required config values. Will verify against RHOAI asap.

tests/model_registry/image_validation/test_verify_rhoai_images.py
lugi0: I have implemented it like this because this is the original requirement that was logged in Jira. If the requirements are updated we can always remove this logic and have a single test case for every component that validates all relatedImages in the CSV.

tests/model_registry/image_validation/test_verify_rhoai_images.py
```diff
@@ -0,0 +1,88 @@
...
+
+        # Get instance pod
+        instance_pod_list = get_pods_by_labels(
+            admin_client=admin_client,
+            namespace=MR_NAMESPACE,
+            label_selector=f"app={MR_INSTANCE_NAME}",
+        )
+        if len(instance_pod_list) > 1:
+            LOGGER.warning(f"Expected 1 instance pod, found {len(instance_pod_list)}. Using first one.")
+        instance_pod = instance_pod_list[0]
```
dbasunag, line 69: We have very similar codes here. Can we have utils that gets the pods and validates their images?

tests/model_registry/image_validation/test_verify_rhoai_images.py
```diff
@@ -0,0 +1,88 @@
...
+
+        # Get operator pod
+        operator_pod_list = get_pods_by_labels(
+            admin_client=admin_client,
+            namespace=py_config["applications_namespace"],
+            label_selector=f"{Labels.OpenDataHubIo.NAME}={MR_OPERATOR_NAME}",
+        )
+        if len(operator_pod_list) > 1:
+            LOGGER.warning(f"Expected 1 operator pod, found {len(operator_pod_list)}. Using first one.")
+        operator_pod = operator_pod_list[0]
```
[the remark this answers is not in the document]
lugi0: yes, but I've defined the helper function to return a list of pods matching a label so I need to check that we have 1 here

tests/model_registry/image_validation/test_verify_rhoai_images.py
lugi0: IMHO this is good as is, the getting of the pods and image validation are already defined as external utils, the code in the test takes care of test-specific steps, i.e. how many pods I expect to find with my label selector

utilities/general.py
```diff
@@ -171,3 +180,144 @@ def create_isvc_label_selector_str(isvc: InferenceService, resource_type: str, r
...
+def get_pod_images(pod: Pod) -> List[str]:
+    """Get all container images from a pod.
+
+    Args:
+        pod: The pod to get images from
+
+    Returns:
+        List of container image strings
+    """
+    return [container.image for container in pod.instance.spec.containers]
```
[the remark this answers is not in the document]
lugi0: Not sure I understand?

utilities/general.py
```diff
@@ -1,13 +1,22 @@
...
 from ocp_resources.pod import Pod
 from simple_logger.logger import get_logger
+from ocp_resources.cluster_service_version import ClusterServiceVersion
+from pytest_testconfig import config as py_config
 
 import utilities.infra
 from utilities.constants import Annotations, KServeDeploymentType, MODELMESH_SERVING
 
+# Constants for image validation
+REGISTRY_REDHAT_IO = "registry.redhat.io"
```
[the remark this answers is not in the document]
lugi0: We can, but are we sure that reference will never change in the wrapper? Also worth considering the fact that this makes the code less self-explanatory for anyone who is not familiar with the wrapper.

fege (dismissed):
/lgtm

utilities/general.py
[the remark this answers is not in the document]
dbasunag: We use wrapper extensively. So anyone looking at our code base would need to be familiar with wrapper. When you use Resource.ApiGroup. REGISTRY_REDHAT_IO, it is pretty easy to infer. Wrapper does not change things often, and it is the same risk we take with any other library.

tests/model_registry/image_validation/test_verify_rhoai_images.py
dbasunag: This can be simplified. As you can see you have same lines of code that calls get_pods_by_labels() and checks the length. Also, if we don't expect >1 pods, should not we fail?

tests/model_registry/image_validation/test_verify_rhoai_images.py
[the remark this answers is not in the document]
dbasunag: Then should not we should assert on >1 pods.

utilities/general.py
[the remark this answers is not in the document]
lugi0: FWIW the wrapper calls it "IMAGE_REGISTRY", which to me is less easy to parse at a glance, but I've already updated it

tests/model_registry/image_validation/test_verify_rhoai_images.py
lugi0: What do you mean by simplified? This code still needs to live somewhere. It is highly unlikely that these same steps will be needed elsewhere other than this test, so to me it makes the most sense to have the code here rather than abstracted in yet another function that only gets called from here

tests/model_registry/image_validation/test_verify_rhoai_images.py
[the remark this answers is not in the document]
lugi0: see response below, debatable if we should fail or not on more than 1 pod

utilities/general.py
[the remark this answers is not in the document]
lugi0: Those are environment variables, not necessarily what the containers will end up using (e.g. the oauth-proxy image which is not currently used). I believe checking these is outside the scope of this test which cares about the images the containers are using.

tests/model_registry/image_validation/test_verify_rhoai_images.py
dbasunag: Let's add a wait then.

utilities/general.py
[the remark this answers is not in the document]
lugi0: These are just environment variables, what we care about are the images used by the containers. If they match there's no need to check them twice, if they don't match we still only care about the images actually used by the deployed containers as those should be in the relatedImages field of the CSV, since it's the source of truth even for these envs

tests/model_registry/image_validation/test_verify_rhoai_images.py
lugi0: done

tests/model_registry/image_validation/test_verify_rhoai_images.py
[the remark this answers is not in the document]
lugi0: done

utilities/general.py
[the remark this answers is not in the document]
dbasunag: I am not entirely sure about this. I have asked for clarification from Marek. We however don't need to block this PR for that and proceed with what you added. If it is deemed important to add that check, we can do it in a follow up PR.

tests/model_registry/conftest.py
```diff
@@ -229,10 +229,12 @@ def updated_dsc_component_state_scope_class(
     request: FixtureRequest, dsc_resource: DataScienceCluster, admin_client: DynamicClient
 ) -> Generator[DataScienceCluster, Any, Any]:
     original_components = dsc_resource.instance.spec.components
-    with ResourceEditor(patches={dsc_resource: {"spec": {"components": request.param["component_patch"]}}}):
-        for component_name in request.param["component_patch"]:
+    component_patch = request.param["component_patch"]
+
+    with ResourceEditor(patches={dsc_resource: {"spec": {"components": component_patch}}}):
+        for component_name in component_patch:
```
dbasunag, line 235: You don't need this change, but I have no issues with it being here either. Take this as a FYI and resolve if this is intentional.

utilities/general.py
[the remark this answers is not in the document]
lugi0: Sounds good, resolving here

tests/model_registry/image_validation/test_verify_rhoai_images.py
```diff
@@ -0,0 +1,103 @@
...
+                func=get_pods_by_labels,
+                admin_client=admin_client,
+                namespace=MR_NAMESPACE,
+                label_selector=f"app={MR_INSTANCE_NAME}",
+            ):
+                if len(instance_pod_list) == 1:
+                    instance_pod = instance_pod_list[0]
+                    break
+        except TimeoutExpiredError:
+            pytest.fail("Failed to find exactly one instance pod after 60 seconds")
```
dbasunag: You can have a utility to simplify this. 
```
errors = []
for label in [ label1, label2]:
      errors.extend(validate_pod_image(label=label1, namespace= py_config[model_registry_namespace]))
if errors:
     pytest.fail(.....)


def validate_pod_image(label: str, namespace: str) -> str:
     try:      
         for pod_list in TimeoutSampler(
                wait_timeout=60,
                sleep=5,
                func=get_pods_by_labels,
                admin_client=admin_client,
                namespace=namespace,
                label_selector=f"app={label}",
            ):
                if len(pod_list) == 1:
                    pod = pod_list[0]
                    return ""
    except TimeoutExpiredError:
          return "your error"
```
This is just a skeleton, I am sure you can write it better :)

utilities/general.py
```diff
@@ -171,3 +180,144 @@ def create_isvc_label_selector_str(isvc: InferenceService, resource_type: str, r
 
     else:
         raise ValueError(f"Unknown deployment mode {deployment_mode}")
+
+
+def get_csv_related_images(admin_client: DynamicClient, csv_name: str | None = None) -> List[Dict[str, str]]:
```
dbasunag, line 185: Let's move it to operators_utils.py

utilities/general.py
```diff
@@ -171,3 +180,144 @@ def create_isvc_label_selector_str(isvc: InferenceService, resource_type: str, r
...
+    if not csvs:
+        raise ValueError(f"CSV {csv_name} not found in namespace {py_config['applications_namespace']}")
+
+    csv = csvs[0]  # Get the first (and should be only) CSV instance
+    if len(csvs) > 1:
+        LOGGER.warning(f"Multiple CSV instances found for {csv_name}, using the first one")
+    if not csv.instance:
+        raise ValueError(f"CSV {csv_name} has no instance data")
+
+    return csv.instance.spec.relatedImages
```
dbasunag, line 224: You can utilize get_cluster_service_version()

utilities/general.py
```diff
@@ -171,3 +180,144 @@ def create_isvc_label_selector_str(isvc: InferenceService, resource_type: str, r
...
+    """
+    pods = list(
+        Pod.get(
+            dyn_client=admin_client,
+            namespace=namespace,
+            label_selector=label_selector,
+        )
+    )
+    if not pods:
+        pytest.fail(f"No pods found with label selector {label_selector} in namespace {namespace}")
```
dbasunag, line 284: Since it is an util, we should raise ResourceNotFoundError. Caller can fail then when such exceptions are raised.

dbasunag (changes requested):

tests/model_registry/image_validation/test_verify_rhoai_images.py
```diff
@@ -0,0 +1,103 @@
...
+        validation_errors.extend(
+            validate_container_images(
+                pod=instance_pod,
+                valid_image_refs=related_image_refs,
+                skip_patterns=["openshift-service-mesh"],
+            )
+        )
+
+        if validation_errors:
+            pytest.fail("\n".join(validation_errors))
```
dbasunag, line 103: This can be simplified. https://github.com/opendatahub-io/opendatahub-tests/pull/311#discussion_r2102862955

utilities/operator_utils.py
```diff
@@ -46,3 +49,31 @@ def validate_operator_subscription_channel(
...
+        csv_name: Optional CSV name. If not provided, will use {operator_name}.{version}
+                 where operator_name is determined by the distribution (rhods-operator for OpenShift AI,
+                 opendatahub-operator for Open Data Hub)
+
+    Returns:
+        List of related images from the CSV
+    """
+    from utilities.infra import (
+        get_product_version,
+    )  # Import here to avoid circular dependency
```
dbasunag, line 68: This should not be relevant anymore.

tests/model_registry/conftest.py
```diff
@@ -289,14 +293,38 @@ def registered_model(request: FixtureRequest, model_registry_client: ModelRegist
...
+            wait_timeout=60,
+            sleep=5,
+            func=Pod.get,
+            dyn_client=admin_client,
+            namespace=py_config["applications_namespace"],
+        ):
+            for pod in pod_list:
+                if re.match(MR_OPERATOR_NAME, pod.name):
+                    return pod
+    except TimeoutExpiredError:
```
dbasunag, line 308: You are returning the 1st match. Don't you want to wait for the number of matches to be 1? You can also potentially reuse get_pods_by_labels() here.

tests/model_registry/conftest.py
```diff
@@ -289,14 +293,38 @@ def registered_model(request: FixtureRequest, model_registry_client: ModelRegist
...
+
+
+@pytest.fixture()
+def model_registry_instance_pod(admin_client: DynamicClient) -> Pod:
+    """Get the model registry instance pod."""
+    try:
+        for instance_pod_list in TimeoutSampler(
+            wait_timeout=60,
+            sleep=5,
+            func=get_pods_by_labels,
```
dbasunag, line 319: If not needed in model_registry_operator_pod(), then you don't need a function. You can directly call Pod.get() with label_selector.

dbasunag (commented):
Looks good! Thank you!

fege (dismissed):
/lgtm

tests/model_registry/conftest.py
```diff
@@ -289,14 +293,38 @@ def registered_model(request: FixtureRequest, model_registry_client: ModelRegist
...
 def model_registry_operator_pod(admin_client: DynamicClient) -> Pod:
-    model_registry_operator_pods = [
-        pod
-        for pod in Pod.get(dyn_client=admin_client, namespace=py_config["applications_namespace"])
-        if re.match(MR_OPERATOR_NAME, pod.name)
-    ]
-    if not model_registry_operator_pods:
+    """Get the model registry operator pod."""
+    try:
+        for pod_list in TimeoutSampler(
```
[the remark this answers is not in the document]
lugi0: My understanding is that we were standardizing on TimeoutSampler for such calls. Did this change?

sheltoncyril (approved):

tests/model_registry/conftest.py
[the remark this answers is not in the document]
dbasunag: Timeout sampler suports using `@retry` decorator, for simple use cases (similar to the one you have). https://github.com/RedHatQE/timeout-sampler/blob/main/README.md#usage.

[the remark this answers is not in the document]
lugi0: I'm not sure I'm understanding the benefit of shifting from the inline call to TimeoutSampler to a decorator. It is the same amount of code, performing the same action, just moved slightly. Am I missing something?

utilities/general.py
```diff
@@ -171,3 +180,106 @@ def create_isvc_label_selector_str(isvc: InferenceService, resource_type: str, r
...
+    """
+    pods = list(
+        Pod.get(
+            dyn_client=admin_client,
+            namespace=namespace,
+            label_selector=label_selector,
+        )
+    )
+    if not pods:
+        raise ResourceNotFoundError(f"No pods found with label selector {label_selector} in namespace {namespace}")
```
dbasunag, line 244: You would need to add these exceptions to the decorator exceptions_dict. Otherwise, they would fail immediately if the conditions are met. Just the same way timeout sampler would, if you raise exceptions.

utilities/general.py
```diff
@@ -171,3 +180,106 @@ def create_isvc_label_selector_str(isvc: InferenceService, resource_type: str, r
...
+        return False, f"Image {image} is not from {Resource.ApiGroup.IMAGE_REGISTRY}"
+
+    if not re.search(SHA256_DIGEST_PATTERN, image):
+        return False, f"Image {image} does not use sha256 digest"
+
+    return True, ""
+
+
+@retry(wait_timeout=60, sleep=5)
+def get_pods_by_labels(
```
dbasunag, line 216: wait_for_pods_by_lebels()

tests/model_registry/image_validation/test_verify_rhoai_images.py
```diff
@@ -0,0 +1,59 @@
...
+        model_registry_instance: ModelRegistry,
+        model_registry_operator_pod: Pod,
+        model_registry_instance_pod: Pod,
+        related_images_refs: Set[str],
+    ):
+        validation_errors = []
+        for pod in [model_registry_operator_pod, model_registry_instance_pod]:
+            validation_errors.extend(
+                validate_container_images(
+                    pod=pod, valid_image_refs=related_images_refs, skip_patterns=["openshift-service-mesh"]
```
dbasunag, line 54: Which of these contains the skip_pattern?

tests/model_registry/conftest.py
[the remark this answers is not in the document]
dbasunag: https://github.com/RedHatQE/timeout-sampler/blob/main/README.md If you take a look at the read me. There is significant improvement in readability and number of lines of code when we use the decorator. However, decorator can't be used everywhere as it only fits simpler usecases.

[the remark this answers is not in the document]
lugi0: yes, I've already looked at the README. It is still not clear to me why we would prefer one over the other, especially since we use TimeoutSampler in its context manager form everywhere else - if we want to standardize the code we should pick one and stick with it, as neither option gives a benefit over the other. 
AFAIK any context manager can also be implemented as a decorator, so this issue is not specific to TimeoutSampler. We could rewrite any `with` statement as a decorator, which I would be against, so I don't understand the need to do it here.
I've implemented it just because I want this PR merged since it has been in review for over a week and it has been reviewed multiple times. Can we please move this forward?
If you have other concerns about style we can work on a style guide in a separate issue and after it is approved by all we can start nitpicking PRs based on that.

tests/model_registry/image_validation/test_verify_rhoai_images.py
dbasunag: Do we have images with skip_patterns `openshift-service-mesh` for both the operator and instance pod?

dbasunag (approved):

PR #312, opened 2025-05-21
Add MR namespace to global config and override it for ODH cluster

dbasunag:
## Merge criteria:

## Summary by CodeRabbit

- **New Features**
  - Added dynamic configuration for the model registry namespace, allowing tests to use runtime values instead of static constants.

- **Refactor**
  - Updated tests to use configuration-driven values for namespaces, improving flexibility and maintainability.
  - Centralized retrieval of cluster resources for more consistent test setup.
  - Introduced utility functions for operator version and subscription channel validation, enhancing reliability and clarity.
  - Simplified operator validation logic with improved distribution checks and exit conditions.
  - Replaced manual operator version and subscription checks with reusable utilities for better error handling and logging.

- **Chores**
  - Removed unused command-line options and constants to streamline configuration and reduce redundancy.
  - Cleaned up imports and reorganized utility functions for better modularity.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -34,6 +34,7 @@ from utilities.infra import (
     get_operator_distribution,
     login_with_user_password,
     get_openshift_token,
+    get_data_science_cluster,
 )
 from utilities.constants import (
     AcceleratorType,
@@ -66,17 +67,19 @@ def tests_tmp_dir(request: FixtureRequest, tmp_path_factory: TempPathFactory) ->
 
 @pytest.fixture(scope="session")
 def updated_global_config(request: FixtureRequest, admin_client: DynamicClient) -> None:
-    if get_operator_distribution(client=admin_client) == "Open Data Hub":
+    distribution = get_operator_distribution(client=admin_client)
+    if distribution == "Open Data Hub":
         py_config["distribution"] = "upstream"
 
-    else:
+    elif distribution.startswith("OpenShift AI"):
         py_config["distribution"] = "downstream"
-
-    if applications_namespace := request.config.getoption("applications_namespace"):
-        py_config["applications_namespace"] = applications_namespace
-
     else:
-        py_config["applications_namespace"] = get_dsci_applications_namespace(client=admin_client)
+        pytest.exit(f"Unknown distribution: {distribution}")
+
+    py_config["applications_namespace"] = get_dsci_applications_namespace(client=admin_client)
+    py_config["model_registry_namespace"] = get_data_science_cluster(
+        client=admin_client
+    ).instance.spec.components.modelregistry.registriesNamespace
 
 
 @pytest.fixture(scope="session")
--- a/tests/global_config.py
+++ b/tests/global_config.py
@@ -1,12 +1,14 @@
 global config  # type:ignore[unused-ignore]
 
-distribution: str = "downstream"
-applications_namespace: str = "redhat-ods-applications"  # overwritten in conftest.py if distribution is upstream
 dsc_name: str = "default-dsc"
 must_gather_base_dir: str = "must-gather-base-dir"
 dsci_name: str = "default-dsci"
 dependent_operators: str = "servicemeshoperator,authorino-operator,serverless-operator"
 use_unprivileged_client: bool = True
+# overwrite the followings in conftest.py, in updated_global_config() if distribution is upstream
+distribution: str = "downstream"
+applications_namespace: str = "redhat-ods-applications"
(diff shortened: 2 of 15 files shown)


mwaykole (approved):

tests/conftest.py
```diff
@@ -72,11 +73,10 @@ def updated_global_config(request: FixtureRequest, admin_client: DynamicClient)
     else:
         py_config["distribution"] = "downstream"
```
lugi0, line 74: if we want to make sure the distribution names in the DSC don't change without us noticing, I would update this to 
```
elif get_operator_distribution(client=admin_client) == "OpenShift AI":
    py_config["distribution"] = "downstream"
else:
    raise UnknownDistribution (or pytest.fail(f"Unknown distribution value: {get_operator_distribution(client=admin_client)}") or LOGGER.error(f"Unknown distribution value: {get_operator_distribution(client=admin_client)}", defaulting to downstream) etc.)
```

tests/global_config.py
```diff
@@ -1,12 +1,14 @@
 global config  # type:ignore[unused-ignore]
 
-distribution: str = "downstream"
-applications_namespace: str = "redhat-ods-applications"  # overwritten in conftest.py if distribution is upstream
 dsc_name: str = "default-dsc"
 must_gather_base_dir: str = "must-gather-base-dir"
 dsci_name: str = "default-dsci"
 dependent_operators: str = "servicemeshoperator,authorino-operator,serverless-operator"
 use_unprivileged_client: bool = True
+# overwrite the followings in conftest.py if distribution is upstream
```
lugi0, line 8: I would specify `tests/conftest.py`, maybe even point to the fixture name (`updated_global_config`)

lugi0 (approved):
A couple nitpicks, otherwise LGTM

tests/conftest.py
dbasunag: Not necessary. By default we have distribution as "downstream" we only set it to "upstream" in updated_global_config() if the dsc tell's us that it is open data hub. No other values are possible.

tests/global_config.py
dbasunag: Going to keep it for now. Since technically we are invoking updated_global_config in conftest.py - which is causing the override.

utilities/operator_utils.py
```diff
@@ -0,0 +1,46 @@
...
+) -> None:
+    operator_subscriptions = [
+        subscription
+        for subscription in Subscription.get(dyn_client=client, namespace=namespace)
+        if subscription.instance.spec.name == operator_name
+    ]
+
+    if not operator_subscriptions:
+        raise ResourceNotFoundError(f"Subscription not found for operator {operator_name}")
+    subscription_channel = operator_subscriptions[0].instance.spec.channel
```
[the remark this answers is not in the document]
dbasunag: This should not happen, as far as I know, but it won't hurt to add a check.

utilities/infra.py
```diff
@@ -770,20 +770,24 @@ def get_product_version(admin_client: DynamicClient) -> Version:
     return Version.parse(version=operator_version)
 
 
+def get_data_science_cluster(client: DynamicClient, dsc_name: str = "default-dsc") -> DataScienceCluster:
+    dsc = DataScienceCluster(client=client, name=dsc_name)
```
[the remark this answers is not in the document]
dbasunag: I will still keep the function, as I am using it in two places.

adolfo-ab (approved):

tests/conftest.py
lugi0: What I mean is that `get_operator_distribution` fetches the value from the DSC, and if there's a new value we don't know of it will set this to "downstream" when we maybe didn't want it to. It would be better to check explicitly that it's returning openshift ai before setting it to downstream.

tests/global_config.py
lugi0: Yes, and I mean that this file should point to that fixture, since if you look at this you don't really know where the override is happening

tests/conftest.py
dbasunag:
```   
if get_operator_distribution(client=admin_client) == "Open Data Hub":
        py_config["distribution"] = "upstream"

    else:
        py_config["distribution"] = "downstream"
``` this is coming from updated_global_config. If you believe it is a possibility to have a different value, then we should set it there. I will add it.

lugi0 (dismissed):

lugi0 (approved):

adolfo-ab (approved):

PR #313, opened 2025-05-22
Validate whether TrustyAI operator, service and oauth-proxy images match the ones in the configmap

sheltoncyril:
Check if TrustyAI operator deployment oauth-proxy
and trustyai-service pods are using the right images from the configmap in redhat-ods-applications namespace.

Makes sure the TrustyAI deployment and service pods are created with the right images using the configmap as the source of truth. 
Adresses [RHOAIENG-22502](https://issues.redhat.com/browse/RHOAIENG-22502) and [RHOAIENG-23499](https://issues.redhat.com/browse/RHOAIENG-23499)

## Description

Added a new test which validates whether the deployment/service pods have the same data in the image fields as the configmap. 

## How Has This Been Tested?

Test run successfully on a PSI cluster.

## Merge criteria:

## Summary by CodeRabbit

- **Tests**
  - Added a new test to verify that TrustyAI Operator and Service pods use the correct container images as specified in the configuration.
  - Introduced a fixture for accessing the TrustyAI Operator configuration.
  - Implemented utility logic to validate container images against expected values from the configuration.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -34,6 +34,7 @@ from utilities.infra import (
     get_operator_distribution,
     login_with_user_password,
     get_openshift_token,
+    get_data_science_cluster,
 )
 from utilities.constants import (
     AcceleratorType,
@@ -66,17 +67,19 @@ def tests_tmp_dir(request: FixtureRequest, tmp_path_factory: TempPathFactory) ->
 
 @pytest.fixture(scope="session")
 def updated_global_config(request: FixtureRequest, admin_client: DynamicClient) -> None:
-    if get_operator_distribution(client=admin_client) == "Open Data Hub":
+    distribution = get_operator_distribution(client=admin_client)
+    if distribution == "Open Data Hub":
         py_config["distribution"] = "upstream"
 
-    else:
+    elif distribution.startswith("OpenShift AI"):
         py_config["distribution"] = "downstream"
-
-    if applications_namespace := request.config.getoption("applications_namespace"):
-        py_config["applications_namespace"] = applications_namespace
-
     else:
-        py_config["applications_namespace"] = get_dsci_applications_namespace(client=admin_client)
+        pytest.exit(f"Unknown distribution: {distribution}")
+
+    py_config["applications_namespace"] = get_dsci_applications_namespace(client=admin_client)
+    py_config["model_registry_namespace"] = get_data_science_cluster(
+        client=admin_client
+    ).instance.spec.components.modelregistry.registriesNamespace
 
 
 @pytest.fixture(scope="session")
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -119,6 +119,16 @@ def model_namespace(
             yield ns
 
 
+@pytest.fixture(scope="class")
+def trustyai_operator_configmap(
+    request: FixtureRequest,
+    pytestconfig: pytest.Config,
+    admin_client: DynamicClient,
+    teardown_resources: bool,
+) -> ConfigMap:
+    return ConfigMap(client=admin_client, namespace="redhat-ods-applications", name="trustyai-service-operator-config")
+
+
 @pytest.fixture(scope="session")
 def aws_access_key_id(pytestconfig: Config) -> str:
     access_key = pytestconfig.option.aws_access_key_id
--- a/tests/model_explainability/trustyai_service/test_trustyai_service.py
+++ b/tests/model_explainability/trustyai_service/test_trustyai_service.py
@@ -1,7 +1,11 @@
 import pytest
+from ocp_resources.config_map import ConfigMap
 from ocp_resources.namespace import Namespace
 
-from tests.model_explainability.trustyai_service.utils import validate_trustyai_service_db_conn_failure
+from tests.model_explainability.trustyai_service.utils import (
+    validate_trustyai_service_db_conn_failure,
+    validate_trustyai_operator_and_service_images,
+)
 
 
 @pytest.mark.parametrize(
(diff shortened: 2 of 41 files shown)


sheltoncyril:
@coderabbitai review

tests/model_explainability/trustyai_service/utils.py
```diff
@@ -167,3 +170,33 @@ def create_trustyai_service(
...
+    if service_pods:
+        if len(service_pods) > 1:
+            raise TooManyPodsError("More than one pod found in TrustyAIService.")
+        for container_status in service_pods[0].instance.status.containerStatuses:
+            if container_status.name == "oauth-proxy":
+                assert configmap_data_dict["oauthProxyImage"] == container_status.image
+            elif container_status.name == TRUSTYAI_SERVICE_NAME:
+                assert configmap_data_dict["trustyaiServiceImage"] == container_status.image
+            else:
+                raise TooManyPodsError(f"An unknown/unhandled service pod {container_status.name} was found.")
```
lugi0, line 202: this would trigger if there are too many containers in the pod rather than too many pods, which is checked in L194

tests/conftest.py
```diff
@@ -119,6 +119,16 @@ def model_namespace(
...
 
 
+@pytest.fixture(scope="class")
+def trustyai_operator_configmap(
+    request: FixtureRequest,
+    pytestconfig: pytest.Config,
+    admin_client: DynamicClient,
+    teardown_resources: bool,
+) -> ConfigMap:
+    return ConfigMap(client=admin_client, namespace="redhat-ods-applications", name="trustyai-service-operator-config")
```
adolfo-ab, line 129: Few things here:

1. yield instead of return
2. use `ensure_exists` to raise if the CM doesn't exist
3. use applications_namespace from pyconfig instead of `redhat-ods-applications`
4. use TRUSTYAI_SERVICE_NAME constant in the name
5. I think you don't need teardown_resources here

tests/model_explainability/trustyai_service/test_trustyai_service.py
```diff
@@ -25,3 +29,27 @@ def test_trustyai_service_with_invalid_db_cert(
...
+@pytest.mark.parametrize(
+    "model_namespace",
+    [
+        pytest.param(
+            {"name": "test-trustyai-operator-and-service-images"},
+        )
+    ],
+    indirect=True,
+)
+def test_trustyai_operator_and_service_images(
```
adolfo-ab, line 43: The tests should be as atomic as possible, so we should have 2 tests, one for the operator and one for the service (maybe encapsulated in a common test class `TestTrustyAIImages` or similar)

tests/model_explainability/trustyai_service/utils.py
```diff
@@ -20,6 +20,9 @@
 
 LOGGER = get_logger(name=__name__)
 
+TRUSTYAI_OPERATOR_NAMESPACE = "redhat-ods-applications"
```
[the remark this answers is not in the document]
adolfo-ab: +1

tests/model_explainability/trustyai_service/utils.py
```diff
@@ -20,6 +20,9 @@
...
+TRUSTYAI_OPERATOR_DEPLOYMENT_NAME = "trustyai-service-operator-controller-manager"
```
adolfo-ab, line 24:
```suggestion
TRUSTYAI_OPERATOR_DEPLOYMENT_NAME = f"{TRUSTYAI_SERVICE_NAME}-operator-controller-manager"
```

tests/model_explainability/trustyai_service/utils.py
```diff
@@ -167,3 +170,33 @@ def create_trustyai_service(
         if wait_for_replicas:
             trustyai_deployment.wait_for_replicas()
         yield trustyai_service
+
+
+def validate_trustyai_operator_and_service_images(
```
adolfo-ab, line 175: Same, we need to split this into 2 functions, one for the service and one for the operator

tests/model_explainability/trustyai_service/test_trustyai_service.py
```diff
@@ -25,3 +29,27 @@ def test_trustyai_service_with_invalid_db_cert(
...
+    admin_client,
+    current_client_token,
+    model_namespace: Namespace,
+    trustyai_operator_configmap: ConfigMap,
+    trustyai_service_with_db_storage,
```
adolfo-ab, line 48: Don't use the service with db storage, use `trustyai_service_with_pvc_storage` instead. We don't want to deploy MariaDB for this test, since that will make this test substantially slower for no reason.

tests/model_explainability/trustyai_service/utils.py
```diff
@@ -167,3 +170,33 @@ def create_trustyai_service(
...
+    client: DynamicClient, model_namespace: str, configmap_data_dict: dict[str, str]
+) -> None:
+    """Validate if the TrustyAI operator deployment and service pods use the correct images from the configmap."""
+    trustyai_deployment = Deployment(
```
adolfo-ab, line 179:
```suggestion
    trustyai_operator_deployment = Deployment(
```

adolfo-ab (changes requested):

tests/model_explainability/trustyai_service/test_trustyai_service.py
```diff
@@ -25,3 +29,27 @@ def test_trustyai_service_with_invalid_db_cert(
         namespace=model_namespace,
         label_selector=f"app.kubernetes.io/instance={trustyai_service_with_invalid_db_cert.name}",
     )
+
+
+@pytest.mark.parametrize(
```
adolfo-ab, line 34: Add smoke tag

adolfo-ab (changes requested):

tests/conftest.py
sheltoncyril: I've made the changes but if you don't mind me asking, why yield if we are only fetching information and do not aim to modify it? I understand if there is some transactional way of doing things and we need to commit at the end but in this case a return would be more apt, right?

sheltoncyril:
Closing

PR #320, opened 2025-05-28
test: add test to initiate mr with secure db connection

fege:
## Merge criteria:

### DIFF
--- a/tests/model_registry/rest_api/conftest.py
+++ b/tests/model_registry/rest_api/conftest.py
@@ -2,7 +2,8 @@ from typing import Any, Generator
 import os
 from kubernetes.dynamic import DynamicClient
 import pytest
-from tests.model_registry.rest_api.constants import MODEL_REGISTRY_BASE_URI
+import copy
+from tests.model_registry.rest_api.constants import MODEL_REGISTRY_BASE_URI, MODEL_REGISTER_DATA
 from tests.model_registry.rest_api.utils import (
     register_model_rest_api,
     execute_model_registry_patch_command,
@@ -11,7 +12,11 @@ from utilities.constants import Protocols
 from ocp_resources.deployment import Deployment
 from tests.model_registry.utils import (
     get_model_registry_deployment_template_dict,
+    apply_mysql_args_and_volume_mounts,
+    add_mysql_certs_volumes_to_deployment,
+    generate_random_name,
 )
+
 from tests.model_registry.constants import (
     DB_RESOURCES_NAME,
     CA_MOUNT_PATH,
@@ -250,51 +255,20 @@ def patch_mysql_deployment_with_ssl_ca(
     """
     CA_CONFIGMAP_NAME = request.param.get("ca_configmap_name", "mysql-ca-configmap")
     CA_MOUNT_PATH = request.param.get("ca_mount_path", "/etc/mysql/ssl")
-    deployment = Deployment(
-        client=admin_client,
-        name=model_registry_db_deployment.name,
-        namespace=model_registry_namespace,
-    )
-    deployment.wait_for_condition(condition="Available", status="True")
-    original_deployment = deployment.instance.to_dict()
-    spec = original_deployment["spec"]["template"]["spec"]
+    deployment = model_registry_db_deployment.instance.to_dict()
+    spec = deployment["spec"]["template"]["spec"]
     my_sql_container = next(container for container in spec["containers"] if container["name"] == "mysql")
     assert my_sql_container is not None, "Mysql container not found"
-    mysql_args = list(my_sql_container.get("args", []))
-    mysql_args.extend([
-        f"--ssl-ca={CA_MOUNT_PATH}/ca/ca-bundle.crt",
-        f"--ssl-cert={CA_MOUNT_PATH}/server_cert/tls.crt",
-        f"--ssl-key={CA_MOUNT_PATH}/server_key/tls.key",
-    ])
 
-    volumes_mounts = list(my_sql_container.get("volumeMounts", []))
-    volumes_mounts.extend([
-        {"name": CA_CONFIGMAP_NAME, "mountPath": f"{CA_MOUNT_PATH}/ca", "readOnly": True},
-        {
-            "name": "mysql-server-cert",
-            "mountPath": f"{CA_MOUNT_PATH}/server_cert",
-            "readOnly": True,
-        },
-        {
-            "name": "mysql-server-key",
-            "mountPath": f"{CA_MOUNT_PATH}/server_key",
-            "readOnly": True,
-        },
-    ])
-
-    my_sql_container["args"] = mysql_args
-    my_sql_container["volumeMounts"] = volumes_mounts
-    volumes = list(spec["volumes"])
-    volumes.extend([
-        {"name": CA_CONFIGMAP_NAME, "configMap": {"name": CA_CONFIGMAP_NAME}},
-        {"name": "mysql-server-cert", "secret": {"secretName": "mysql-server-cert"}},  # pragma: allowlist secret
-        {"name": "mysql-server-key", "secret": {"secretName": "mysql-server-key"}},  # pragma: allowlist secret
-    ])
+    my_sql_container = apply_mysql_args_and_volume_mounts(
+        my_sql_container=my_sql_container, ca_configmap_name=CA_CONFIGMAP_NAME, ca_mount_path=CA_MOUNT_PATH
+    )
+    volumes = add_mysql_certs_volumes_to_deployment(spec=spec, ca_configmap_name=CA_CONFIGMAP_NAME)
 
     patch = {"spec": {"template": {"spec": {"volumes": volumes, "containers": [my_sql_container]}}}}
-    with ResourceEditor(patches={deployment: patch}):
-        deployment.wait_for_condition(condition="Available", status="True")
-        yield deployment
+    with ResourceEditor(patches={model_registry_db_deployment: patch}):
+        model_registry_db_deployment.wait_for_condition(condition="Available", status="True")
+        yield model_registry_db_deployment
 
 
 @pytest.fixture(scope="class")
@@ -359,3 +333,17 @@ def mysql_ssl_secrets(
         "server_cert_secret": server_cert_secret,
         "server_key_secret": server_key_secret,
     }
+
+
+@pytest.fixture(scope="function")
+def model_data_for_test() -> Generator[dict[str, Any], None, None]:
+    """
+    Generates a model data for the test.
+
+    Returns:
+        dict[str, Any]: The model data for the test
+    """
+    model_name = generate_random_name(prefix="model-rest-api")
+    model_data = copy.deepcopy(MODEL_REGISTER_DATA)
+    model_data["register_model_data"]["name"] = model_name
+    yield model_data
--- a/tests/model_registry/rest_api/test_model_registry_secure_db.py
+++ b/tests/model_registry/rest_api/test_model_registry_secure_db.py
@@ -1,11 +1,8 @@
 import pytest
 import requests
-from typing import Self
-import copy
+from typing import Self, Any
 from pytest_testconfig import config as py_config
 from tests.model_registry.rest_api.utils import register_model_rest_api, validate_resource_attributes
-from tests.model_registry.rest_api.constants import MODEL_REGISTER_DATA
-from tests.model_registry.utils import generate_random_name
 from tests.model_registry.constants import CA_MOUNT_PATH
 from tests.model_registry.utils import get_mr_service_by_label, get_endpoint_from_mr_service
 from kubernetes.dynamic import DynamicClient
@@ -62,6 +59,7 @@ class TestModelRegistryWithSecureDB:
         model_registry_rest_headers: dict[str, str],
         local_ca_bundle: str,
         deploy_secure_mysql_and_mr: ModelRegistry,
+        model_data_for_test: dict[str, Any],
     ):
         """
         Test that model registration fails with an SSLError when the Model Registry is deployed
@@ -72,15 +70,11 @@ class TestModelRegistryWithSecureDB:
         )
         model_registry_rest_url = get_endpoint_from_mr_service(svc=service, protocol=Protocols.REST)
 
-        model_name = generate_random_name(prefix="model-rest-api")
-        model_data = copy.deepcopy(MODEL_REGISTER_DATA)
-        model_data["register_model_data"]["name"] = model_name
-
         with pytest.raises(requests.exceptions.SSLError) as exc_info:
             register_model_rest_api(
                 model_registry_rest_url=f"https://{model_registry_rest_url}",
                 model_registry_rest_headers=model_registry_rest_headers,
-                data_dict=model_data,
+                data_dict=model_data_for_test,
                 verify=local_ca_bundle,
             )
         assert "SSLError" in str(exc_info.value), (
@@ -109,24 +103,22 @@ class TestModelRegistryWithSecureDB:
         model_registry_rest_headers: dict[str, str],
         local_ca_bundle: str,
         deploy_secure_mysql_and_mr: ModelRegistry,
+        model_data_for_test: dict[str, Any],
     ):
         service = get_mr_service_by_label(
             client=admin_client, namespace_name=model_registry_namespace, mr_instance=deploy_secure_mysql_and_mr
         )
         model_registry_rest_url = get_endpoint_from_mr_service(svc=service, protocol=Protocols.REST)
-        model_name = generate_random_name(prefix="model-rest-api")
-        model_data = copy.deepcopy(MODEL_REGISTER_DATA)
-        model_data["register_model_data"]["name"] = model_name
 
         result = register_model_rest_api(
             model_registry_rest_url=f"https://{model_registry_rest_url}",
             model_registry_rest_headers=model_registry_rest_headers,
-            data_dict=model_data,
+            data_dict=model_data_for_test,
             verify=local_ca_bundle,
         )
         assert result["register_model"].get("id"), "Model registration failed with secure DB connection."
         validate_resource_attributes(
-            expected_params=model_data["register_model_data"],
+            expected_params=model_data_for_test["register_model_data"],
             actual_resource_data=result["register_model"],
             resource_name="register_model",
         )
--- a/tests/model_registry/utils.py
+++ b/tests/model_registry/utils.py
@@ -261,3 +261,72 @@ def generate_random_name(prefix: str = "", length: int = 8) -> str:
 
 def generate_namespace_name(file_path: str) -> str:
     return (file_path.removesuffix(".py").replace("/", "-").replace("_", "-"))[-63:].split("-", 1)[-1]
+
+
+def add_mysql_certs_volumes_to_deployment(
+    spec: dict[str, Any],
+    ca_configmap_name: str,
+) -> list[dict[str, Any]]:
+    """
+    Adds the MySQL certs volumes to the deployment.
+
+    Args:
+        spec: The spec of the deployment
+        ca_configmap_name: The name of the CA configmap
+
+    Returns:
+        The volumes with the MySQL certs volumes added
+    """
+
+    volumes = list(spec["volumes"])
+    volumes.extend([
+        {"name": ca_configmap_name, "configMap": {"name": ca_configmap_name}},
+        {"name": "mysql-server-cert", "secret": {"secretName": "mysql-server-cert"}},  # pragma: allowlist secret
+        {"name": "mysql-server-key", "secret": {"secretName": "mysql-server-key"}},  # pragma: allowlist secret
+    ])
+
+    return volumes
+
+
+def apply_mysql_args_and_volume_mounts(
+    my_sql_container: dict[str, Any],
+    ca_configmap_name: str,
+    ca_mount_path: str,
+) -> dict[str, Any]:
+    """
+    Applies the MySQL args and volume mounts to the MySQL container.
+
+    Args:
+        my_sql_container: The MySQL container
+        ca_configmap_name: The name of the CA configmap
+        ca_mount_path: The mount path of the CA
+
+    Returns:
+        The MySQL container with the MySQL args and volume mounts applied
+    """
+
+    mysql_args = list(my_sql_container.get("args", []))
+    mysql_args.extend([
+        f"--ssl-ca={ca_mount_path}/ca/ca-bundle.crt",
+        f"--ssl-cert={ca_mount_path}/server_cert/tls.crt",
+        f"--ssl-key={ca_mount_path}/server_key/tls.key",
+    ])
+
+    volumes_mounts = list(my_sql_container.get("volumeMounts", []))
+    volumes_mounts.extend([
+        {"name": ca_configmap_name, "mountPath": f"{ca_mount_path}/ca", "readOnly": True},
+        {
+            "name": "mysql-server-cert",
+            "mountPath": f"{ca_mount_path}/server_cert",
+            "readOnly": True,
+        },
+        {
+            "name": "mysql-server-key",
+            "mountPath": f"{ca_mount_path}/server_key",
+            "readOnly": True,
+        },
+    ])
+
+    my_sql_container["args"] = mysql_args
+    my_sql_container["volumeMounts"] = volumes_mounts
+    return my_sql_container
--- a/utilities/certificates_utils.py
+++ b/utilities/certificates_utils.py
@@ -149,7 +149,7 @@ def create_ca_bundle_with_router_cert(
     Creates a CA bundle file by fetching the CA bundle from a ConfigMap and appending the router CA from a Secret.
 
     Args:
-        client: The admin client to get the CA bundle from a ConfigMap and append the router CA from a Secret.
+        client: The client to get the CA bundle from a ConfigMap and append the router CA from a Secret.
         namespace: The namespace of the ConfigMap and Secret.
         ca_bundle_path: The path to the CA bundle file.
         cert_name: The name of the certificate in the ConfigMap.


tests/model_registry/rest_api/conftest.py
```diff
@@ -47,3 +59,47 @@ def updated_model_artifact(
...
+        model_registry_db_service (Service): Service resource for MySQL DB.
+        model_registry_db_deployment (Deployment): Deployment resource for MySQL DB.
+    """
+
+    mysql_template = get_model_registry_deployment_template_dict(
+        secret_name=model_registry_db_secret.name,
+        resource_name=DB_RESOURCES_NAME,
+    )
+
+    ssl_ca_arg = f"--ssl-ca={CA_FILE_PATH}"
```
dbasunag, line 87: You don't need to save it, you can use it directly below. Same comment for ca_volume_mount, ca_volume and patch.

dbasunag (commented):
I started reviewing before noticing it is draft and I feel giving this reviews now, would help with moving the PR fast. Let me know if you disagree, then I will totally stay off till you indicate otherwise.

fege:
> I started reviewing before noticing it is draft and I feel giving this reviews now, would help with moving the PR fast. Let me know if you disagree, then I will totally stay off till you indicate otherwise.

yes that was exactly my intention, to get early reviews to avoid to take a "wrong" path in the solution 👍

tests/model_registry/utils.py
```diff
@@ -315,3 +316,34 @@ def create_model_registry_instance(
...
+    """
+    return {
+        "host": f"{db_service.name}.{db_service.namespace}.svc.cluster.local",
+        "database": db_secret.string_data["database-name"],
+        "passwordSecret": {"key": "database-password", "name": db_resource_name},
+        "port": port,
+        "skipDBCreation": False,
+        "username": db_secret.string_data["database-user"],
+        "ssl_ca": ssl_ca,
+    }
```
lugi0, line 349: I like this a lot. I had added a similar fixture in my pr (`model_registry_mysql_config` in model_registry/conftest.py) - could you make ssl_ca optional and reuse this in that conftest as well?

lugi0 (approved):
A potential improvement that we can also handle in a separate PR, other than that LGTM

tests/model_registry/rest_api/test_model_registry_secure_db.py
```diff
@@ -0,0 +1,96 @@
...
+        result = register_model_rest_api(
+            model_registry_rest_url=model_registry_rest_url,
+            model_registry_rest_headers=model_registry_rest_headers,
+            data_dict=MODEL_REGISTER_DATA,
+            verify=local_ca_bundle,
+        )
+        assert result["register_model"].get("id"), "Model registration failed with secure DB connection."
+        for k, v in MODEL_REGISTER_DATA["register_model_data"].items():
+            assert result["register_model"][k] == v, f"Expected {k}={v}, got {result[k]}"
+        LOGGER.info(f"Model registered successfully with secure DB using {local_ca_bundle}")
```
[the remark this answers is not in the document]
fege: nope, the check it is asserting more then the id, so no need to change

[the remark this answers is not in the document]
fege: result[k] is what we fetch so I want to log out what is coming out to understand the value

tests/model_registry/rest_api/conftest.py
```diff
@@ -47,3 +71,132 @@ def updated_model_artifact(
...
+        router_ca_b64 = router_secret.instance.data.get("tls.crt")
+        if router_ca_b64:
+            router_ca_content = base64.b64decode(router_ca_b64).decode("utf-8")
+            with open(ca_bundle_path, "r", encoding="utf-8") as bundle:
+                bundle_content = bundle.read()
+            if router_ca_content not in bundle_content:
+                with open(ca_bundle_path, "a", encoding="utf-8") as bundle_append:
+                    bundle_append.write("\n" + router_ca_content)
+    except Exception as e:
+        pytest.fail(f"router-ca secret not found in openshift-ingress-operator. Error: {e}")
```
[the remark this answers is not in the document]
fege: the current form is fine

tests/model_registry/conftest.py
```diff
@@ -188,16 +188,22 @@ def model_registry_instance_oauth_proxy(
 
 @pytest.fixture(scope="class")
 def model_registry_mysql_config(
-    model_registry_db_deployment: Deployment, model_registry_db_secret: Secret
+    model_registry_db_deployment: Deployment,
+    model_registry_db_secret: Secret,
+    ssl_ca: str | None = None,
+    port: int = 3306,
```
dbasunag, line 194: I don't believe this is how a fixture works. It can request another fixture, but one does not pass args to it outside of request.param https://docs.pytest.org/en/stable/how-to/fixtures.html

tests/model_registry/conftest.py
```diff
@@ -188,16 +188,22 @@ def model_registry_instance_oauth_proxy(
...
 ) -> dict[str, Any]:
-    return {
+    config = {
         "host": f"{model_registry_db_deployment.name}.{model_registry_db_deployment.namespace}.svc.cluster.local",
         "database": model_registry_db_secret.string_data["database-name"],
         "passwordSecret": {"key": "database-password", "name": model_registry_db_deployment.name},
-        "port": 3306,
+        "port": port,
```
dbasunag, line 200: if this needs to be read from somewhere, then please use request.param

tests/model_registry/conftest.py
```diff
@@ -188,16 +188,22 @@ def model_registry_instance_oauth_proxy(
...
         "skipDBCreation": False,
         "username": model_registry_db_secret.string_data["database-user"],
     }
+    if ssl_ca is not None:
+        config["ssl_ca"] = ssl_ca
+    return config
```
dbasunag, line 206: Ditto

tests/model_registry/rest_api/conftest.py
```diff
@@ -47,3 +71,132 @@ def updated_model_artifact(
...
+@pytest.fixture
+def local_ca_bundle(request: pytest.FixtureRequest, admin_client: DynamicClient) -> Generator[str, Any, Any]:
+    """
+    Creates a local CA bundle file by fetching the CA bundle from a ConfigMap and appending the router CA from a Secret.
+    """
+    namespace = getattr(request, "param", {}).get("namespace", "test-model-registry-namespace")
+    ca_bundle_path = getattr(request, "param", {}).get("ca_bundle_path", "ca-bundle.crt")
+    cert_name = getattr(request, "param", {}).get("cert_name", "ca-bundle.crt")
+
+    cm = ConfigMap(client=admin_client, name="odh-trusted-ca-bundle", namespace=namespace)
```
dbasunag, line 180: Please use ensure_exists=True. Is it named the same way in the on odh cluster?

tests/model_registry/rest_api/conftest.py
```diff
@@ -47,3 +71,132 @@ def updated_model_artifact(
...
+            mr.wait_for_condition(condition="Available", status="True")
+            yield mr
+
+
+@pytest.fixture
+def local_ca_bundle(request: pytest.FixtureRequest, admin_client: DynamicClient) -> Generator[str, Any, Any]:
+    """
+    Creates a local CA bundle file by fetching the CA bundle from a ConfigMap and appending the router CA from a Secret.
+    """
+    namespace = getattr(request, "param", {}).get("namespace", "test-model-registry-namespace")
```
dbasunag, line 176: Why not use the  py_config["model_registry_namespace"] as the default?

tests/model_registry/rest_api/conftest.py
```diff
@@ -47,3 +71,132 @@ def updated_model_artifact(
...
+                with open(ca_bundle_path, "a", encoding="utf-8") as bundle_append:
+                    bundle_append.write("\n" + router_ca_content)
+    except Exception as e:
+        pytest.fail(f"router-ca secret not found in openshift-ingress-operator. Error: {e}")
+    yield ca_bundle_path
+
+    try:
+        os.remove(ca_bundle_path)
+    except FileNotFoundError:
+        pass
```
dbasunag, line 202: We should not hit this path, right? If we have problem removing something we created, that should not be hidden.

tests/model_registry/rest_api/test_model_registry_secure_db.py
```diff
@@ -0,0 +1,109 @@
...
+
+@pytest.mark.parametrize(
+    "updated_dsc_component_state_scope_class",
+    [
+        pytest.param(
+            {
+                "component_patch": {
+                    DscComponents.MODELREGISTRY: {
+                        "managementState": DscComponents.ManagementState.MANAGED,
+                        "registriesNamespace": "test-model-registry-namespace",
```
dbasunag, line 22: Please use constant, since it is being used in multiple places.

tests/model_registry/rest_api/test_model_registry_secure_db.py
```diff
@@ -0,0 +1,109 @@
...
+)
+@pytest.mark.usefixtures("updated_dsc_component_state_scope_class")
+class TestModelRegistryWithSecureDB:
+    """
+    Test suite for validating Model Registry functionality with a secure MySQL database connection (SSL/TLS).
+    Includes tests for both invalid and valid CA certificate scenarios.
+    """
+
+    @pytest.fixture
+    def invalid_ca_mysql_config(patch_invalid_ca):
```
dbasunag, line 38: Please move this to conftest.py

tests/model_registry/rest_api/conftest.py
fege: you are right, I am going to change it

tests/model_registry/rest_api/conftest.py
fege: this is what I found in the doc https://opendatahub.io/docs/working-with-model-registries/

tests/model_registry/conftest.py
```diff
@@ -190,15 +190,25 @@ def model_registry_instance_oauth_proxy(
...
+    request: FixtureRequest,
+    model_registry_db_deployment: Deployment,
+    model_registry_db_secret: Secret,
 ) -> dict[str, Any]:
+    """
+    Fixture to build the MySQL config dictionary for Model Registry.
+    Expects request.param to be a dict. If 'ssl_ca' is not present, it defaults to None.
+    """
+    param = request.param if hasattr(request, "param") else {}
+    ssl_ca = param.get("ssl_ca", None)
```
dbasunag, line 202: get() defaults to None, you only need to set `non-None` defaults.

tests/model_registry/conftest.py
```diff
@@ -190,15 +190,25 @@ def model_registry_instance_oauth_proxy(
...
+    port = param.get("port", 3306)
```
dbasunag, line 203: you don't need to save these. You can directly do the param.get() to the line below, where they are used.

tests/model_registry/rest_api/conftest.py
```diff
@@ -47,3 +71,131 @@ def updated_model_artifact(
...
+@pytest.fixture(scope="class")
+def patch_invalid_ca(
+    admin_client: DynamicClient,
+    model_registry_namespace: str,
+    request: pytest.FixtureRequest,
+) -> Generator[str, Any, Any]:
+    """
+    Patches the odh-trusted-ca-bundle ConfigMap with an invalid CA certificate.
+    """
+    ca_configmap_name = request.param.get("ca_configmap_name", "odh-trusted-ca-bundle")
```
dbasunag, line 85: Would this be same for ODH and RHOAI?

tests/model_registry/rest_api/conftest.py
```diff
@@ -47,3 +71,131 @@ def updated_model_artifact(
...
+    ca_file_name = request.param.get("ca_file_name", "invalid-ca.crt")
+    ca_file_path = f"{CA_MOUNT_PATH}/{ca_file_name}"
+    ca_data = {ca_file_name: "-----BEGIN CERTIFICATE-----\nINVALIDCERTIFICATE\n-----END CERTIFICATE-----"}
+    ca_configmap = ConfigMap(
+        client=admin_client,
+        name=ca_configmap_name,
+        namespace=model_registry_namespace,
```
dbasunag, line 92: Please use ensure_exists=True.

tests/model_registry/rest_api/conftest.py
```diff
@@ -47,3 +71,131 @@ def updated_model_artifact(
...
+        resource_name=DB_RESOURCES_NAME,
+    )
+
+    mysql_template["spec"]["containers"][0]["args"].append(f"--ssl-ca={CA_FILE_PATH}")
+    mysql_template["spec"]["containers"][0]["volumeMounts"].append({
+        "mountPath": CA_MOUNT_PATH,
+        "name": CA_CONFIGMAP_NAME,
+        "readOnly": True,
+    })
+    mysql_template["spec"]["volumes"].append({"name": CA_CONFIGMAP_NAME, "configMap": {"name": CA_CONFIGMAP_NAME}})
```
dbasunag, line 151: This should be a separate fixture.

tests/model_registry/rest_api/conftest.py
```diff
@@ -47,3 +71,131 @@ def updated_model_artifact(
...
+            rest={},
+            istio=ISTIO_CONFIG_DICT,
+            mysql=model_registry_mysql_config,
+            wait_for_resource=True,
+        ) as mr:
+            mr.wait_for_condition(condition="Available", status="True")
+            yield mr
+
+
+@pytest.fixture
```
dbasunag, line 170: @pytest.fixture().

tests/model_registry/rest_api/conftest.py
```diff
@@ -47,3 +71,131 @@ def updated_model_artifact(
...
+        client=admin_client, name="router-ca", namespace="openshift-ingress-operator", ensure_exists=True
+    )
+    router_ca_b64 = router_secret.instance.data.get("tls.crt")
+    if router_ca_b64:
+        router_ca_content = base64.b64decode(router_ca_b64).decode("utf-8")
+        with open(ca_bundle_path, "r", encoding="utf-8") as bundle:
+            bundle_content = bundle.read()
+        if router_ca_content not in bundle_content:
+            with open(ca_bundle_path, "a", encoding="utf-8") as bundle_append:
+                bundle_append.write("\n" + router_ca_content)
```
dbasunag, line 194: Should they be separate fixtures, utilizing a util?

tests/model_registry/rest_api/conftest.py
```diff
@@ -47,3 +71,131 @@ def updated_model_artifact(
...
+
+    yield ca_bundle_path
+
+    try:
+        os.remove(ca_bundle_path)
+    except FileNotFoundError:
+        pytest.fail(f"CA bundle file not removed: {ca_bundle_path}")
```
dbasunag, line 201: This is teardown, we should let it raise exception instead.

tests/model_registry/rest_api/test_model_registry_secure_db.py
```diff
@@ -0,0 +1,105 @@
...
+        )
+
+    # Implements RHOAIENG-26150
+    @pytest.mark.usefixtures("deploy_secure_mysql_and_mr")
+    @pytest.mark.parametrize(
+        "local_ca_bundle",
+        [{"cert_name": "ca-bundle.crt"}, {"cert_name": "odh-ca-bundle.crt"}],
+        indirect=True,
+    )
+    @pytest.mark.smoke
```
dbasunag, line 83: Is this RHOAI only or this test is applicable for ODH? Does it require any adjustments to run against ODH cluster?

tests/model_registry/rest_api/conftest.py
```diff
@@ -47,3 +74,128 @@ def updated_model_artifact(
...
+    cert_name = getattr(request, "param", {}).get("cert_name", "ca-bundle.crt")
+    create_ca_bundle_file(
+        admin_client=admin_client, namespace=namespace, ca_bundle_path=ca_bundle_path, cert_name=cert_name
+    )
+    yield ca_bundle_path
+
+    try:
+        os.remove(ca_bundle_path)
+    except FileNotFoundError as e:
+        raise FileNotFoundError(f"CA bundle file not removed: {ca_bundle_path}, Error: {e}")
```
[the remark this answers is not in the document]
fege: this is wanted

conftest.py
```diff
@@ -233,6 +236,32 @@ def pytest_sessionstart(session: Session) -> None:
         path=must_gather_dict["must_gather_base_directory"],
         ignore_errors=True,
     )
+    updated_global_config(admin_client=get_client())
```
dbasunag, line 239: this should be skipped `--collect-only` or `--setup-plan` are passed. Otherwise you would see failures in tox.

conftest.py
```diff
@@ -237,8 +237,8 @@ def pytest_sessionstart(session: Session) -> None:
         ignore_errors=True,
     )
     config = session.config
-    if config.getoption("collectonly") or config.getoption("setupplan", default=False):
-        return  #
+    if config.getoption("--collect-only") or config.getoption("--setup-plan", default=False):
+        return
```
dbasunag, line 241: Can you please log something? I don't believe you need to use defaults here

tests/model_registry/rest_api/conftest.py
```diff
@@ -63,3 +89,128 @@ def updated_model_registry_resource(
...
+    namespace = getattr(request, "param", {}).get("namespace", py_config["model_registry_namespace"])
+    ca_bundle_path = getattr(request, "param", {}).get("ca_bundle_path", "ca-bundle.crt")
+    cert_name = getattr(request, "param", {}).get("cert_name", "ca-bundle.crt")
+    create_ca_bundle_file(
+        admin_client=admin_client, namespace=namespace, ca_bundle_path=ca_bundle_path, cert_name=cert_name
+    )
+    yield ca_bundle_path
+
+    try:
+        os.remove(ca_bundle_path)
```
dbasunag, line 214: You can remove the try/except. Yiu are simply catching and raising the same exception

conftest.py
```diff
@@ -277,7 +310,7 @@ def pytest_runtest_setup(item: Item) -> None:
         item.fixturenames.insert(0, "enabled_modelmesh_in_dsc")
 
     # The above fixtures require the global config to be updated before being called
-    item.fixturenames.insert(0, "updated_global_config")
+    updated_global_config(admin_client=get_client())
```
dbasunag, line 313: you need to remove this.

tests/model_registry/conftest.py
```diff
@@ -177,15 +177,23 @@ def model_registry_instance(
...
+    param = request.param if hasattr(request, "param") else {}
     return {
         "host": f"{model_registry_db_deployment.name}.{model_registry_db_deployment.namespace}.svc.cluster.local",
         "database": model_registry_db_secret.string_data["database-name"],
         "passwordSecret": {"key": "database-password", "name": model_registry_db_deployment.name},
-        "port": 3306,
+        "port": param.get("port", 3306),
         "skipDBCreation": False,
         "username": model_registry_db_secret.string_data["database-user"],
+        "ssl_ca": param.get("ssl_ca"),
```
lugi0, line 196: won't this fail if you don't pass in ssl_ca?

conftest.py
lugi0: yep, this seems like a leftover

fege: I put the if to guard to no have tox to fail, but I kept the original flow that was calling the fixture, now the fixture is a normal function. But if you both think that is not needed anymore for `pytest_runtest_setup` I can remove it

tests/model_registry/conftest.py
lugi0: and what happens to the mysql config if it has `ssl_ca` defined None/empty? I'm afraid that's what would break in the MR instance down the line

fege: you are right if it is not passed the attribute should not be present at all in the config

lugi0 (approved):

tests/model_registry/conftest.py
```diff
@@ -177,16 +180,40 @@ def model_registry_instance(
...
-        "port": 3306,
+        "port": param.get("port", 3306),
         "skipDBCreation": False,
         "username": model_registry_db_secret.string_data["database-user"],
     }
+    if param.get("sslRootCertificateConfigMap"):
+        ssl_config = param.get("sslRootCertificateConfigMap")
+        if ssl_config:
+            config["sslRootCertificateConfigMap"] = {
+                "name": ssl_config.get("name"),
```
dbasunag, line 212: you can set the param in your test such that you can directly set `config["sslRootCertificateConfigMap"]=param.get("sslRootCertificateConfigMap")`. This way it would be None, if the param does not exist.

tests/model_registry/rest_api/conftest.py
```diff
@@ -63,3 +92,250 @@ def updated_model_registry_resource(
...
+        yield ca_file_path
+
+
+@pytest.fixture(scope="class")
+def mysql_template_with_ca(model_registry_db_secret: Secret) -> dict[str, Any]:
+    """
+    Patches the MySQL template with the CA file path and volume mount.
+    Args:
+        model_registry_db_secret: The secret for the model registry's MySQL database
+    Returns:
```
dbasunag, line 133: if you want to use doc headers, then let's put a new line before Args and Returns.

tests/model_registry/rest_api/conftest.py
```diff
@@ -63,3 +92,250 @@ def updated_model_registry_resource(
...
+        model_registry_mysql_config: The MySQL config dictionary
+        mysql_template_with_ca: The MySQL template with the CA file path and volume mount
+    """
+    with ModelRegistry(
+        name=SECURE_MR_NAME,
+        namespace=model_registry_namespace,
+        label=MODEL_REGISTRY_STANDARD_LABELS,
+        grpc={},
+        rest={},
+        istio=ISTIO_CONFIG_DICT,
```
dbasunag, line 174: We don't need this anymore and should use oauth.

tests/model_registry/rest_api/conftest.py
```diff
@@ -63,3 +92,250 @@ def updated_model_registry_resource(
...
+        ca_bundle_path=ca_bundle_path,
+        cert_name=ca_bundle_path,
+    )
+    yield ca_bundle_path
+
+    os.remove(ca_bundle_path)
+
+
+@pytest.fixture(scope="class")
+def test_ca_configmap(
```
dbasunag, line 205: We should not name anything starting with `test`. May be rename it to ca_configmap_for_test or something similar.

tests/model_registry/rest_api/conftest.py
```diff
@@ -63,3 +92,250 @@ def updated_model_registry_resource(
...
+                namespace=model_registry_namespace,
+                data_dict=data,
+                wait_for_resource=True,
+            )
+            secret.create()
+            return secret
+
+        ca_secret = create_secret(name="mysql-ca", file_path=paths["ca_crt"], key_name="ca.crt")
+        server_cert_secret = create_secret(name="mysql-server-cert", file_path=paths["server_crt"], key_name="tls.crt")
+        server_key_secret = create_secret(name="mysql-server-key", file_path=paths["server_key"], key_name="tls.key")
```
dbasunag, line 332: This fixture is doing too many things at a time. How about having smaller fixtures. They are typically supposed to do one thing, that way, triage is easy.

tests/model_registry/rest_api/conftest.py
```diff
@@ -63,3 +92,250 @@ def updated_model_registry_resource(
...
+        admin_client: The Kubernetes dynamic client used to create secrets.
+        model_registry_namespace: The namespace in which to create the secrets.
+
+    Yields:
+        dict: A dictionary containing file paths and secret objects for the CA, server certificate, and server key.
+    """
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        paths = generate_ca_and_server_cert(tmp_dir=tmp_dir)
+
+        def create_secret(name: str, file_path: str, key_name: str) -> Secret:
```
dbasunag, line 315: Does not have to be internal function. Any specific reason for this?

tests/model_registry/rest_api/test_model_registry_secure_db.py
```diff
@@ -0,0 +1,131 @@
...
+from utilities.constants import DscComponents
+
+from simple_logger.logger import get_logger
+
+
+LOGGER = get_logger(name=__name__)
+
+
+@pytest.mark.parametrize(
+    "updated_dsc_component_state_scope_class",
```
dbasunag, line 21: For main we must use oauth.

tests/model_registry/rest_api/test_model_registry_secure_db.py
```diff
@@ -0,0 +1,131 @@
...
+        model_name = generate_random_name(prefix="model-rest-api")
+        model_data = copy.deepcopy(MODEL_REGISTER_DATA)
+        model_data["register_model_data"]["name"] = model_name
+
+        endpoint = get_model_registry_endpoint(
+            admin_client=admin_client,
+            model_registry_namespace=model_registry_namespace,
+            model_registry_name=SECURE_MR_NAME,
+        )
+        model_registry_rest_url = endpoint.split(":")[0]
```
dbasunag, line 77: Should be a fixture.

tests/model_registry/rest_api/test_model_registry_secure_db.py
```diff
@@ -0,0 +1,131 @@
...
+        local_ca_bundle: str,
+    ):
+        model_name = generate_random_name(prefix="model-rest-api")
+        model_data = copy.deepcopy(MODEL_REGISTER_DATA)
+        model_data["register_model_data"]["name"] = model_name
+        endpoint = get_model_registry_endpoint(
+            admin_client=admin_client,
+            model_registry_namespace=model_registry_namespace,
+            model_registry_name=SECURE_MR_NAME,
+        )
```
dbasunag, line 117: Should be a fixture.

tests/model_registry/rest_api/test_model_registry_secure_db.py
```diff
@@ -0,0 +1,131 @@
...
+        model_registry_rest_url = endpoint.split(":")[0]
+        result = register_model_rest_api(
+            model_registry_rest_url=f"https://{model_registry_rest_url}",
+            model_registry_rest_headers=model_registry_rest_headers,
+            data_dict=model_data,
+            verify=local_ca_bundle,
+        )
+        assert result["register_model"].get("id"), "Model registration failed with secure DB connection."
```
dbasunag, line 125: Should be fixture too.

tests/model_registry/rest_api/utils.py
```diff
@@ -91,6 +105,38 @@ def register_model_rest_api(
     return {"register_model": register_model, "model_version": model_version, "model_artifact": model_artifact}
 
 
+def create_ca_bundle_file(
```
dbasunag, line 108: Can you enhance https://github.com/opendatahub-io/opendatahub-tests/blob/main/utilities/certificates_utils.py#L21 or rename and move it to the same file?

tests/model_registry/rest_api/utils.py
```diff
@@ -113,3 +159,87 @@ def validate_resource_attributes(
...
+    ca_key = os.path.join(tmp_dir, "ca.key")
+    ca_crt = os.path.join(tmp_dir, "ca.crt")
+    server_key = os.path.join(tmp_dir, "server-key.pem")
+    server_csr = os.path.join(tmp_dir, "server.csr")
+    server_crt = os.path.join(tmp_dir, "server-cert.pem")
+
+    LOGGER.info(f"Generating CA and server cert in {tmp_dir} for DB hostname {db_service_hostname}")
+
+    # --- 1. Create CA private key and certificate ---
+    subprocess.run(args=["openssl", "genrsa", "-out", ca_key, "2048"], check=True)
```
dbasunag, line 191: Can you use run_command() instead?

tests/model_registry/rest_api/utils.py
```diff
@@ -113,3 +159,87 @@ def validate_resource_attributes(
...
+        args=[
+            "openssl",
+            "x509",
+            "-req",
+            "-in",
+            server_csr,
+            "-CA",
+            ca_crt,
+            "-CAkey",
+            ca_key,
```
dbasunag, line 230: Please add small and separate utils.

tests/model_registry/rest_api/utils.py
```diff
@@ -31,8 +37,10 @@ def execute_model_registry_patch_command(
         raise
 
 
-def execute_model_registry_post_command(url: str, headers: dict[str, str], data_json: dict[str, Any]) -> dict[Any, Any]:
-    resp = requests.post(url=url, json=data_json, headers=headers, verify=False, timeout=60)
+def execute_model_registry_post_command(
+    url: str, headers: dict[str, str], data_json: dict[str, Any], verify: bool | str = False
+) -> dict[Any, Any]:
+    resp = requests.post(url=url, json=data_json, headers=headers, verify=verify, timeout=60)
     LOGGER.info(f"url: {url}, status code: {resp.status_code}, rep: {resp.text}")
```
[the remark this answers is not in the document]
fege: I don't want to modify the old default

tests/model_registry/rest_api/conftest.py
fege: you are right, thanks, I was still gettin MR `from ocp_resources.model_registry import ModelRegistry`

tests/model_registry/rest_api/conftest.py
```diff
@@ -63,3 +92,250 @@ def updated_model_registry_resource(
...
+    original_deployment = deployment.instance.to_dict()
+    spec = original_deployment["spec"]["template"]["spec"]
+    my_sql_container = next(container for container in spec["containers"] if container["name"] == "mysql")
+    assert my_sql_container is not None, "Mysql container not found"
+    mysql_args = list(my_sql_container.get("args", []))
+    mysql_args.extend([
+        f"--ssl-ca={CA_MOUNT_PATH}/ca/ca-bundle.crt",
+        f"--ssl-cert={CA_MOUNT_PATH}/server_cert/tls.crt",
+        f"--ssl-key={CA_MOUNT_PATH}/server_key/tls.key",
+    ])
```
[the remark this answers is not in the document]
fege: this is wanted

tests/model_registry/rest_api/test_model_registry_secure_db.py
fege: if it is by default and not secure_db related it should not be addressed in this change to avoid scope creep

tests/model_registry/rest_api/conftest.py
fege: splitted

utilities/certificates_utils.py
```diff
@@ -101,3 +102,72 @@ def get_ca_bundle(client: DynamicClient, deployment_mode: str) -> str:
...
+
+    router_secret = Secret(client=client, name="router-ca", namespace="openshift-ingress-operator", ensure_exists=True)
+    router_ca_b64 = router_secret.instance.data.get("tls.crt")
+    if router_ca_b64:
+        router_ca_content = base64.b64decode(router_ca_b64).decode("utf-8")
+        with open(ca_bundle_path, "r", encoding="utf-8") as bundle:
+            bundle_content = bundle.read()
+        if router_ca_content not in bundle_content:
+            with open(ca_bundle_path, "a", encoding="utf-8") as bundle_append:
+                bundle_append.write("\n" + router_ca_content)
```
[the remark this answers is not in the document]
fege: this is ok

tests/model_registry/rest_api/conftest.py
```diff
@@ -63,3 +90,272 @@ def updated_model_registry_resource(
...
+    ca_bundle_path = getattr(request, "param", {}).get("cert_name", "ca-bundle.crt")
+    create_ca_bundle_with_router_cert(
+        client=admin_client,
+        namespace=py_config["model_registry_namespace"],
+        ca_bundle_path=ca_bundle_path,
+        cert_name=ca_bundle_path,
+    )
+    yield ca_bundle_path
+
+    os.remove(ca_bundle_path)
```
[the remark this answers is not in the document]
fege: for the moment is ok like this

lugi0 (dismissed):

tests/model_registry/rest_api/conftest.py
```diff
@@ -63,3 +90,272 @@ def updated_model_registry_resource(
...
+    and mount the server cert/key for SSL.
+    """
+    CA_CONFIGMAP_NAME = request.param.get("ca_configmap_name", "mysql-ca-configmap")
+    CA_MOUNT_PATH = request.param.get("ca_mount_path", "/etc/mysql/ssl")
+    deployment = Deployment(
+        client=admin_client,
+        name=model_registry_db_deployment.name,
+        namespace=model_registry_namespace,
+    )
+    deployment.wait_for_condition(condition="Available", status="True")
```
dbasunag, line 258: you already are requesting model_registry_db_deployment in this fixture, you should be able to use it here.

tests/model_registry/rest_api/conftest.py
```diff
@@ -63,3 +90,272 @@ def updated_model_registry_resource(
...
+            "mountPath": f"{CA_MOUNT_PATH}/server_key",
+            "readOnly": True,
+        },
+    ])
+
+    my_sql_container["args"] = mysql_args
+    my_sql_container["volumeMounts"] = volumes_mounts
+    volumes = list(spec["volumes"])
+    volumes.extend([
+        {"name": CA_CONFIGMAP_NAME, "configMap": {"name": CA_CONFIGMAP_NAME}},
```
dbasunag, line 289: May be move to a separate fixture, this fixture is doing a lot of things - setting values for different fields, and then updating the deployment. Also, why don't we create the deployment with right values? Instead of creating and patching, you can simply have a separate fixture for secure db.

tests/model_registry/rest_api/test_model_registry_secure_db.py
dbasunag: With Luca's change merged, this comment is no longer valid. Resolving it.

tests/model_registry/rest_api/test_model_registry_secure_db.py
```diff
@@ -0,0 +1,133 @@
...
+        model_registry_namespace: str,
+        model_registry_rest_headers: dict[str, str],
+        local_ca_bundle: str,
+        deploy_secure_mysql_and_mr: ModelRegistry,
+    ):
+        """
+        Test that model registration fails with an SSLError when the Model Registry is deployed
+        with an invalid CA certificate.
+        """
+        service = get_mr_service_by_label(
```
dbasunag, line 70: Why not use model_registry_instance_service or it's equivalent fixture here?

tests/model_registry/rest_api/test_model_registry_secure_db.py
```diff
@@ -0,0 +1,133 @@
...
+        local_ca_bundle: str,
+        deploy_secure_mysql_and_mr: ModelRegistry,
+    ):
+        service = get_mr_service_by_label(
+            client=admin_client, namespace_name=model_registry_namespace, mr_instance=deploy_secure_mysql_and_mr
+        )
+        model_registry_rest_url = get_endpoint_from_mr_service(svc=service, protocol=Protocols.REST)
+        model_name = generate_random_name(prefix="model-rest-api")
+        model_data = copy.deepcopy(MODEL_REGISTER_DATA)
+        model_data["register_model_data"]["name"] = model_name
```
dbasunag, line 119: Please make this a fixture, it is being used in both the tests.

dbasunag (commented):
Looks good @fege . I added a few comments though.

tests/model_registry/rest_api/test_model_registry_secure_db.py
fege: because it is getting the value of `MR_INSTANCE_NAME` . it should be to parametrize, but I prefer to not do in this PR that is already big

dbasunag: Ok, let's handle it in a follow up PR.

tests/model_registry/utils.py
```diff
@@ -261,3 +261,72 @@ def generate_random_name(prefix: str = "", length: int = 8) -> str:
...
+        {
+            "name": "mysql-server-key",
+            "mountPath": f"{ca_mount_path}/server_key",
+            "readOnly": True,
+        },
+    ])
+
+    my_sql_container["args"] = mysql_args
+    my_sql_container["volumeMounts"] = volumes_mounts
+    return my_sql_container
```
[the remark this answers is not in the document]
fege: no

tests/model_registry/rest_api/conftest.py
lugi0: Agreed @fege, as long as this works for the time being IMHO it's ok to merge and we can improve it in a separate PR

PR #321, opened 2025-05-29
Add test for oauth deployment of MR

lugi0:
## Merge criteria:

## Summary by CodeRabbit

- **New Features**
  - Added support for testing model registry creation and registration with OAuth proxy enabled.
  - Introduced a new test for registering and validating models using OAuth authentication.

- **Tests**
  - Improved test setup by modularizing model registry instance creation and adding reusable fixtures.
  - Added constants for OAuth proxy configuration to streamline test configuration.
  - Updated import paths for better consistency and clarity in test files.
  - Enhanced flexibility in test client fixture to support dynamic service selection.

- **Chores**
  - Updated a dependency version requirement for improved compatibility.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -35,7 +35,7 @@ repos:
       - id: detect-secrets
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.11.11
+    rev: v0.11.12
     hooks:
       - id: ruff
       - id: ruff-format
@@ -49,12 +49,12 @@ repos:
   #     - id: renovate-config-validator
 
   - repo: https://github.com/gitleaks/gitleaks
-    rev: v8.26.0
+    rev: v8.27.0
     hooks:
       - id: gitleaks
 
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v1.15.0
+    rev: v1.16.0
     hooks:
       - id: mypy
         additional_dependencies: ["types-PyYAML", "types-requests"]
--- a/utilities/plugins/openai_plugin.py
+++ b/utilities/plugins/openai_plugin.py
@@ -99,7 +99,7 @@ class OpenAIClient:
             for line in response.iter_lines():
                 _, found, data = line.partition(b"data: ")
                 if found and data != b"[DONE]":
-                    message = json.loads(data)  # type: ignore
+                    message = json.loads(data)
                     token = self._parse_streaming_response(endpoint, message)
                     tokens.append(token)
         except (requests.exceptions.RequestException, json.JSONDecodeError):


lugi0:
@coderabbit review

tests/model_registry/conftest.py
```diff
@@ -143,34 +144,130 @@ def model_registry_db_deployment(
         yield mr_db_deployment
 
 
+def create_model_registry_instance(
+    admin_client: DynamicClient,
```
dbasunag, line 148: Let's move functions to utils.py please :)

tests/model_registry/conftest.py
```diff
@@ -143,34 +144,130 @@ def model_registry_db_deployment(
...
-    with ModelRegistry(
-        name=MR_INSTANCE_NAME,
+    """Creates a model registry instance with service mesh configuration."""
+    labels = {
+        Annotations.KubernetesIo.NAME: MR_INSTANCE_NAME,
+        Annotations.KubernetesIo.INSTANCE: MR_INSTANCE_NAME,
+        Annotations.KubernetesIo.PART_OF: MR_OPERATOR_NAME,
+        Annotations.KubernetesIo.CREATED_BY: MR_OPERATOR_NAME,
+    }
+
```
dbasunag, line 213: Can this be a separate fixture? We don't expect it to change. So it can be session/package scoped.

tests/model_registry/conftest.py
```diff
@@ -143,34 +144,130 @@ def model_registry_db_deployment(
...
+    mysql_config = {
+        "host": f"{model_registry_db_deployment.name}.{model_registry_db_deployment.namespace}.svc.cluster.local",
+        "database": model_registry_db_secret.string_data["database-name"],
+        "passwordSecret": {"key": "database-password", "name": model_registry_db_deployment.name},
+        "port": 3306,
+        "skipDBCreation": False,
+        "username": model_registry_db_secret.string_data["database-user"],
+    }
```
dbasunag, line 221: Ditto, can this be a class scoped fixture?

tests/model_registry/python_client/test_model_registry_creation_oauth.py
```diff
@@ -0,0 +1,93 @@
...
+                    },
+                },
+            },
+        ),
+    ],
+    indirect=True,
+)
+@pytest.mark.usefixtures("updated_dsc_component_state_scope_class")
+class TestModelRegistryCreationOAuth:
+    """
```
dbasunag, line 33: It would be great if you can include the jiras here.

tests/model_registry/python_client/test_model_registry_creation_oauth.py
```diff
@@ -0,0 +1,93 @@
...
+    Tests the creation of a model registry with OAuth proxy configuration.
+    """
+
+    @pytest.mark.smoke
```
dbasunag, line 37: TestDevops still uses `latest` tag for all versions. So you may want to hold off adding it to smoke for the time being. May be we should have a jira to track and enable it once TestDevops starts using right images for right versions.

tests/model_registry/conftest.py
```diff
@@ -143,34 +144,130 @@ def model_registry_db_deployment(
         yield mr_db_deployment
 
 
+def create_model_registry_instance(
```
fege, line 147: here `@dataclass` could be applied, but it is just a niptick

tests/model_registry/conftest.py
lugi0: I think rather than a fixture this would make more sense as a constant

tests/model_registry/python_client/test_model_registry_creation_oauth.py
lugi0: Mh, we do want this to be run though. Perhaps we can add a dynamic check here on the installed version and execute only if >=2.21?

dbasunag: Let's do that in a separate PR. The reason being, your PR is ready to be merged. I don't want to have to wait for it, just because of this marker.

tests/model_registry/python_client/test_model_registry_creation_oauth.py
```diff
@@ -0,0 +1,77 @@
...
+    def test_registering_model_with_oauth(
+        self: Self,
+        admin_client,
+        model_registry_client: ModelRegistryClient,
+    ):
+        if py_config["distribution"] == "downstream" and get_product_version(admin_client=admin_client) < MINVER:
+            pytest.skip("Skipping test for RHOAI < 2.21")
+
```
dbasunag, line 48: Is it not valid for odh? I still feel this should be moved out of this PR. We can mark it as smoke, after we get a chance to discuss and push for use of right tags for jenkins jobs.

dbasunag (commented):
Other than the smoke marker and skip, this PR is good from my side.

tests/model_registry/python_client/test_model_registry_creation_oauth.py
lugi0: It is valid but we don't know from which ODH release, in ODH right now it won't ever be skipped. AFAIK we don't have automated runs that include opendatahub-tests against ODH so it shouldn't be a problem

fege (approved):
/lgtm

PR #332, opened 2025-06-03
adding e2e tests for openshift route timeout

brettmthompson:
## Description

[RHOAIENG-24705](https://issues.redhat.com/browse/RHOAIENG-24705)
Adding e2e tests to cover openshift route timeout for inference services. Now if the `haproxy.router.openshift.io/timeout` is added to an ISVC, this annotation will be propagated to the openshift route created.

These tests create raw and serverless inference services with the default timeout applied (30 seconds) and validate that inference is performed successfully. Then the ISVC is updated with the `haproxy.router.openshift.io/timeout` annotation and  a value equivalent to 1 microsecond. To validate this annotation is properly applied to the route, inference is tested with the expectation of failing due to the low timeout.

**NOTE**
- From manual testing we must wait after updating the ISVC with the `haproxy.router.openshift.io/timeout` for the annotation to be propagated to the route and for the route to enforce that timeout. I have created a new method `wait_for_route_timeout` and instituted a counter that will loop through the `TimeoutSampler` until the annotation is detected on the route twice. This allows for the `TimeoutSampler`'s `sleep` statement to be triggered, providing enough time for the timeout to be applied to the route. 
- For the serverless test, it appears that no error response is returned by the service when timeout is exceeded. Therefore we cannot parse the inference response when timeout is exceeded. The inference fails while running the command, which should throw a `ValueError`. However, the method to run the command is wrapped in the `@retry` wrapper, which means the top level exception thrown when inference fails will be a `TimeoutExpiredError`

## How Has This Been Tested?

Run the new tests in an env where odh-model-controller code is used from https://github.com/brettmthompson/odh-model-controller/tree/feature/add-timeout-annotation-to-openshift-route. An ODH installation for version 2.30.0 from the `fast` channel contains this code. 

## Merge criteria:


tests/model_serving/model_server/routes/test_raw_deployment.py
```diff
@@ -98,6 +104,69 @@ def test_disabled_rest_raw_deployment_exposed_route(self, patched_s3_caikit_kser
...
+        """Test HTTP inference using exposed (external) route fails when timeout is set too low"""
+        ResourceEditor(
+            patches={
+                s3_models_inference_service: {
+                    "metadata": {
+                        "annotations": {Annotations.HaproxyRouterOpenshiftIo.TIMEOUT: "1us"},
+                    }
+                }
+            }
+        ).update()
```
dbasunag, line 153: Should be a fixture I feel.

tests/model_serving/model_server/routes/test_raw_deployment.py
```diff
@@ -98,6 +104,69 @@ def test_disabled_rest_raw_deployment_exposed_route(self, patched_s3_caikit_kser
...
+
+        # Wait for route to be updated with the annotation and timeout to be applied
+        time.sleep(10)  # noqa: FCN001
```
dbasunag, line 156: Instead of sleep, please check the right kind to see if the update was reflected.

tests/model_serving/model_server/routes/test_raw_deployment.py
```diff
@@ -151,3 +220,67 @@ def test_grpc_raw_deployment_exposed_route(self, patched_s3_caikit_kserve_isvc_v
...
```
dbasunag, line 275: Same comments here. Also, seems like test_grpc_raw_deployment_exposed_route_with_timeout() and test_rest_raw_deployment_exposed_route_with_timeout() could be parameterized.

tests/model_serving/model_server/routes/test_serverless.py
```diff
@@ -94,3 +98,63 @@ def test_rest_serverless_exposed_label_route(self, patched_s3_caikit_kserve_isvc
...
+                s3_models_inference_service: {
+                    "metadata": {
+                        "annotations": {Annotations.HaproxyRouterOpenshiftIo.TIMEOUT: "1us"},
+                    }
+                }
+            }
+        ).update()
+
+        # Wait for route to be updated with the annotation
+        time.sleep(10)  # noqa: FCN001
```
dbasunag, line 150: Same comment. Please create a utility function to check if the route was updated.

tests/model_serving/model_server/routes/test_raw_deployment.py
brettmthompson: I have created a utility function that polls the route and waits until it is updated with the proper timeout annotation. However, it seems that the route does not enforce this timeout immediately after the annotation is set, meaning that if we go off of just the annotation being present the next step that queries the model and expects it to fail will actually receive a valid response. 

I do not see any other status or field in the route manifest that would indicate that is actually enforcing the timeout. Hence why I used sleep here. If we can think of a better way to do this I can make the change.

dbasunag: If 10 seconds works consistently, then can you use timeout sampler with 10 seconds sleep time while waiting for the annotations?

brettmthompson: Great suggestion, let me give that a try.

brettmthompson: @dbasunag was able to find some time today to test this. I have a doubt about this approach after looking through the code and testing manually. Given that the sleep time for `TimeoutSampler`'s  [__iter__ method](https://github.com/RedHatQE/timeout-sampler/blob/main/timeout_sampler/__init__.py#L160) is not applied until after the result is yielded, if the first result to be yielded is correct, no delay will be applied and thus the same issue mentioned in the first bullet will persist.

In the case of this annotation being added to the route, it is set very quickly so the first iteration here will yield the correct result thus the sleep time will never be applied.

brettmthompson: I like the approach of using a counter, great suggestion. As for the `get` method, I may have been incorrectly accessing the nested `annotatoins` field within the `Route` object. I was using `route.annotations`, but I think it needs to be `route.instance.metadata.annotations`.

tests/model_serving/model_server/routes/test_raw_deployment.py
brettmthompson: Created fixture in latest commit.

tests/model_serving/model_server/routes/test_raw_deployment.py
brettmthompson: Instituted a counter in the `wait_for_route_timeout` that will loop through the TimeoutSampler until the annotation is detected on the route twice. This allows for the `TimeoutSampler`'s sleep statement to be triggered, providing enough time for the timeout to be applied to the route. I have also updated the method to directly use the init function of the `Route` class instead of calling `get_model_route`.

tests/model_serving/model_server/routes/test_raw_deployment.py
brettmthompson: Created `s3_models_inference_service_patched_annotations` fixture in latest commit.

I was following the existing raw route tests here which have completed separate tests for http vs grpc. I can parameterize them further if we think it is necessary, but if we want to do that for the timeout ones then I think we should do them for all of them.

tests/model_serving/model_server/routes/test_serverless.py
brettmthompson: Created `wait_for_route_timeout` function in latest commit.

tests/model_serving/model_server/conftest.py
```diff
@@ -138,6 +139,24 @@ def s3_models_inference_service(
...
+    if hasattr(request, "param"):
+        ResourceEditor(
+            patches={
+                s3_models_inference_service: {
+                    "metadata": {
+                        "annotations": request.param["annotations"],
+                    }
+                }
+            }
+        ).update()
```
dbasunag, line 155: How about undoing the changes at the end of the test?

tests/model_serving/model_server/routes/test_serverless.py
```diff
@@ -94,3 +98,68 @@ def test_rest_serverless_exposed_label_route(self, patched_s3_caikit_kserve_isvc
...
+        ],
+        indirect=True,
+    )
+    @pytest.mark.dependency(depends=["test_rest_serverless_external_route"])
+    def test_rest_serverless_external_route_with_timeout(self, s3_models_inference_service_patched_annotations):
+        """Test HTTP inference using external route fails when timeout is set too low"""
+        wait_for_route_timeout(
+            name=s3_models_inference_service_patched_annotations.name
+            + "-"
+            + s3_models_inference_service_patched_annotations.namespace,
```
dbasunag, line 152: Please use f-string instead of `+`.

utilities/infra.py
```diff
@@ -905,6 +905,40 @@ def get_operator_distribution(client: DynamicClient, dsc_name: str = "default-ds
...
+    for route in TimeoutSampler(
+        wait_timeout=Timeout.TIMEOUT_30SEC,
+        sleep=10,
+        exceptions_dict={ResourceNotFoundError: []},
+        func=Route,
+        name=name,
+        namespace=namespace,
+        ensure_exists=True,
+    ):
+        annotations = getattr(route.instance.metadata, "annotations", {}) or {}
```
dbasunag, line 934: you can do route.instance.metadata.get('annotations',{}).get(Annotations.HaproxyRouterOpenshiftIo.TIMEOUT) directly in the line below.

brettmthompson: I think code rabbit was nervous that route.instance.metadata.annotations could be `None`, hence why it recommended the `or {}` on a separate line. Will update to your recommendation.

tests/model_serving/model_server/conftest.py
brettmthompson: Updated the fixture scope to `function` and the logic to the below:
```
    if hasattr(request, "param"):
        with ResourceEditor(
            patches={
                s3_models_inference_service: {
                    "metadata": {
                        "annotations": request.param["annotations"],
                    }
                }
            }
        ):
            yield s3_models_inference_service
```

dbasunag (approved):

PR #350, opened 2025-06-12
Test LMEval and Guardrails orchestrator images

sheltoncyril:
Test LMEval and Guardrails orchestrator images
## Description

check if images from configmap are the same as the ones in the pod and check if they are pinned using a sha256 digest.

### DIFF
--- a/tests/model_explainability/guardrails/test_guardrails.py
+++ b/tests/model_explainability/guardrails/test_guardrails.py
@@ -41,7 +41,11 @@ class TestGuardrails:
         assert response_data["services"]["chat_generation"]["status"] == healthy_status
         assert response_data["services"]["regex"]["status"] == healthy_status
 
+    @pytest.mark.smoke
     def test_validate_guardrails_orchestrator_images(self, guardrails_orchestrator_pod, trustyai_operator_configmap):
+        """Test to verify Guardrails pod images.
+        Checks if the image tag from the ConfigMap is used within the Pod and if it's pinned using a sha256 digest.
+        """
         validate_tai_component_images(
             pod=guardrails_orchestrator_pod, tai_operator_configmap=trustyai_operator_configmap
         )
--- a/tests/model_explainability/lm_eval/test_lm_eval.py
+++ b/tests/model_explainability/lm_eval/test_lm_eval.py
@@ -173,9 +173,12 @@ def test_lmeval_s3_storage(
     ],
     indirect=True,
 )
+@pytest.mark.smoke
 def test_verify_lmeval_pod_images(lmevaljob_s3_offline_pod, trustyai_operator_configmap) -> None:
     """Test to verify LMEval pod images.
-    Verify:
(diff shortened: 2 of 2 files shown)


tests/model_explainability/guardrails/test_guardrails.py
```diff
@@ -39,3 +40,20 @@ def test_guardrails_info_endpoint(self, admin_client, qwen_llm_model, guardrails
         response_data = response.json()
         assert response_data["services"]["chat_generation"]["status"] == healthy_status
         assert response_data["services"]["regex"]["status"] == healthy_status
+
+    def test_validate_guardrails_orchestrator_images(self, guardrails_orchestrator_pod, trustyai_operator_configmap):
```
adolfo-ab, line 44: Add smoke tag

tests/model_explainability/guardrails/test_guardrails.py
```diff
@@ -39,3 +40,20 @@ def test_guardrails_info_endpoint(self, admin_client, qwen_llm_model, guardrails
...
+
+        Args:
+
+            guardrails_orchestrator_pod: Pod
+            trustyai_operator_configmap: ConfigMap
+
+        Returns:
+            None
+        Raises:
+            AssertionError: If validation fails.
```
adolfo-ab, line 55: In the tests you don't need to provide the args and returns (a test should never return anything), but you should  explain what the test does, in this case, check against the ConfigMap and check the digest.

tests/model_explainability/guardrails/utils.py
```diff
@@ -0,0 +1,24 @@
...
+
+    Returns:
+        None
+
+    Raises:
+        AssertionError: If validation fails.
+    """
+    validate_pod_image_against_tai_configmap_images_and_check_digest(
+        pod=guardrails_orchestrator_pod, tai_operator_configmap=tai_operator_configmap
+    )
```
adolfo-ab, line 24: I don't think we need this function, we only use it once and the only thing it does is calling `validate_pod_image_against_tai_configmap_images_and_check_digest`

tests/model_explainability/lm_eval/utils.py
```diff
@@ -30,3 +32,21 @@ def get_lmevaljob_pod(client: DynamicClient, lmevaljob: LMEvalJob, timeout: int
...
+
+    Returns:
+        None
+
+    Raises:
+        AssertionError: if pod images don't match or don't have a sha256 digest.
+    """
+    validate_pod_image_against_tai_configmap_images_and_check_digest(
+        pod=lmeval_pod, tai_operator_configmap=tai_operator_configmap, include_init_containers=True
+    )
```
adolfo-ab, line 52: I don't think we need this function, we only use it once and the only thing it does is calling validate_pod_image_against_tai_configmap_images_and_check_digest

adolfo-ab (changes requested):

adolfo-ab (approved):

PR #354, opened 2025-06-16
Simplify RBAC tests

dbasunag:
## Merge criteria:


tests/model_registry/rbac/conftest.py
```diff
@@ -250,3 +350,56 @@ def mr_access_role_binding(
...
+        user=test_idp_user_session.username,
+        password=test_idp_user_session.password,
+    )
+    yield
+    LOGGER.info(f"Logging in as {original_user}")
+    login_with_user_password(
+        api_address=api_server_url,
+        user=original_user,
+    )
+
```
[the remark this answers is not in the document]
dbasunag: This is intentional.

tests/model_registry/rbac/test_mr_rbac.py
```diff
@@ -53,144 +60,71 @@
...
-                mr_client = ModelRegistryClient(**client_args)
-                assert mr_client is not None, "Client initialization failed for admin user"
-                LOGGER.info("Successfully created Model Registry client for admin user")
-            else:
-                with pytest.raises(ForbiddenException) as exc_info:
-                    _ = ModelRegistryClient(**client_args)
-                assert exc_info.value.status == 403, f"Expected HTTP 403 Forbidden, but got {exc_info.value.status}"
-                LOGGER.info("Successfully received expected HTTP 403 status code")
+    @pytest.mark.sanity
+    def test_user_permission_non_admin_user(
```
fege, line 79: same as above

tests/model_registry/rbac/test_mr_rbac.py
```diff
@@ -53,144 +60,71 @@
...
     @pytest.mark.sanity
-    @pytest.mark.parametrize(
-        "use_admin_context",
-        [
-            pytest.param(True, id="admin_user"),
-            pytest.param(False, id="normal_user"),
-        ],
-    )
-    def test_user_permission(
+    def test_user_permission_admin_user(
```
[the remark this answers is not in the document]
dbasunag: Yes, to keep it simple.

tests/model_registry/rbac/test_mr_rbac.py
dbasunag: Same response.

utilities/user_utils.py
```diff
@@ -155,96 +65,26 @@ def _create_htpasswd_file(username: str, password: str) -> tuple[Path, str]:
...
-        # Login was successful, get the current context
-        _, stdout, _ = run_command(command=["oc", "config", "current-context"], check=True)
-        user_context = stdout.strip()
-        LOGGER.info(f"Successfully created context for user {username}: {user_context}")
-        return user_context
-
-    LOGGER.error(f"Login failed for user {username}")
-    return None
+    res = login_with_user_password(api_address=cluster_url, user=username, password=password)
+    login_with_user_password(api_address=cluster_url, user=original_user)
```
fege, line 87: I don't understand why you log in again

dbasunag: They should be together. User does not have same priviledges as admin user, so whereever we are logging in as the user, if that is successful, we should undo it. Do we want me to add a try/finally block?

tests/model_registry/rbac/conftest.py
```diff
@@ -159,18 +161,116 @@ def model_registry_group_with_user(
...
 
+@pytest.fixture(scope="package")
+def user_credentials_rbac() -> dict[str, str]:
+    random_str = generate_random_name()
+    return {
+        "username": f"test-user-{random_str}",
+        "password": f"test-password-{random_str}",
+        "idp_name": f"test-htpasswd-idp-{random_str}",
+        "secret_name": f"test-htpasswd-secret-{random_str}",
+    }
```
lugi0, line 172: doesn't this force to have a single user in the idp for the whole execution, given the scope of the fixture?

tests/model_registry/rbac/conftest.py
```diff
@@ -159,18 +161,116 @@ def model_registry_group_with_user(
...
 @pytest.fixture(scope="session")
-def test_idp_user_session() -> Generator[UserTestSession, None, None]:
+def original_user() -> str:
+    current_user = run_command(command=["oc", "whoami"])[1].strip()
+    LOGGER.info(f"Original user: {current_user}")
+    return current_user
+
+
+@pytest.fixture(scope="module")
+def created_htpasswd_secret(
```
lugi0, line 183: does it make sense to have this at this scope? the data for the user_credentials has a wider scope.
I also think this forces us into having a single instance of the idp/user/etc. in the scope, is that what we actually want?

tests/model_registry/rbac/conftest.py
```diff
@@ -159,18 +161,116 @@ def model_registry_group_with_user(
...
+            wait_for_resource=True,
+        ) as secret:
+            yield secret
+    finally:
+        # Clean up the temporary file
+        temp_path.unlink(missing_ok=True)
+
+
+@pytest.fixture(scope="module")
+def updated_oauth_config(
```
lugi0, line 210: same question as above

tests/model_registry/rbac/conftest.py
```diff
@@ -250,3 +350,56 @@ def mr_access_role_binding(
...
+        api_address=api_server_url,
+        user=test_idp_user_session.username,
+        password=test_idp_user_session.password,
+    )
+    yield
+    LOGGER.info(f"Logging in as {original_user}")
+    login_with_user_password(
+        api_address=api_server_url,
+        user=original_user,
+    )
```
lugi0, line 370: why using logins instead of context? with context we should have easier code since we don't really need to keep track of anything other than the context name

tests/model_registry/rbac/test_mr_rbac.py
```diff
@@ -53,144 +60,71 @@
...
+        """
+        This test verifies that non-admin users cannot access the Model Registry (403 Forbidden)
+        """
+        client_args = build_mr_client_args(
+            rest_endpoint=model_registry_instance_rest_endpoint, token=get_openshift_token()
+        )
+        with pytest.raises(ForbiddenException) as exc_info:
+            ModelRegistryClient(**client_args)
+        assert exc_info.value.status == 403, f"Expected HTTP 403 Forbidden, but got {exc_info.value.status}"
+        LOGGER.info("Successfully received expected HTTP 403 status code")
```
lugi0, line 94: Are you sure `.value.status` exists? I think the only supported syntax is `.value` from pytest: https://docs.pytest.org/en/stable/reference/reference.html#pytest.ExceptionInfo

utilities/user_utils.py
```diff
@@ -28,111 +25,24 @@ class UserTestSession:
...
-    """
-    Context manager to create and manage a test HTPasswd IDP in OpenShift.
-    Creates the IDP and test user, then cleans up after use.
-
-    Args:
-        idp_name: Name for the IDP
-        secret_name: Name for the secret
+        user = User(name=self.username)
+        if user.exists:
+            user.delete()
```
lugi0, line 42: shouldn't this be handled by the teardown of the fixture creating the user in the first place?

tests/model_registry/rbac/conftest.py
dbasunag: test_idp_user_session was a session scoped fixture originally. I changed it and any associated fixtures to package scoped. It can be changed to module too, since outside of test_mr_rbac.py, these fixtures are not used anywhere. This is creating the dict to be used with respect to test_idp_user_session. I don't understand the concern though, as it does not change any other users or logs in as this user.

tests/model_registry/rbac/conftest.py
dbasunag: You are correct. I see a scope mismatch. I will change the test_idp_user_session and associated fixtures to module. Currently none of these are being used anywhere outside of rbac, so we should either stick with module or package scoped fixtures. If that changes, we can always update the scopes.

tests/model_registry/rbac/conftest.py
dbasunag: Same response. This needs to be updated or it would introduce another bug that would be uncovered if these needs to be parameterized.

tests/model_registry/rbac/test_mr_rbac.py
dbasunag: I did test with changing this value and it worked. Also, this code was not changed https://github.com/opendatahub-io/opendatahub-tests/blob/main/tests/model_registry/rbac/test_mr_rbac.py#L119. So far I see it is working fine.

tests/model_registry/conftest.py
```diff
@@ -205,9 +209,7 @@ def model_registry_instance_service(
         Service: The service for the model registry instance
     """
     return get_mr_service_by_label(
-        client=admin_client,
-        ns=Namespace(name=model_registry_namespace),
-        mr_instance=model_registry_instance,
+        client=admin_client, namespace_name=model_registry_namespace, mr_instance=model_registry_instance
```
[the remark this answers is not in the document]
dbasunag: No, it is a string, not a namespace object. We should not have to create a python object, just so that get_mr_service_by_label() is able to call namespace.name.

tests/model_registry/utils.py
```diff
@@ -19,11 +18,11 @@
 LOGGER = get_logger(name=__name__)
 
 
-def get_mr_service_by_label(client: DynamicClient, ns: Namespace, mr_instance: ModelRegistry) -> Service:
+def get_mr_service_by_label(client: DynamicClient, namespace_name: str, mr_instance: ModelRegistry) -> Service:
     """
     Args:
         client (DynamicClient): OCP Client to use.
-        ns (Namespace): Namespace object where to find the Service
+        namespace_name (str): Namespace name associated with the service
```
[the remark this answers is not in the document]
dbasunag: Avoiding, unnecessary object creation to call this function. get_mr_service_by_label() needs to create a namespace object, so that this call works. This comment of yours is related to https://github.com/opendatahub-io/opendatahub-tests/pull/354/files#r2149906679.

utilities/user_utils.py
dbasunag: I don't see it happening, I saw a bunch of leftover users instead.

utilities/user_utils.py
```diff
@@ -155,96 +65,26 @@ def _create_htpasswd_file(username: str, password: str) -> tuple[Path, str]:
...
-        user_context = stdout.strip()
-        LOGGER.info(f"Successfully created context for user {username}: {user_context}")
-        return user_context
-
-    LOGGER.error(f"Login failed for user {username}")
-    return None
+    res = login_with_user_password(api_address=cluster_url, user=username, password=password)
+    login_with_user_password(api_address=cluster_url, user=original_user)
+    if res:
+        return True
```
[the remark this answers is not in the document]
dbasunag: It ensures we can login as the new user. Moved the original user part to the fixture calling it, based on @fege 's comment.

utilities/user_utils.py
```diff
@@ -155,96 +68,26 @@ def _create_htpasswd_file(username: str, password: str) -> tuple[Path, str]:
...
-        yield
-
-
-def _create_user_context(username: str, password: str, cluster_url: str) -> Optional[str]:
+@retry(
+    wait_timeout=240,
+    sleep=10,
+    exceptions_dict={ExceptionUserLogin: []},
+)
+def wait_for_user_creation(username: str, password: str, cluster_url: str) -> bool:
```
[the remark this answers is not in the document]
dbasunag: Not needed.

tests/model_registry/rbac/conftest.py
lugi0: My concern is that this generates a single user for the idp. I'm confident that we will require more than a single user in the future, and this fixture (along with any related ones relying on this dict) will need to be updated

fege (approved):
/lgtm

tests/model_registry/conftest.py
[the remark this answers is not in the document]
dbasunag: Why would we. We should use objects when it is needed (e.g. you need to access many of the object attributes). We should not have to use namespace object just to access it's name, when name is the only required attribute to create namespace.

tests/model_registry/utils.py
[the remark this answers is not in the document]
dbasunag: Answered in the other thread. It depends on the situation. If I need to access multiple attributes of an object, it makes sense to create one vs passing them individually to the function.

PR #360, opened 2025-06-18
fix: redact sensitive strings

adolfo-ab:
Redact sensitive strings for safer logging

## Description
Now instead of printing the value of the fixture when failing, we print this:
```
________________________________________________________________________________ ERROR at setup of TestDriftMetrics.test_drift_send_inference_and_verify_trustyai_service[model_namespace0-minio_pod0-minio_data_connection0] _________________________________________________________________________________

current_client_token = '***REDACTED***'


TEST: TestDriftMetrics.test_drift_send_inference_and_verify_trustyai_service[model_namespace0-minio_pod0-minio_data_connection0] [setup] STATUS: ERROR
    def dummy_fixture(current_client_token) -> str:
>       raise ValueError("DUMMY FAILURE")
E       ValueError: DUMMY FAILURE
```

## How Has This Been Tested?
Running on PSI

## Merge criteria:

### DIFF
--- a/utilities/logger.py
+++ b/utilities/logger.py
@@ -10,6 +10,13 @@ LOGGER = logging.getLogger(__name__)
 
 
 class RedactedString(str):
+    """
+    Used to redact the representation of a sensitive string.
+    """
+
+    def __new__(cls, *, value: object) -> "RedactedString":
+        return super().__new__(cls, value)
+
     def __repr__(self) -> str:
         return "'***REDACTED***'"
 
--- a/utilities/logger.py
+++ b/utilities/logger.py
@@ -9,6 +9,11 @@ from simple_logger.logger import DuplicateFilter, WrapperLogFormatter
 LOGGER = logging.getLogger(__name__)
 
 
+class RedactedString(str):
+    def __repr__(self) -> str:
+        return "'***REDACTED***'"
+
+
 def setup_logging(log_level: int, log_file: str = "/tmp/pytest-tests.log") -> QueueListener:
     """
     Setup basic/root logging using QueueHandler/QueueListener
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -69,7 +69,7 @@ def tests_tmp_dir(request: FixtureRequest, tmp_path_factory: TempPathFactory) ->
 
 @pytest.fixture(scope="session")
 def current_client_token(admin_client: DynamicClient) -> str:
-    return RedactedString(object=get_openshift_token())
+    return RedactedString(value=get_openshift_token())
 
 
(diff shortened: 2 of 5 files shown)


dbasunag (approved):

lugi0 (approved):

utilities/logger.py
```diff
@@ -9,6 +9,11 @@
 LOGGER = logging.getLogger(__name__)
 
 
+class RedactedString(str):
+    def __repr__(self) -> str:
+        return "'***REDACTED***'"
+
```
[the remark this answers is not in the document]
sheltoncyril: @adolfo-ab could you PTAL at this?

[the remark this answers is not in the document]
adolfo-ab: lol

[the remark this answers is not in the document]
adolfo-ab: Fixed it

utilities/logger.py
```diff
@@ -9,6 +9,18 @@
 LOGGER = logging.getLogger(__name__)
 
 
+class RedactedString(str):
+    """
+    Used to redact the representation of a sensitive string.
+    """
+
+    def __new__(cls, *, value: object) -> "RedactedString":
```
dbasunag, line 17: Can we use __init__ please?

mwaykole (approved):

utilities/logger.py
adolfo-ab: str is immutable so if we do:

```
class RedactedStringWithAttribute(str):
    def __init__(self, *, value: object) -> None:
        self.value = value
```

We're setting the `value` attribute, but not what the string contains. The string content itself remains empty/default because str objects are created during __new__(), not __init__(). I had to look that up. Tested it anyways and indeed doesn't work.

See for reference:
https://medium.com/@wazedkhan111024/sub-classing-pythons-extending-built-in-immutable-classes-in-python-a06c9ecb759d
https://stackoverflow.com/questions/3120562/python-subclassing-immutable-types

sheltoncyril, line 17: Just curious, why is this a key-worded argument and why not allow positional arguments?
`RedactedString("TEST")` should work just fine right? and what happens when we have multiple values returned?
`RedactedString(( OR [ "TEST1", "TEST2"] OR ))` should we cover this case as well?

adolfo-ab: TBH I just use keyword arguments for consistency with the rest of the codebase. About your second question, we can do that when/if we need it I think.

dbasunag (approved):

sheltoncyril (approved):
/lgtm

lugi0 (approved):

PR #371, opened 2025-06-20
Add first pre/post upgrade skeleton for Model Registry

lugi0:
## Description

This draft PR adds a skeleton for pre and post upgrade tests for Model Registry. They currently don't do anything too interesting (register a model in pre-upgrade, retrieve it in post), but they implement the logic needed to keep the environment we set up prior to the upgrade and tear it down after the upgrade. 
Due to the current implementation of our fixtures it needed a fair amount of refactoring and additional checks needed to be implemented in order to not try to create duplicate resources; this might not be the best approach overall but it's the only solution I could come up with within our constraints. 

## How Has This Been Tested?

Tested it locally (without triggering a RHOAI upgrade in between) to check the validity of the setup/teardown logic between the two scenarios.

## Merge criteria:

### DIFF
--- a/tests/model_registry/conftest.py
+++ b/tests/model_registry/conftest.py
@@ -47,6 +47,7 @@ from utilities.constants import Protocols, DscComponents
 from model_registry import ModelRegistry as ModelRegistryClient
 from semver import Version
 from utilities.general import wait_for_pods_by_labels
+from kubernetes.dynamic.exceptions import ResourceNotFoundError
 
 LOGGER = get_logger(name=__name__)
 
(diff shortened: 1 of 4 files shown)


lugi0:
@coderabbit review

lugi0:
@coderabbitai review

tests/model_registry/conftest.py
```diff
@@ -60,121 +61,190 @@ def model_registry_namespace(updated_dsc_component_state_scope_class: DataScienc
...
-            "name": DB_RESOURCES_NAME,
-        },
-        label=get_model_registry_db_label_dict(db_resource_name=DB_RESOURCES_NAME),
-        annotations={
-            "template.openshift.io/expose-uri": r"mysql://{.spec.clusterIP}:{.spec.ports[?(.name==\mysql\)].port}",
-        },
-    ) as mr_db_service:
+    if pytestconfig.option.post_upgrade:
+        mr_db_service = next(
+            Service.get(client=admin_client, name=DB_RESOURCES_NAME, namespace=model_registry_namespace), None
```
dbasunag, line 71: You already know the name and namespace. You don't need get() call. Instead, you can make call like this 
```
yield Secret(name=name=DB_RESOURCES_NAME, namespace=model_registry_namespace, ensure_exists=True)

File: tests/model_registry/conftest.py [hunk 2]
```diff
@@ -60,121 +61,190 @@ def model_registry_namespace(updated_dsc_component_state_scope_class: DataScienc
...
-        yield pvc
+    if pytestconfig.option.post_upgrade:
+        mr_db_pvc = next(
+            PersistentVolumeClaim.get(client=admin_client, name=DB_RESOURCES_NAME, namespace=model_registry_namespace),
+            None,
+        )
+        if not mr_db_pvc:
+            raise ResourceNotFoundError(
+                f"PersistentVolumeClaim {DB_RESOURCES_NAME} not found in namespace {model_registry_namespace}"
+            )
```
dbasunag, line 121: Ditto.

tests/model_registry/conftest.py
```diff
@@ -60,121 +61,190 @@ def model_registry_namespace(updated_dsc_component_state_scope_class: DataScienc
...
-        string_data=MODEL_REGISTRY_DB_SECRET_STR_DATA,
-        label=get_model_registry_db_label_dict(db_resource_name=DB_RESOURCES_NAME),
-        annotations=MODEL_REGISTRY_DB_SECRET_ANNOTATIONS,
-    ) as mr_db_secret:
+    if pytestconfig.option.post_upgrade:
+        mr_db_secret = next(
+            Secret.get(client=admin_client, name=DB_RESOURCES_NAME, namespace=model_registry_namespace), None
+        )
+        if not mr_db_secret:
+            raise ResourceNotFoundError(f"Secret {DB_RESOURCES_NAME} not found in namespace {model_registry_namespace}")
```
dbasunag, line 149: Ditto.

tests/model_registry/conftest.py
```diff
@@ -60,121 +61,190 @@ def model_registry_namespace(updated_dsc_component_state_scope_class: DataScienc
...
-        mr_db_deployment.wait_for_replicas(deployed=True)
-        yield mr_db_deployment
+    if pytestconfig.option.post_upgrade:
+        db_deployment = next(
+            Deployment.get(client=admin_client, name=DB_RESOURCES_NAME, namespace=model_registry_namespace), None
+        )
+        if not db_deployment:
+            raise ResourceNotFoundError(
+                f"Deployment {DB_RESOURCES_NAME} not found in namespace {model_registry_namespace}"
+            )
```
dbasunag, line 182: Ditto.

tests/model_registry/conftest.py
```diff
@@ -60,121 +61,190 @@ def model_registry_namespace(updated_dsc_component_state_scope_class: DataScienc
...
-        istio=None,
-        oauth_proxy=oauth_config,
-        mysql=model_registry_mysql_config,
-        wait_for_resource=True,
-    ) as mr:
-        mr.wait_for_condition(condition="Available", status="True")
-        yield mr
+    if pytestconfig.option.post_upgrade:
+        mr_instance = next(
+            ModelRegistry.get(client=admin_client, name=MR_INSTANCE_NAME, namespace=model_registry_namespace), None
```
dbasunag, line 218: Ditto

tests/model_registry/conftest.py
```diff
@@ -277,11 +347,18 @@ def after_call(self, response: Response, case: Case) -> None:
...
     dsc_resource: DataScienceCluster,
     admin_client: DynamicClient,
-    is_model_registry_oauth: bool,
+    teardown_resources: bool,
 ) -> Generator[DataScienceCluster, Any, Any]:
+    if not teardown_resources or pytestconfig.option.post_upgrade:
+        # if we are not tearing down resources or we are in post upgrade, we don't need to do anything
+        # the pre_upgrade/post_upgrade fixtures will handle the rest
+        yield dsc_resource
+        return
```
dbasunag, line 360: Why both yield and return?

tests/model_registry/conftest.py
```diff
@@ -316,7 +393,68 @@ def updated_dsc_component_state_scope_class(
...
+    admin_client: DynamicClient,
+) -> DataScienceCluster:
+    original_components = dsc_resource.instance.spec.components
+    component_patch = {DscComponents.MODELREGISTRY: {"managementState": DscComponents.ManagementState.MANAGED}}
+    if (
+        original_components.get(DscComponents.MODELREGISTRY).get("managementState")
+        == DscComponents.ManagementState.MANAGED
+    ):
+        LOGGER.error("Model Registry is already set to Managed before upgrade - was this intentional?")
+        return dsc_resource
```
dbasunag, line 407: We should fail.

tests/model_registry/conftest.py
```diff
@@ -316,7 +393,68 @@ def updated_dsc_component_state_scope_class(
...
+    # the state we found after the upgrade
+    original_components = dsc_resource.instance.spec.components
+    # We don't have an easy way to figure out the state of the components before the upgrade at runtime
+    # For now we know that MR has to go back to Removed after post upgrade tests are run
+    component_patch = {DscComponents.MODELREGISTRY: {"managementState": DscComponents.ManagementState.REMOVED}}
+    if (
+        original_components.get(DscComponents.MODELREGISTRY).get("managementState")
+        == DscComponents.ManagementState.REMOVED
+    ):
+        LOGGER.error("Model Registry is already set to Removed after upgrade - was this intentional?")
```
dbasunag, line 440: Ditto. We should fail.

tests/model_registry/conftest.py
```diff
@@ -372,11 +510,6 @@ def model_registry_instance_pod(admin_client: DynamicClient) -> Generator[Pod, A
     )[0]
 
 
-@pytest.fixture(scope="class")
```
dbasunag, line 375: We should keep it.

tests/model_registry/upgrade/test_model_registry_upgrade.py
```diff
@@ -0,0 +1,58 @@
...
+)
+@pytest.mark.usefixtures("pre_upgrade_dsc_patch")
+class TestPreUpgradeModelRegistry:
+    @pytest.mark.pre_upgrade
+    def test_registering_model_pre_upgrade(
+        self: Self,
+        model_registry_client: ModelRegistryClient,
+        registered_model: RegisteredModel,
+    ):
+        pass
```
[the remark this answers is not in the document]
dbasunag: Instead of pass, let's do basic get calls for registered model etc.

tests/model_registry/upgrade/test_model_registry_upgrade.py
```diff
@@ -0,0 +1,58 @@
...
+        ]
+        if errors:
+            LOGGER.error(f"received model: {model}")
+            pytest.fail("errors found in model registry response validation:\n{}".format("\n".join(errors)))
+
+        # TODO: if we are in >= 2.22, we can check that this is using oauth instead of servicemesh
+        # TODO: if we are in >= 2.22, we can check that the conversion webhook is working as expected, i.e.
+        # the MR instance has api version v1beta1, and that when querying for older api version (e.g.
+        # `oc get modelregistries.v1alpha1.modelregistry.opendatahub.io -o wide -n rhoai-model-registries`)
+        # it also returns the status stanza (used by dashboard)
```
dbasunag, line 58: These can be follow up pr. but we expect to run with this branch on only 2.22 onwards.

tests/model_registry/conftest.py
lugi0: Because it's a Generator so we need to yield, and the execution during the teardown should do nothing and only return

tests/model_registry/conftest.py
lugi0: I had tried doing this, but then the yielded resource would create problems for other fixtures/function (e.g. in the mysql config we would be unable to get the Secret data (database/username)) - we would need to return `resource.instance`, or call `.instance` on the resource in the other code, but this would not work for the usual resource yielded by the fixture when we are not in post upgrade. I've updated at least one fixture that had problems with this to use values from constants instead, so maybe it's worth another try. I'll let you know how it goes.

lugi0: Ok, tried again, I think with the changes I had made to the mysql fixture we are good with yielding the resources like this as I didn't encounter any failures in pre/post upgrade. Let's keep an eye on this if it were to ever fail.

fege (dismissed):
/lgtm

tests/model_registry/conftest.py
lugi0: fixed with the if/else

dbasunag (approved):

fege (approved):
/lgtm

tests/model_registry/conftest.py
```diff
@@ -61,128 +61,171 @@ def model_registry_namespace(updated_dsc_component_state_scope_class: DataScienc
 
 @pytest.fixture(scope="class")
 def model_registry_db_service(
-    admin_client: DynamicClient, model_registry_namespace: str, is_model_registry_oauth: bool
+    pytestconfig: Config,
+    admin_client: DynamicClient,
+    model_registry_namespace: str,
+    teardown_resources: bool,
```
dbasunag, line 67: We need is_model_registry_oauth here.

tests/model_registry/conftest.py
```diff
@@ -61,128 +61,171 @@ def model_registry_namespace(updated_dsc_component_state_scope_class: DataScienc
...
+                "template.openshift.io/expose-uri": r"mysql://{.spec.clusterIP}:{.spec.ports[?(.name==\mysql\)].port}",
+            },
+            teardown=teardown_resources,
+        ) as mr_db_service:
+            yield mr_db_service
 
 
 @pytest.fixture(scope="class")
 def model_registry_db_pvc(
-    admin_client: DynamicClient, model_registry_namespace: str, is_model_registry_oauth: bool
```
dbasunag, line 93: ditto. is_model_registry_oauth is needed here.

tests/model_registry/conftest.py
```diff
@@ -61,128 +61,171 @@ def model_registry_namespace(updated_dsc_component_state_scope_class: DataScienc
...
+        ) as pvc:
+            yield pvc
 
 
 @pytest.fixture(scope="class")
 def model_registry_db_secret(
+    pytestconfig: Config,
     admin_client: DynamicClient,
     model_registry_namespace: str,
-    is_model_registry_oauth: bool,
```
dbasunag, line 110: We need is_model_registry_oauth here.

tests/model_registry/conftest.py
```diff
@@ -61,128 +61,171 @@ def model_registry_namespace(updated_dsc_component_state_scope_class: DataScienc
...
 
 @pytest.fixture(scope="class")
 def model_registry_db_deployment(
+    pytestconfig: Config,
     admin_client: DynamicClient,
     model_registry_namespace: str,
     model_registry_db_secret: Secret,
     model_registry_db_pvc: PersistentVolumeClaim,
     model_registry_db_service: Service,
-    is_model_registry_oauth: bool,
```
dbasunag, line 130: ditto

tests/model_registry/conftest.py
```diff
@@ -285,46 +328,113 @@ def after_call(self, response: Response, case: Case) -> None:
 
 @pytest.fixture(scope="class")
 def updated_dsc_component_state_scope_class(
+    pytestconfig: Config,
     request: FixtureRequest,
     dsc_resource: DataScienceCluster,
     admin_client: DynamicClient,
-    is_model_registry_oauth: bool,
```
dbasunag, line 291: ditto. This is needed.

dbasunag (changes requested):
Those missing is_model_registry_oauth are causing test failures.

dbasunag (dismissed):

dbasunag (approved):

PR #384, opened 2025-06-26
add markers for multi-node and model server-gpu

mwaykole:


israel-hdez (approved):
Thanks for the quick follow-up @mwaykole !

PR #427, opened 2025-07-11
feat: add pre-commit to uv dependencies

sheltoncyril:
adding pre-commit as a dependency as pre-commit gets uninstalled when `uv sync` is performed

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -70,6 +70,7 @@ dependencies = [
     "pytest-html>=4.1.1",
     "fire",
     "llama_stack_client==0.2.10",
+    "pre-commit>=4.2.0",
 ]
 
 [project.urls]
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -70,6 +70,7 @@ dependencies = [
     "pytest-html>=4.1.1",
     "fire",
     "llama_stack_client==0.2.10",
+    "pre-commit>=4.2.0",
```
dbasunag, line 73: Precommit should not be needed with in uv. Why do we need it?

dbasunag (changes requested):

pyproject.toml
sheltoncyril: Got it, closing it as this was discussed offline. Thanks!

PR #443, opened 2025-07-16
fix: teardown issue where unprivileged user cannot delete namespace

mwaykole:


Raghul-M (approved):

utilities/infra.py
```diff
@@ -138,32 +137,23 @@ def create_ns(
...
-            if _labels := namespace_kwargs.get("label", {}):
-                # To patch the namespace, admin client is required
-                ns = Namespace(client=get_client(), name=name)
-                ResourceEditor({
-                    ns: {
-                        "metadata": {
-                            "labels": _labels,
-                        }
-                    }
-                }).update()
```
rnetser, line 154: this is still needed for unpriv ns

rnetser (changes requested):

Raghul-M (approved):

PR #461, opened 2025-07-25
Parametrized fixtures for multitenancy

fege:
## Merge criteria:


tests/global_config.py
```diff
@@ -9,6 +9,7 @@
 distribution: str = "downstream"
 applications_namespace: str = "redhat-ods-applications"
 model_registry_namespace: str = "rhoai-model-registries"
+num_mr_instances: int = 2
```
lugi0, line 12: I think this shouldn't be in the global config, maybe in our config / constants / utils

tests/model_registry/conftest.py
```diff
@@ -535,236 +536,220 @@ def model_registry_pod(admin_client: DynamicClient, model_registry_namespace: st
...
+    # Get the namespace name from the parameter
+    namespace_name = request.param["ns_name"]
+    original_components = dsc_resource.instance.spec.components
 
+    # First, remove the Model Registry to allow namespace changes
+    remove_patch = {
+        DscComponents.MODELREGISTRY: {
+            "managementState": DscComponents.ManagementState.REMOVED,
+        },
+    }
```
lugi0, line 562: The only concern I have left for the PR is this remove patch, other than that LGTM

tests/model_registry/conftest.py
```diff
@@ -535,236 +536,220 @@ def model_registry_pod(admin_client: DynamicClient, model_registry_namespace: st
...
+def updated_dsc_component_state_parametrized(
+    request: FixtureRequest,
+    admin_client: DynamicClient,
+    dsc_resource: DataScienceCluster,
+    teardown_resources: bool,
+) -> Generator[DataScienceCluster, Any, Any]:
+    """Configure DSC to use parametrized Model Registry namespace"""
+    if not teardown_resources:
+        yield dsc_resource
+        return
```
dbasunag, line 551: No need to have yield and return.

tests/model_registry/conftest.py
```diff
@@ -535,236 +536,220 @@ def model_registry_pod(admin_client: DynamicClient, model_registry_namespace: st
...
-        name=db_name_1,
-        namespace=model_registry_namespace,
-        accessmodes="ReadWriteOnce",
-        size="5Gi",
-        label=get_model_registry_db_label_dict(db_resource_name=db_name_1),
-        teardown=teardown_resources,
-    ) as pvc:
-        yield pvc
+    # Get the namespace name from the parameter
+    namespace_name = request.param["ns_name"]
```
dbasunag, line 554: Please add default

tests/model_registry/conftest.py
dbasunag: Is MR enabled by default? I don't understand this part.

dbasunag (changes requested):

tests/model_registry/conftest.py
fege: yup, removing it

tests/model_registry/multiple_instance_utils.py
```diff
@@ -0,0 +1,55 @@
...
+    pytest.param(
+        ns_params,  # updated_dsc_component_state_parametrized (expects dict)
+        db_secret_params,
+        db_pvc_params,
+        db_service_params,
+        db_deployment_params,
+        model_registry_instance_params,
+        id=f"mr-scenario-{len(db_names)}-instances",  # Unique ID for pytest output
+    )
+]
```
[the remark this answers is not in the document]
fege: my point here is that you can build your own list of pytest.param based on your test case, so for now is only one but in the near future I hope that we can have ALL_MR_TEST_SCENARIOS_X with a different set of params, if I put in a function it will defeat the scope

tests/model_registry/rbac/test_mr_rbac.py
```diff
@@ -137,70 +139,78 @@ def test_add_single_user_role_binding(
...
-        db_pvc_2: PersistentVolumeClaim,
-        db_service_2: Service,
-        db_deployment_2: Deployment,
-        model_registry_instance_2: ModelRegistry,
+        updated_dsc_component_state_parametrized: DataScienceCluster,
+        db_secret_parametrized: List[Secret],
+        db_pvc_parametrized: List[PersistentVolumeClaim],
+        db_service_parametrized: List[Service],
+        db_deployment_parametrized: List[Deployment],
+        model_registry_instance_parametrized: List[ModelRegistry],
```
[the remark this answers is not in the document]
fege: Agree, but before to reach that we need to change all the other tests to not use the non parametrize fixture, and if we agree with the approach I will take care in a follow pr

adolfo-ab (approved):

PR #488, opened 2025-08-04
Add embeddings and vector_io tests

jgarciao:
- Add test to verify embedding model functionality and vector generation.
- Add tests to verify  registering, inserting and retrieving information from a milvus vector db database
- Based on the example available at:  https://llama-stack.readthedocs.io/en/latest/building_applications/rag.html


tests/rag/test_rag.py
```diff
@@ -77,6 +79,87 @@ def test_rag_basic_inference(self, rag_lls_client: LlamaStackClient) -> None:
...
+            assert len(chunks_response.chunks) > 0
+            assert chunks_response.chunks[0].metadata["document_id"] == "doc1"
+            assert chunks_response.chunks[0].metadata["source"] == "precomputed"
+
+        finally:
+            # Cleanup: unregister the vector database to prevent resource leaks
+            try:
+                rag_lls_client.vector_dbs.unregister(vector_db_id)
+            except Exception as e:
+                LOGGER.warning(f"Failed to unregister vector database {vector_db_id}: {e}")
```
adolfo-ab, line 161: Just out of curiosity, isn't this solved just by cleaning up the LlamaStackDistribution and the NS?

adolfo-ab (approved):

tests/rag/test_rag.py
jgarciao: I was hesitating about this, but as we create the LlamaStackDistribution and namespace just once for all the tests in this class I thought it was better to delete it here, so the next test won't have left overs.

jgarciao:
> Looks good to me. I'm concerned, though, that we might have too many smoke tests. Perhaps we can move some of them to a higher tier.
@jiripetrlik , it is a valid concern. Currently, all rag tests (including these 2 new) run in 3m 30s in  my machine. In my opinion, we can merge this PR and move some tests to sanity when we have more or they require more time

jiripetrlik (approved):
Yes @jgarciao , it is true. Let's merge this PR and decide about moving tests to sanity in the future.

tests/rag/test_rag.py
adolfo-ab: Makes sense, ty for the clarification @jgarciao

adolfo-ab (approved):

PR #501, opened 2025-08-07
test(workbenches): add tests for workbench managed by kueue

jstourac:
https://issues.redhat.com/browse/RHOAIENG-30290

## How Has This Been Tested?
```
uv run pytest tests/workbenches/notebook-controller/test_kueue_notebook_spawning.py -v -s
```

## Merge criteria:

### DIFF
--- a/tests/workbenches/notebook-controller/conftest.py
+++ b/tests/workbenches/notebook-controller/conftest.py
@@ -1,0 +1,245 @@
+from typing import Generator, Any, Dict
+
+import pytest
+from kubernetes.dynamic import DynamicClient
+from _pytest.fixtures import FixtureRequest
+from pytest_testconfig import config as py_config
+
+from utilities.kueue_utils import (
+    create_local_queue,
+    create_cluster_queue,
+    create_resource_flavor,
+    LocalQueue,
+    ClusterQueue,
+    ResourceFlavor,
+)
+from ocp_resources.namespace import Namespace
+from ocp_resources.config_map import ConfigMap
+from ocp_resources.deployment import Deployment
+from ocp_resources.persistent_volume_claim import PersistentVolumeClaim
+from ocp_resources.resource import ResourceEditor
+from utilities.constants import Labels, Annotations
+import yaml
+import logging
+
+LOGGER = logging.getLogger(__name__)
+
+
+def kueue_resource_groups_for_notebooks(
+    flavor_name: str,
+    cpu_quota: int,
+    memory_quota: str,
+) -> list[Dict[str, Any]]:
+    """Create resource groups configuration for Kueue with notebook-specific resources"""
+    return [
+        {
+            "coveredResources": ["cpu", "memory"],
+            "flavors": [
+                {
+                    "name": flavor_name,
+                    "resources": [
+                        {"name": "cpu", "nominalQuota": cpu_quota},
+                        {"name": "memory", "nominalQuota": memory_quota},
+                    ],
+                }
+            ],
+        }
+    ]
+
+
+@pytest.fixture(scope="class")
+def kueue_notebook_resource_flavor(
+    request: FixtureRequest,
+    admin_client: DynamicClient,
+) -> Generator[ResourceFlavor, Any, Any]:
+    """Create a ResourceFlavor for notebook workloads"""
+    with create_resource_flavor(
+        client=admin_client,
+        name=request.param["name"],
+        teardown=True,
+    ) as resource_flavor:
+        yield resource_flavor
+
+
+@pytest.fixture(scope="class")
+def kueue_notebook_cluster_queue(
+    request: FixtureRequest,
+    admin_client: DynamicClient,
+) -> Generator[ClusterQueue, Any, Any]:
+    """Create a ClusterQueue for notebook workloads"""
+    resource_groups = kueue_resource_groups_for_notebooks(
+        flavor_name=request.param["resource_flavor_name"],
+        cpu_quota=request.param["cpu_quota"],
+        memory_quota=request.param["memory_quota"],
+    )
+
+    with create_cluster_queue(
+        client=admin_client,
+        name=request.param["name"],
+        resource_groups=resource_groups,
+        namespace_selector=request.param.get("namespace_selector", {}),
+        teardown=True,
+    ) as cluster_queue:
+        yield cluster_queue
+
+
+@pytest.fixture(scope="class")
+def kueue_notebook_local_queue(
+    request: FixtureRequest,
+    admin_client: DynamicClient,
+    kueue_enabled_notebook_namespace: Namespace,
+) -> Generator[LocalQueue, Any, Any]:
+    """Create a LocalQueue for notebook workloads"""
+    with create_local_queue(
+        client=admin_client,
+        name=request.param["name"],
+        cluster_queue=request.param["cluster_queue"],
+        namespace=kueue_enabled_notebook_namespace.name,
+        teardown=True,
+    ) as local_queue:
+        yield local_queue
+
+
+@pytest.fixture(scope="function")
+def kueue_notebook_persistent_volume_claim(
+    request: FixtureRequest,
+    kueue_enabled_notebook_namespace: Namespace,
+    unprivileged_client: DynamicClient,
+) -> Generator[PersistentVolumeClaim, Any, Any]:
+    """Create a PersistentVolumeClaim in the Kueue-enabled namespace"""
+    with PersistentVolumeClaim(
+        client=unprivileged_client,
+        name=request.param["name"],
+        namespace=kueue_enabled_notebook_namespace.name,
+        label={Labels.OpenDataHub.DASHBOARD: "true"},
+        accessmodes=PersistentVolumeClaim.AccessMode.RWO,
+        size="10Gi",
+        volume_mode=PersistentVolumeClaim.VolumeMode.FILE,
+    ) as pvc:
+        yield pvc
+
+
+@pytest.fixture(scope="class")
+def kueue_enabled_notebook_namespace(
+    request: FixtureRequest,
+    admin_client: DynamicClient,
+    unprivileged_client: DynamicClient,
+) -> Generator[Namespace, Any, Any]:
+    """Create a namespace with Kueue label enabled for notebook workloads"""
+    from utilities.infra import create_ns
+
+    namespace_name = request.param["name"]
+    add_kueue_label = request.param.get("add-kueue-label", True)
+
+    with create_ns(
+        admin_client=admin_client,
+        name=namespace_name,
+        unprivileged_client=unprivileged_client,
+        teardown=True,
+        add_kueue_label=add_kueue_label,
+        pytest_request=request,
+    ) as ns:
+        yield ns
+
+
+@pytest.fixture(scope="class")
+def patched_kueue_manager_config(
+    admin_client: DynamicClient,
+    request: FixtureRequest,
+) -> Generator[ConfigMap, Any, Any]:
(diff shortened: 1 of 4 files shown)


utilities/kueue_utils.py
```diff
@@ -1,5 +1,5 @@
 from contextlib import contextmanager
-from typing import Optional, Dict, Any, List, Generator
+from typing import Optional, Dict, Any, List, Generator, Union
```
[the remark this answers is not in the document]
jiridanek: @coderabbitai ^^^ explain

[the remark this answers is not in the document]
jstourac: thank you, updated

tests/workbenches/notebook-controller/test_kueue_notebook_spawning.py
```diff
@@ -0,0 +1,695 @@
...
+        assert notebook_pod_labels.get("kueue.x-k8s.io/queue-name") == LOW_RESOURCE_LOCAL_QUEUE_NAME, (
+            "Pod should reference correct queue"
+        )
+
+        # Define pod labels for the check function
+        pod_labels = [f"app={default_notebook.name}"]
+
+        # Wait a bit to ensure the pod state stabilizes
+        time.sleep(10)  # noqa: FCN001
+
```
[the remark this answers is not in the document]
jstourac: also updated

jstourac:
@coderabbitai full review

tests/workbenches/notebook-controller/conftest.py
```diff
@@ -0,0 +1,282 @@
...
+    from utilities.infra import create_ns
+
+    namespace_name = request.param["name"]
+    add_kueue_label = request.param.get("add-kueue-label", True)
+
+    with create_ns(
+        admin_client=admin_client,
+        name=namespace_name,
+        unprivileged_client=unprivileged_client,
+        teardown=True,
```
dbasunag, line 143: default is true, you don't need to pass it.

tests/workbenches/notebook-controller/conftest.py
```diff
@@ -0,0 +1,282 @@
...
+        size="10Gi",
+        volume_mode=PersistentVolumeClaim.VolumeMode.FILE,
+    ) as pvc:
+        yield pvc
+
+
+@pytest.fixture(scope="class")
+def kueue_enabled_notebook_namespace(
+    request: FixtureRequest,
+    admin_client: DynamicClient,
```
dbasunag, line 130: admin client is only used if unprivileged client is not passed. So there is no need to pass both. If you want to use unprivileged client, then only pass it, else pass admin client.

tests/workbenches/notebook-controller/conftest.py
```diff
@@ -0,0 +1,282 @@
...
+    unprivileged_client: DynamicClient,
+) -> Generator[Namespace, Any, Any]:
+    """Create a namespace with Kueue label enabled for notebook workloads"""
+    from utilities.infra import create_ns
```
dbasunag, line 134: Please import  at the top of the file.

tests/workbenches/notebook-controller/conftest.py
```diff
@@ -0,0 +1,282 @@
...
+    """
+    # Detect Kueue installation scenario
+    scenario = detect_kueue_installation_scenario(client=admin_client)
+    LOGGER.info(f"Detected Kueue installation scenario: {scenario}")
+
+    # Check if we should patch config for this scenario
+    if not should_patch_kueue_config(scenario):
+        LOGGER.info(f"Skipping kueue-manager-config patching for scenario: {scenario}")
+        yield None
+        return
```
dbasunag, line 172: why both yield and return are needed?

tests/workbenches/notebook-controller/conftest.py
```diff
@@ -0,0 +1,282 @@
...
+
+    namespace = py_config["applications_namespace"]
```
dbasunag, line 174: Please use else.

dbasunag, line 174: no need to save. Use directly.

tests/workbenches/notebook-controller/conftest.py
```diff
@@ -0,0 +1,282 @@
...
+    config_map_name = "kueue-manager-config"
```
dbasunag, line 175: ditto.

tests/workbenches/notebook-controller/conftest.py
```diff
@@ -0,0 +1,282 @@
...
+            client=admin_client,
+            name=config_map_name,
+            namespace=namespace,
+            ensure_exists=True,
+        )
+    except Exception as e:
+        LOGGER.warning(f"Could not find kueue-manager-config ConfigMap: {e}")
+        LOGGER.info("This is expected for Red Hat Kueue operator scenario")
+        yield None
+        return
```
dbasunag, line 189: same comment.

tests/workbenches/notebook-controller/conftest.py
```diff
@@ -0,0 +1,282 @@
...
+
+    # Store original data and annotations for restoration
+    original_data = config_map.instance.data.copy() if config_map.instance.data else {}
```
dbasunag, line 192: each time `.instance` would make api call. Do you need that?

tests/workbenches/notebook-controller/conftest.py
```diff
@@ -0,0 +1,282 @@
...
+    original_annotations = (
+        config_map.instance.metadata.annotations.copy() if config_map.instance.metadata.annotations else {}
+    )
+
+    LOGGER.info("Storing original kueue-manager-config data for restoration")
```
dbasunag, line 197: Can you store everything in one fixture? storing config_map.instance.to_dict() will store everything

tests/workbenches/notebook-controller/conftest.py
```diff
@@ -0,0 +1,282 @@
...
+
+    # Get current config data
+    current_data = config_map.instance.data or {}
+    config_yaml = current_data.get("controller_manager_config.yaml", "{}")
+
+    # Parse the existing configuration
+    try:
+        config_dict = yaml.safe_load(config_yaml) or {}
+    except yaml.YAMLError:
+        config_dict = {}
```
dbasunag, line 207: When can this happen? Should not we fail then?

tests/workbenches/notebook-controller/conftest.py
```diff
@@ -0,0 +1,282 @@
...
+
+    # Ensure integrations section exists
```
dbasunag, line 209: Can you please update the comment? We are not ensuring it exists, we are basically padding it with missing information.

tests/workbenches/notebook-controller/conftest.py
```diff
@@ -0,0 +1,282 @@
...
+    # Add pod and statefulset if not already present
+    frameworks = config_dict["integrations"]["frameworks"]
+    if "pod" not in frameworks:
+        frameworks.append("pod")
+    if "statefulset" not in frameworks:
+        frameworks.append("statefulset")
+
+    # Convert back to YAML
+    updated_config_yaml = yaml.dump(config_dict, default_flow_style=False)
+    updated_data = {**current_data, "controller_manager_config.yaml": updated_config_yaml}
```
dbasunag, line 225: May be upto this line it should be an utility function?

tests/workbenches/notebook-controller/conftest.py
```diff
@@ -0,0 +1,282 @@
...
+
+    # Apply the patch with both data and metadata annotations
+    patch = {"metadata": {"annotations": {Annotations.OpenDataHubIo.MANAGED: "false"}}, "data": updated_data}
+
+    def restart_kueue_deployment(reason: str):
```
dbasunag, line 230: Do you need this to be internal function?

tests/workbenches/notebook-controller/conftest.py
```diff
@@ -0,0 +1,282 @@
...
+                client=admin_client,
+                name="kueue-controller-manager",
+                namespace=namespace,
+                ensure_exists=True,
+            )
+
+            # Get current replica count before restart
+            current_replicas = kueue_deployment.replicas
+            if current_replicas is None:
+                current_replicas = 1
```
dbasunag, line 249: Is this possible?

tests/workbenches/notebook-controller/test_kueue_notebook_spawning.py
```diff
@@ -0,0 +1,710 @@
...
+            flavor_name=RESOURCE_FLAVOR_NAME,
+            expected_cpu=expected_cpu,
+            expected_memory=expected_memory,
+        )
+
+        assert cluster_tracks_exact or local_tracks_exact, (
+            f"Either ClusterQueue or LocalQueue should show exact resource usage matching notebook requests "
+            f"(CPU: {expected_cpu}, Memory: {expected_memory}). "
+            f"ClusterQueue usage: {cluster_queue_usage}, LocalQueue usage: {local_queue_usage}"
+        )
```
dbasunag, line 229: Can you please use utilities? Also, the test seems to be too big, is it possible to split it?

tests/workbenches/notebook-controller/conftest.py
jstourac: updated

tests/workbenches/notebook-controller/conftest.py
jstourac: Why? Isn't this way better readable? Am I missing something?

tests/workbenches/notebook-controller/conftest.py
jstourac: intentional

tests/workbenches/notebook-controller/conftest.py
jstourac: let's keep it separate - we're saving only what we need to amend for the patch itself - we are also avoiding immutable fields this way

tests/workbenches/notebook-controller/conftest.py
jstourac: true, updated

tests/workbenches/notebook-controller/conftest.py
jstourac: hehe :slightly_smiling_face: TBH, I added it here during my debugging as sometimes it happened that the deployment didn't spin up back for me - maybe some timing issues or anything... this was to be sure that we are always scaling back to 1

tests/workbenches/notebook-controller/conftest.py
```diff
@@ -0,0 +1,281 @@
...
+        name=namespace_name,
+        unprivileged_client=unprivileged_client,
+        add_kueue_label=add_kueue_label,
+        pytest_request=request,
+    ) as ns:
+        yield ns
+
+
+@pytest.fixture(scope="class")
+def patched_kueue_manager_config(
```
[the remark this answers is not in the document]
jstourac: updated

tests/workbenches/notebook-controller/conftest.py
```diff
@@ -0,0 +1,281 @@
...
+    try:
+        config_map = ConfigMap(
+            client=admin_client,
+            name=config_map_name,
+            namespace=namespace,
+            ensure_exists=True,
+        )
+    except Exception as e:
+        LOGGER.warning(f"Could not find kueue-manager-config ConfigMap: {e}")
+        LOGGER.info("This is expected for Red Hat build of Kueue operator scenario")
```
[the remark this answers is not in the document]
jstourac: updated - I removed the info line and the first one is now changed to error and also raise an exception in this case as with current implementation, this shouldn't happen anymore

PR #513, opened 2025-08-11
update DEVELOPER_GUIDE.md to ensure the requirement of adding resources in wrapper librabry

dbasunag:
## Merge criteria:

### DIFF
--- a/docs/DEVELOPER_GUIDE.md
+++ b/docs/DEVELOPER_GUIDE.md
@@ -66,6 +66,12 @@ The project utilizes [openshift-python-wrapper](https://github.com/RedHatQE/open
 Please refer to the [documentation](https://github.com/RedHatQE/openshift-python-wrapper/blob/main/README.md)  
 and the [examples](https://github.com/RedHatQE/openshift-python-wrapper/tree/main/examples) for more information.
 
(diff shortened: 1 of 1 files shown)


adolfo-ab (approved):

jiridanek (changes requested):
I'm not a fan, I believe that if it is an important requirement, it should be managed the way @russellb did it in vllm in

* https://github.com/vllm-project/vllm/pull/18018

Hopefully a helpful chatbot, especially when given the vllm code as an example, can generate the pre-commit check with not too much effort.

I believe that documenting this will only help in disciplining the offenders (RTFM you fool!) but it will not have preventative effect for newcomers.

Combination of automated check that references the contributing documentation when failing, is imo the way.

The code to forbid is something like

```python
        api_version = "imageregistry.operator.openshift.io/v1"
        kind = "Config"

        resource = admin_client.resources.get(api_version=api_version, kind=kind)
```

taken from https://github.com/opendatahub-io/opendatahub-tests/commit/6a9dc1e09e4f7a20e73a620696f9b1a4b5c8776d#diff-67c9d3198e7209553e9352e8d116d28a2c1008331bf7e391b3acfc26707a98eeR18-R50

PR #569, opened 2025-08-29
Fix flaky snapshot tests by normalizing outputs

rpancham:
Addresses:[RHOAIENG-32598](https://issues.redhat.com/browse/RHOAIENG-32598)

## Summary by CodeRabbit

* **Tests**
  * Added recursive output normalization to ignore volatile metadata and mask UUID-like strings for stable comparisons.
  * Normalizes nested structures and deterministically sorts list items to reduce flakiness.
  * Broadened validation to accept a wider range of data types when comparing snapshots.
  * Added numerous new model validation snapshots covering multiple model configurations and inference modes.

### DIFF
--- a/tests/model_serving/model_runtime/model_validation/__snapshots__/test_modelvalidation/TestVLLMModelCarRaw.test_oci_model_car_raw_openai_inference[DeepSeek-R1-Distill-Llama-8B-FP8-dynamic-raw].1.json
+++ b/tests/model_serving/model_runtime/model_validation/__snapshots__/test_modelvalidation/TestVLLMModelCarRaw.test_oci_model_car_raw_openai_inference[DeepSeek-R1-Distill-Llama-8B-FP8-dynamic-raw].1.json
@@ -1,0 +1,50 @@
+[
+  {
+    "finish_reason": "length",
+    "index": 0,
+    "logprobs": null,
+    "prompt_logprobs": null,
+    "stop_reason": null,
+    "text": " Let me think through this step by step.\n\nFirst, I know that renewable energy sources like solar, wind, and hydro are cleaner than fossil fuels. They produce less pollution and greenhouse gases, which helps protect the environment. But I need to elaborate on this.\n\nRenewable energy sources are sustainable because they don't deplete the Earth's resources as much as fossil fuels. For example, solar panels can be recycled, and wind turbines can be reused or repaired. This reduces waste and the need for"
+  },
+  {
+    "finish_reason": "length",
+    "index": 0,
+    "logprobs": null,
+    "prompt_logprobs": null,
+    "stop_reason": null,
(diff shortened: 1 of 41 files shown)


edwardquarm:
@rpancham thanks for working on this. This is probably a big ask, but can you run the test to generate __snapshots__ folder and commit to this same branch using your fix? Snapshots were deleted because it was flaky.
The command to run the test is
``
uv run pytest -vv tests/model_serving/model_runtime/model_validation/test_modelvalidation.py \
  --model_car_yaml_path=./tests/model_serving/model_runtime/model_validation/sample_modelcar_config.yaml \
--vllm-runtime-image=quay.io/modh/vllm@sha256:db766445a1e3455e1bf7d16b008f8946fcbe9f277377af7abb81ae358805e7e2 \
--supported-accelerator-type=Nvidia \
--registry-host=[registry.redhat.io](http://registry.redhat.io/) \
  --snapshot-update \
  --log-file=pytest-logs.log
``

rpancham:
@Raghul-M Yes, that’s correct - the snapshot normalization routine is designed to remove volatile metadata so the comparisons focus on stable fields, mostly just the "text" outputs. This helps keep the tests from failing due to non-deterministic data like timestamps, IDs, or UUIDs.

Regarding model updates, as @edwardquarm mentioned my assumption was also that we’re validating against a set of fixed modelcar images each release, as provided by Rob. If this list is stable, then regenerating consistent snapshots should work well. If there’s any change in the modelcar images or model outputs, though, those specific snapshot files would need to be updated accordingly. Please let me know if there’s any change in how models are shipped or if we should anticipate variations in output fields between releases.

Raghul-M:
> > @rpancham correct me if I’m wrong it’s normalizing the snapshot output at runtime and comparing it with the existing ones, right? Since the models differ each time, your workaround checks only the specific text field in the output JSON, not the volatile metadata. But the responses to the queries can change from model to model, if I’m not mistaken. So instead of generating snapshots for all models, we save a specific JSON file containing the text and compare it with the model output each time it runs. @edwardquarm your thoughts on this?
> 
> @Raghul-M are the models going to change each release? I thought we're getting a fixed list of modelcar images to validate from Rob.

@edwardquarm  Yes the Modelcar Images change every release.

Raghul-M:
> @Raghul-M Yes, that’s correct - the snapshot normalization routine is designed to remove volatile metadata so the comparisons focus on stable fields, mostly just the "text" outputs. This helps keep the tests from failing due to non-deterministic data like timestamps, IDs, or UUIDs.
> 
> Regarding model updates, as @edwardquarm mentioned my assumption was also that we’re validating against a set of fixed modelcar images each release, as provided by Rob. If this list is stable, then regenerating consistent snapshots should work well. If there’s any change in the modelcar images or model outputs, though, those specific snapshot files would need to be updated accordingly. Please let me know if there’s any change in how models are shipped or if we should anticipate variations in output fields between releases.

@rpancham As of now, the images  differ across releases, so storing a snapshot of the model output is not useful here . If its a fixed list of models it will work .

PR #588, opened 2025-09-05
feat: migrate TrustyAIService tests to KServeRaw

adolfo-ab:
Migrates all the existing tests related to TrustyAIService that used KServe Serverless to using KServeRaw instead.
KServeRaw is now the preferred deployment mode, but it is also a requirement to run the tests in Power/Z  environments, which don't support Serverless.

This PR also improves the check to verify that the model pods have been registered by the TrustyAIService. Previously it relied only on the status of the model deployment, but this was a bit brittle with KServeRaw, and sometimes the function reported back even though the old pods were still in "Terminating" state, so I added an additional check to make sure that the old pods are deleted.

This PR has been tested by running all the affected tests several times on a PSI cluster.

## Summary by CodeRabbit

- New Features
  - Added support for RAW deployment in TrustyAI model tests, including external routing and per-namespace CA bundle configuration.
- Bug Fixes
  - Improved deployment readiness checks by validating pod availability and running status, and aligning logger URL comparison to HTTPS for increased reliability.
- Tests
  - Updated test fixtures to provision RAW deployment configs and CA bundles.
  - Switched test markers from serverless to rawdeployment across drift, fairness, single- and multi-namespace suites.
  - Expanded multi-namespace setups to provision required resources and dependencies for RAW deployment.

### DIFF
--- a/tests/model_explainability/trustyai_service/service/multi_ns/test_trustyai_service_multi_ns.py
+++ b/tests/model_explainability/trustyai_service/service/multi_ns/test_trustyai_service_multi_ns.py
@@ -123,6 +123,7 @@ class TestTrustyAIServiceMultipleNS:
     indirect=True,
 )
 @pytest.mark.usefixtures("minio_pod")
+@pytest.mark.rawdeployment
 class TestDriftMetricsWithDBStorageMultiNs:
     """
     Verifies drift metric functionality in TrustyAI using MariaDB storage


sheltoncyril (approved):
/lgtm

kpunwatk (approved):
/lgtm

dbasunag (approved):

PR #624, opened 2025-09-18
fix: backport rag test folder location to 2.24

jgarciao:
The location of RAG tests is in tests/llama_stack/rag in the main branch, and jenkins was updated to use that path
This PR is backporting this folder location to the 2.24 branch so the tests run properly for RHOAI 2.24
I didn't move the lm_eval tests because I was not 100% sure of the changes required (to be done in another PR)


kpunwatk (approved):
/lgtm

PR #664, opened 2025-10-01
2.16 fix for managed state throws error 

mwaykole:
**

## Merge criteria:

### DIFF
--- a/tests/model_registry/conftest.py
+++ b/tests/model_registry/conftest.py
@@ -186,7 +186,7 @@ def model_registry_db_deployment(
                             "/var/lib/mysql/datadir",
                             "--default-authentication-plugin=mysql_native_password",
                         ],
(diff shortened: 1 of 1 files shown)


kpunwatk (approved):
/lgtm

utilities/infra.py
```diff
@@ -362,7 +362,9 @@ def is_managed_cluster(client: DynamicClient) -> bool:
     for entries in platform_statuses.values():
         if isinstance(entries, kubernetes.dynamic.resource.ResourceField):
             if tags := entries.resourceTags:
-                return next(b["value"] == "true" for b in tags if b["key"] == "red-hat-managed")
```
dbasunag, line 365: Is the problem because b is `None`? Then would `return next(b["value"] == "true" for b in tags if b and b["key"] == "red-hat-managed")` help?

PR #675, opened 2025-10-03
Use random name for RAG namespace

jiripetrlik:
Based on https://issues.redhat.com/browse/RHAIENG-999 it will be good to use random namespace name.

## Summary by CodeRabbit

* **Tests**
  * Test run identifiers are now randomized to ensure unique names per run, improving isolation and repeatability.
  * Reduces naming collisions and flakiness in parallel and CI environments.
  * Improves overall test reliability while leaving user-facing behavior, features, and performance unchanged.

### DIFF
--- a/tests/llama_stack/rag/test_rag.py
+++ b/tests/llama_stack/rag/test_rag.py
@@ -9,6 +9,7 @@ from simple_logger.logger import get_logger
 from utilities.rag_utils import validate_rag_agent_responses, validate_api_responses, ModelInfo
 
 from tests.llama_stack.utils import get_torchtune_test_expectations, create_response_function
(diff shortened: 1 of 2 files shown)


tests/llama_stack/rag/test_rag.py
```diff
@@ -17,7 +18,7 @@
     "unprivileged_model_namespace, llama_stack_server_config",
```
jgarciao, line 18: How about adding a boolean fixture param `randomize_name` to the fixture `unprivileged_model_namespace` and add the random part there?

jiripetrlik:
@jgarciao 
Thank you! I've tried to address your comment.

jgarciao (approved):

PR #1132, opened 2026-02-19
feat: add remote garak  integration tests

kpunwatk:
## Merge criteria:

## Summary by CodeRabbit

* **New Features**
  * Added Garak as a selectable remote evaluation provider with provider-aware defaults and conditional Kubeflow activation.
  * Allow in-cluster requests to reach LlamaStack deployments.

* **Tests**
  * Added end-to-end tests for the Garak remote evaluation provider (quick, custom benchmark, and shield scan flows).
  * Improved test helpers: more robust job polling, clearer failure reporting, and result/metadata validation utilities.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -75,6 +75,7 @@ dependencies = [
     "dictdiffer>=0.9.0",
     "pytest>=9.0.0",
     "huggingface-hub>=1.2.3",
+    "httpx[http2]>=0.28.1",
 ]
 
 [project.urls]
--- a/tests/llama_stack/conftest.py
+++ b/tests/llama_stack/conftest.py
@@ -283,8 +283,9 @@ def llama_stack_server_config(
 
         # KUBEFLOW_BASE_IMAGE: Select based on provider
         # Garak uses KUBEFLOW_GARAK_BASE_IMAGE; Ragas uses KUBEFLOW_BASE_IMAGE
-        # quay.io/rhoai/odh-trustyai-garak-lls-provider-dsp-rhel9@sha256:75eb795e9e459c0f6951ee1fc3ee325ae593d6aab32eee203723d28880c7ca31
-        default_garak_image = "quay.io/opendatahub/odh-trustyai-garak-lls-provider-dsp@sha256:a3b65a9fdb6996fdaac45286b17522806cdf5af133275806fef5f93265103fc9"
+        # quay.io/rhoai/odh-trustyai-garak-lls-provider-dsp-rhel9@sha256:75eb795e9e459c0f6951ee1fc3ee325ae593d6aab32eee203723d28880c7ca31 (3.4-ea.1 sha)
+        # quay.io/opendatahub/odh-trustyai-garak-lls-provider-dsp@sha256:a3b65a9fdb6996fdaac45286b17522806cdf5af133275806fef5f93265103fc9
+        default_garak_image = "quay.io/rhoai/odh-trustyai-garak-lls-provider-dsp-rhel9:rhoai-3.4-ea.1"
         default_ragas_image = "quay.io/diegosquayorg/my-ragas-provider-image@sha256:3749096c47f7536d6be2a7932e691abebacd578bafbe65bad2f7db475e2b93fb"
 
         selected_image = params.get("kubeflow_base_image")
--- a/tests/llama_stack/conftest.py
+++ b/tests/llama_stack/conftest.py
@@ -255,45 +255,50 @@ def llama_stack_server_config(
         env_vars.append({"name": "TRUSTYAI_EMBEDDING_MODEL", "value": trustyai_embedding_model})
 
     # Kubeflow-related environment variables
-    if params.get("enable_ragas_remote"):
-        # Get fixtures only when Ragas Remote/Kubeflow is enabled
+    enable_ragas_remote = params.get("enable_ragas_remote", False)
+    enable_garak_remote = params.get("enable_garak_remote", False)
+    enable_kubeflow_eval = enable_ragas_remote or enable_garak_remote
+
+    if enable_kubeflow_eval:
+        # Get fixtures only when Kubeflow is enabled
         model_namespace = request.getfixturevalue(argname="model_namespace")
         current_client_token = request.getfixturevalue(argname="current_client_token")
         dspa_route = request.getfixturevalue(argname="dspa_route")
         dspa_s3_secret = request.getfixturevalue(argname="dspa_s3_secret")
 
-        # KUBEFLOW_LLAMA_STACK_URL: Build from LlamaStackDistribution service
+        # New Garak-specific flag
+        env_vars.append({"name": "ENABLE_KUBEFLOW_GARAK", "value": str(enable_garak_remote).lower()})
+
+        # KUBEFLOW_LLAMA_STACK_URL: Use internal service endpoint for KFP to callback to
+        distribution_name = params.get("distribution_name", "rh-dev")
         env_vars.append({
             "name": "KUBEFLOW_LLAMA_STACK_URL",
             "value": f"http://{distribution_name}-service.{model_namespace.name}.svc.cluster.local:8321",
         })
 
-        # KUBEFLOW_PIPELINES_ENDPOINT: Get from DSPA route
+        # KUBEFLOW_PIPELINES_ENDPOINT: Injected from DSPA
         env_vars.append({"name": "KUBEFLOW_PIPELINES_ENDPOINT", "value": f"https://{dspa_route.instance.spec.host}"})
-
-        # KUBEFLOW_NAMESPACE: Use model namespace
         env_vars.append({"name": "KUBEFLOW_NAMESPACE", "value": model_namespace.name})
 
-        # KUBEFLOW_BASE_IMAGE
-        env_vars.append({
-            "name": "KUBEFLOW_BASE_IMAGE",
-            "value": params.get(
-                "kubeflow_base_image",
-                "quay.io/diegosquayorg/my-ragas-provider-image"
-                "@sha256:3749096c47f7536d6be2a7932e691abebacd578bafbe65bad2f7db475e2b93fb",
-            ),
-        })
+        # KUBEFLOW_BASE_IMAGE: Select based on provider
+        default_garak_image = "quay.io/trustyai/garak-remote-provider:latest"
+        default_ragas_image = "quay.io/diegosquayorg/my-ragas-provider-image@sha256:3749096c47f7536d6be2a7932e691abebacd578bafbe65bad2f7db475e2b93fb"
 
-        # KUBEFLOW_RESULTS_S3_PREFIX: Build from MinIO bucket
+        selected_image = params.get("kubeflow_base_image")
+        if not selected_image:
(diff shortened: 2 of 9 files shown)


mwaykole (commented):
## PR Review: `feat: add remote garak integration tests`

**Reviewer:** Milind Waykole (autonomous dev-engineer agent)
**Risk Level:** 🟡 MEDIUM
**Overall Assessment:** Good foundational work with real architectural thinking. Several correctness and robustness issues need addressing before merge.

---

## Summary

This PR adds end-to-end integration tests for the TrustAI Garak remote (KFP-backed) evaluation provider across three tiers: quick scan (smoke), custom benchmark (tier1), and shield scan (tier2). It also expands the Kubeflow fixture to be provider-aware (Ragas vs. Garak), adds in-cluster service URL fixtures, and introduces the `httpx[http2]` dependency. I read the full diff plus the `llama-stack-provider-trustyai-garak` architecture doc (rhoai-3.4-ea.2) to ground the review.

---

## 🔴 Must-Fix Issues

### 1. `_cr_name` side-channel through `server_config` dict mutation is fragile

**File:** `tests/llama_stack/conftest.py` L347-350 and L411, L460

```python
# In llama_stack_server_config fixture:
if enable_garak_remote or params.get("enable_kubeflow_ragas"):
    server_config["_cr_name"] = cr_name   # underscore key injected into dict

# In llama_stack_distribution / unprivileged_llama_stack_distribution:
distribution_name = llama_stack_server_config.pop("_cr_name", None) or generate_random_name(...)
```

**Problem:** `cr_name` is only generated when `enable_kubeflow_eval` is `True` (i.e., inside the `if enable_kubeflow_eval:` block, L268). But `_cr_name` is set outside that block at L347, gated on `enable_garak_remote or params.get("enable_kubeflow_ragas")`. If `enable_garak_remote=True` but the fixtures resolve without entering the `enable_kubeflow_eval` block (e.g., a new future code path), `cr_name` will be a `NameError` at runtime. The variable is only defined in one scope but consumed in another.

**Fix:** Move `cr_name = generate_random_name(prefix="llama-stack-distribution")` to the top of the fixture, unconditionally. The cost is one extra `generate_random_name` call even when Kubeflow is disabled — that is completely acceptable.

Also, using a private underscore key to pass data through the config dict is an opaque contract. Consider a named dataclass or a dedicated fixture `llama_stack_cr_name` to make the dependency explicit.

---

### 2. `qwen_isvc_service_url` fixture: missing wait / no readiness guard

**File:** `tests/llama_stack/eval/conftest.py` L17-37

```python
svc_list = list(Service.get(client=admin_client, namespace=qwen_isvc.namespace, name=svc_name))
if not svc_list:
    raise RuntimeError(f"Service {svc_name} not found in namespace {qwen_isvc.namespace}")
port = svc.instance.spec.ports[0].port
```

**Problem 1:** This is a point-in-time lookup with no polling. The KServe predictor service may not exist yet when the fixture runs (the ISVC might still be reconciling). Unlike the `guardrails_orchestrator_service_url` fixture added in the same PR (which correctly uses `TimeoutSampler`), this fixture will raise immediately if the service isn't ready.

**Problem 2:** `svc.instance.spec.ports[0].port` will raise `IndexError` if the service exists but has no ports (pathological but possible during operator reconciliation).

**Fix:** Wrap the lookup in a `TimeoutSampler` loop (same pattern as `guardrails_orchestrator_service_url`), and guard the port access:
```python
if not svc.instance.spec.ports:
    return None  # retry
port = svc.instance.spec.ports[0].port
```

---

### 3. `guardrails_orchestrator_service_url`: port 443 included in HTTP-ports list causes silent TLS mismatch

**File:** `tests/llama_stack/eval/conftest.py` L248-265

```python
_HTTP_PORTS = (8033, 8034, 80, 443)
# ...
return f"http://{svc.name}.{ns}.svc.cluster.local:{http_port}"
```

**Problem:** Port 443 is in `_HTTP_PORTS` but the returned URL scheme is always `http://`. If the orchestrator service exposes port 443 (which it can for HTTPS), the test will construct `http://…:443` — an HTTP request to an HTTPS port. This will either timeout or get a TLS handshake error that surfaces as a confusing connection reset, not a clear test failure.

The architecture doc for FMS Guardrails Orchestrator confirms it typically exposes 8033 (HTTP) and 8034 (HTTPS). Port 443 inclusion in this list is risky.

**Fix:** Remove 443 from the HTTP ports tuple. If TLS is ever needed, handle it explicitly with a separate scheme selection:
```python
_HTTP_PORTS = (8033, 8034, 80)
```

---

### 4. `RuntimeError` in `wait_for_eval_job_completion` leaks job metadata into exception messages

**File:** `tests/llama_stack/eval/utils.py` L55-60

```python
raise RuntimeError(
    f"Eval job {job_id} for benchmark {benchmark_id} terminated with "
    f"status: {sample}. Metadata: {metadata}"
)
```

**Problem:** `metadata` from a failed KFP job can contain execution context, internal URLs, or timing data that may be sensitive. More critically, `getattr(final_job, "metadata", {})` returns `{}` as default, but if the LlamaStack client returns a Pydantic model rather than a plain dict, `{}" as the default will silently hide the real metadata object (the `getattr` default only fires if the attribute is absent entirely). Add `or {}` after the `getattr` for safety.

**Also:** CodeRabbit flagged this as CWE-532 (Information Exposure Through Log Files). While test infra is lower stakes than production, tokens and run IDs in exception messages can appear in CI logs. Sanitize `metadata` before including in the message (e.g., log it at DEBUG and omit from the raised exception).

---

## 🟡 Should-Fix Issues

### 5. Shared mutable class-level `_job_id` is not thread-safe and breaks test isolation between parametrize runs

**File:** `tests/llama_stack/eval/test_garak_provider.py`

```python
class TestGarakRemoteQuickScan:
    _job_id: str = ""
    # ...
    TestGarakRemoteQuickScan._job_id = job.job_id  # set on class, not instance
```

Each test class uses a class-level `_job_id` string to pass the job ID from `test_run_*` to `test_*_results`. This works because `pytest.mark.dependency` serializes the tests within a class. However:

1. If pytest ever runs parametrized variants in parallel (xdist), the class attribute is a shared race condition.
2. The pattern ties test state to class-level mutation, which is a pytest anti-pattern. The standard approach is a `pytest.fixture` with `scope="class"` that yields a mutable container (e.g., `{"job_id": None}`).

**Recommendation:** At minimum, add a comment documenting why class-level state is used and that xdist is not supported. Ideally, refactor to use a class-scoped fixture.

---

### 6. Benchmark re-registration pattern is duplicated three times — extract to a helper

**File:** `tests/llama_stack/eval/test_garak_provider.py`

The try/except/unregister/re-register pattern for idempotent benchmark registration appears identically in `TestGarakRemoteQuickScan`, `TestGarakRemoteCustomBenchmark`, and `TestGarakRemoteShieldScan`. This is ~20 lines repeated three times. It should be extracted to a utility function in `tests/llama_stack/eval/utils.py`:

```python
def register_benchmark_idempotent(client, benchmark_id, dataset_id, scoring_functions, provider_id, metadata, provider_benchmark_id=None):
    """Register a benchmark, unregistering first if it already exists."""
    ...
```

Same for the shield re-registration pattern.

---

### 7. `distribution` field override logic inconsistency: `enable_kubeflow_ragas` vs `enable_ragas_remote`

**File:** `tests/llama_stack/conftest.py` L367 and L371

```python
if enable_garak_remote or params.get("enable_kubeflow_ragas"):
    server_config["_cr_name"] = cr_name
```

The existing Ragas path uses `params.get("enable_ragas_remote")` everywhere else in this file, but here the code checks `params.get("enable_kubeflow_ragas")` — a different key name that doesn't appear to be set anywhere in the existing tests. This looks like a leftover from a rename. Please verify and use the canonical key consistently.

---

### 8. `httpx[http2]` dependency: HTTP/2 is not needed for in-cluster svc:// requests

**File:** `pyproject.toml` L78

```toml
"httpx[http2]>=0.28.1",
```

The PR description and commit messages don't explain why HTTP/2 is required. In-cluster Kubernetes service communication uses HTTP/1.1 by default. HTTP/2 requires the `h2`, `hpack`, and `hyperframe` transitive packages (added to uv.lock), expanding the attack surface unnecessarily.

If HTTP/2 is not actually used in the test code (no `http2=True` in any `httpx.Client` call), use plain `httpx>=0.28.1`. If it IS needed (e.g., for gRPC-over-HTTP2 or specific LlamaStack endpoints), document why in a comment.

---

## 🟢 Positive Observations

1. **Architecture alignment is correct.** The `KUBEFLOW_LLAMA_STACK_URL` pre-generation pattern (generating `cr_name` before the LlamaStack distribution is created so KFP pods can find it) directly addresses the real-world ordering problem described in the Garak architecture doc (Flow 2, Step 3: KFP pods call back to `http://{lsd-name}-service:8321`). This is the right fix.

2. **Three-tier test structure is well-reasoned.** Smoke / tier1 / tier2 separation maps cleanly to the Garak provider's three feature surfaces (quick benchmark, custom benchmark, shield scan), making CI triaging straightforward.

3. **Provider-aware image selection** (`KUBEFLOW_GARAK_BASE_IMAGE` vs `KUBEFLOW_BASE_IMAGE`) is clean and backward-compatible with existing Ragas tests.

4. **`validate_eval_result_structure` and `validate_job_metadata` utilities** are well-factored and add real value for future tests building on this provider.

5. **Network policy expansion** in `tests/llama_stack/utils.py` (adding `namespace` to `allowedFrom`) is architecturally correct per the Garak architecture doc (Network Policy: `allow-kfp-to-llamastack`).

6. **`guardrails_orchestrator_service_url` using `TimeoutSampler`** with 120s/5s is the right pattern for async operator-created services.

---

## 📋 PR Hygiene

- **PR description is empty** — the template placeholders are unfilled. The CodeRabbit pre-merge check already flagged this. Please add: what was changed, how it was tested (environment, cluster version, RHOAI version), and check off the merge criteria. The commit messages from `saichandrapandraju` are detailed; a summary of those belongs in the PR body.
- **WIP label is still set** — remove before requesting final review.
- **Merge conflict** — CodeRabbit reports a conflict in the `garak_inline` branch. Needs rebase.
- **Multiple authors** — commits from `kpunwatk` and `saichandrapandraju` with a merge commit (`58d6c04`). Consider squashing for a clean history per the merge criteria checklist.

---

## Risk Classification

| Area | Risk | Reason |
|------|------|--------|
| `llama_stack_server_config` fixture mutation | 🔴 HIGH | `cr_name` variable scope + `_cr_name` dict key is implicit contract |
| `qwen_isvc_service_url` no polling | 🔴 HIGH | Will flake on slow clusters |
| Port 443 in HTTP URLs | 🔴 HIGH | Silent TLS mismatch |
| `RuntimeError` metadata exposure | 🟡 MEDIUM | CWE-532, CI log exposure |
| Class-level `_job_id` | 🟡 MEDIUM | xdist incompatible |
| Duplicated re-registration pattern | 🟡 MEDIUM | Maintenance burden |
| `httpx[http2]` justification | 🟡 MEDIUM | Unexplained dependency expansion |
| PR description empty | 🟡 MEDIUM | Process compliance |

---

Please address the 🔴 must-fix items before requesting re-review. Happy to discuss the `_cr_name` refactor approach if you want to explore fixture-based alternatives.

PR #1391, opened 2026-04-13
fix: modify PDF used for vector_stores test for FIPS compatibility

jgarciao:
Modify PDF used in RHOAI 3.3's vector stores tests to be compatible with FIPS clusters


Bobbins228 (commented):
/lgtm thanks Jorge

Bobbins228 (approved):
/lgtm

PR #1394, opened 2026-04-13
feat: add RAGAS evaluation tests for RAG pipeline

Bobbins228:
# Pull Request

## Summary

Refactor RAG inference and RAGAS evaluation into a dedicated `TestLlamaStackRagasEval` class with tier-based markers. This separates RAGAS evaluation tests from the core vector store functional tests, adds configurable sample limits via the `RAGAS_MAX_SAMPLES` env var (default 5), and moves shared RAGAS fixtures and helpers to `utils.py` for reuse.

Running `-m tier1` executes only the faithfulness metric; `-m tier2` runs all four metrics (faithfulness, answer relevancy, context precision, context recall) without marker bleed-through from other test classes.

## Related Issues

- Fixes: 
- JIRA: RHAIENG-3229

## Changes

- **`tests/llama_stack/vector_io/test_vector_stores.py`** — Add `TestLlamaStackRagasEval` class with four RAGAS metric tests (`faithfulness`, `answer_relevancy`, `context_precision`, `context_recall`), each with tier1/tier2 markers and threshold assertions
- **`tests/llama_stack/utils.py`** — Add `ragas_samples`, `ragas_evaluator_llm`, and `ragas_evaluator_embeddings` fixtures; add `extract_retrieved_contexts` and `mean_ragas_score` helpers
- **`tests/llama_stack/constants.py`** — Add RAGAS threshold constants and configurable `RAGAS_MAX_SAMPLES` (env var, default 5)
- **`tests/llama_stack/conftest.py`** — Import RAGAS fixtures from `utils.py` for pytest discovery

- [x] Locally
- [ ] Jenkins

## Additional Requirements

- [x] If this PR introduces new marker(s)/adds a new component, was relevant ticket created to update relevant Jenkins job?

## Summary by CodeRabbit

* **Tests**
  * Added RAG evaluation tests covering faithfulness, answer relevancy, context precision, and context recall with configurable thresholds.
  * Added end-to-end evaluation flows that build sample sets from vector stores, extract retrieved contexts, compute per-sample and aggregate metric scores, and assert thresholds.

* **Chores**
  * Added evaluation tooling/dependencies and a configurable max-samples setting for test runs.


tests/llama_stack/vector_io/test_vector_stores.py
```diff
@@ -260,3 +273,124 @@ def test_response_file_search_tool_invocation(
...
+    pipeline using RAGAS metrics. This class runs end-to-end evaluations over the Responses
+    API with vector store-backed file search, computing faithfulness, answer relevancy,
+    context precision, and context recall. It ensures that LlamaStack's RAG responses
+    are factual, relevant, and grounded in the retrieved context, using multiple metrics
+    and data samples. The test configures model and provider parameters via pytest,
+    and leverages fixtures to build the RAGAS evaluation dataset.
+    """
+
+    @pytest.mark.tier1
+    @pytest.mark.tier2
```
dbasunag: Tests should be under separate gating brackets.
``` 
tier1: Mark tests as tier1. High-priority tests.
tier2: Mark tests as tier2. Medium/low-priority positive tests.
tier3: Mark tests as tier3. Negative and destructive tests.
```
One test should never be part of both smoke, tier1 and tier2

tests/llama_stack/vector_io/test_vector_stores.py
```diff
@@ -260,3 +273,124 @@ def test_response_file_search_tool_invocation(
...
+        """
+        result = evaluate(
+            dataset=EvaluationDataset(samples=ragas_samples),
+            metrics=[AnswerRelevancy(llm=ragas_evaluator_llm, embeddings=ragas_evaluator_embeddings)],
+        )
+        score = mean_ragas_score(scores=result["answer_relevancy"])
+        LOGGER.info(f"RAGAS Answer Relevancy score: {score:.3f} (threshold: {ANSWER_RELEVANCY_THRESHOLD})")
+        assert score >= ANSWER_RELEVANCY_THRESHOLD, (
+            f"RAGAS Answer Relevancy score {score:.3f} is below threshold {ANSWER_RELEVANCY_THRESHOLD}"
+        )
```
dbasunag: Looks like both test_ragas_faithfulness, test_ragas_context_precision, test_ragas_answer_relevancy and test_ragas_context_recall can be parameterized. 
You can pass the value of metrics param of your evaluate call using pytest.param and use distinct ids for the test and the value you are comparing your score with can be passed to the pytest.param too. The test body is very similar across all 4 tests and at least 3 of them uses the same exact fixtures as well.

tests/llama_stack/utils.py
```diff
@@ -391,3 +401,156 @@ def vector_store_upload_dataset(
...
+    return retrieved_contexts
+
+
+def mean_ragas_score(scores: list[float | None]) -> float:
+    """Compute mean of RAGAS per-sample scores, filtering out NaN values."""
+    valid = [s for s in scores if s is not None and not math.isnan(s)]
+    return sum(valid) / len(valid) if valid else 0.0
+
+
+@pytest.fixture(scope="class")
```
jgarciao, line 438: I think fixtures should be (by convention) in tests/llama_stack/conftest.py  or, if only used in vector stores, in tests/llama_stack/vector_io/conftest.py  (I think this is better)

tests/llama_stack/vector_io/test_vector_stores.py
jgarciao, line 1: Could  you add the new tests in a new file tests/llama_stack/vector_io/test_vector_stores_evaluation.py or similar?

jgarciao (changes requested):

tests/llama_stack/utils.py
```diff
@@ -398,3 +399,41 @@ def vector_store_upload_dataset(
...
+                    seen.add(text)
+                    retrieved_contexts.append(text)
+
+    return retrieved_contexts
+
+
+def mean_ragas_score(scores: list[float | None]) -> float:
+    """Compute mean of RAGAS per-sample scores, filtering out NaN values."""
+    valid = [s for s in scores if s is not None and not math.isnan(s)]
+    return sum(valid) / len(valid) if valid else 0.0
```
Ygnas: This function is returning 0.0 when all scores are `None` or `NaN`. When test fails later with message like `score 0.0 is below threshold 0.5`, person looking at CI will think RAG pipeline is giving bad answers, but actually RAGAS did not produce any scores at all. Maybe better to raise error/log here so its clear what happened. WDYT?

pyproject.toml
```diff
@@ -78,6 +83,9 @@ dependencies = [
     "dictdiffer>=0.9.0",
     "pytest>=9.0.0",
     "huggingface-hub>=1.2.3",
+    "openai>=1.0",
+    "langchain-openai>=0.3",
```
jgarciao: Could you specify a a recent version of openai?

See comment below about a way to avoid adding langchain-openai

tests/llama_stack/vector_io/test_vector_stores_evaluation.py
```diff
@@ -0,0 +1,118 @@
...
+
+@pytest.mark.parametrize(
+    "unprivileged_model_namespace, llama_stack_server_config, vector_store, dataset",
+    [
+        pytest.param(
+            {"name": "test-llamastack-ragas-eval", "randomize_name": True},
+            {
+                "llama_stack_storage_size": "2Gi",
+                "vector_io_provider": "pgvector",
+                "embedding_provider": "sentence-transformers",
```
jgarciao: The sentence-transformers provider is deprecated in 3.5 EA1.
Can you set `"embedding_provider": "vllm-embedding", ` ?

tests/llama_stack/vector_io/test_vector_stores_evaluation.py
```diff
@@ -0,0 +1,118 @@
...
+                "files_provider": "local",
```
jgarciao: The local provider is not deprecated yet but there are discussions about it. Can you use s3 instead?
`"files_provider": "s3",`

tests/llama_stack/vector_io/conftest.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+@pytest.fixture(scope="class")
+def ragas_evaluator_embeddings(
+    unprivileged_llama_stack_client: LlamaStackClient,
+    llama_stack_models: ModelInfo,
+) -> Generator[Any, Any, Any]:
+    """Create RAGAS embeddings backed by the Llama Stack sentence-transformers provider.
+
+    Uses langchain_openai.OpenAIEmbeddings because the old ragas.metrics.AnswerRelevancy
```
jgarciao: I've asked a chatbot to find a way not to use the langchain workaround and it proposed this:


Use RAGAS-native embeddings + adapter (recommended)

Wrap your embedding model with a thin adapter that provides the expected methods:
```
class RagasEmbeddingAdapter:
    def __init__(self, embeddings):
        self.embeddings = embeddings

    def embed_query(self, text: str):
        return self.embeddings.embed(text)

    def embed_documents(self, texts: list[str]):
        return [self.embeddings.embed(t) for t in texts]
```
Then:
```
from ragas.embeddings import OpenAIEmbeddings
from ragas.metrics.collections import AnswerRelevancy

emb = OpenAIEmbeddings(
    model=llama_stack_models.embedding_model.id,
    base_url=f"{base_url}/v1",
)

adapted_emb = RagasEmbeddingAdapter(emb)

metric = AnswerRelevancy(embeddings=adapted_emb)
```
Not sure if it will work, but it's worth trying to avoid adding a new dependency

tests/llama_stack/vector_io/conftest.py
```diff
@@ -0,0 +1,169 @@
...
+            client=openai_client,
+        )
+        evaluator_llm.model_args["max_tokens"] = 4096
+
+        yield evaluator_llm
+    finally:
+        http_client.close()
+
+
+class _LangChainCompatibleEmbeddings:
```
jgarciao: given that we no longer used LangChain, how about naming this OpenAIEmbeddingsAdapter or similar?

Dockerfile
```diff
@@ -13,7 +13,7 @@ ENV PATH="$PATH:$BIN_DIR"
 
 # Install system dependencies using dnf
 RUN dnf update -y \
-    && dnf install -y python3 python3-pip ssh gnupg curl gpg wget vim httpd-tools rsync openssl openssl-devel skopeo\
+    && dnf install -y python3 python3-pip python3-devel ssh gnupg curl gpg wget vim httpd-tools rsync openssl openssl-devel skopeo gcc-c++\
```
jgarciao, line 16: Are these changes still needed now that we don't use LangChain

pyproject.toml
```diff
@@ -78,6 +83,8 @@ dependencies = [
     "dictdiffer>=0.9.0",
     "pytest>=9.0.0",
     "huggingface-hub>=1.2.3",
+    "openai>=1.75",
+    "ragas>=0.2",
```
jgarciao: How about adding a recent version for openai>=2.36 and ragas>=0.4 ?

uv.lock
```diff
@@ -1211,6 +1340,182 @@ wheels = [
...
+    { name = "typing-extensions" },
+    { name = "uuid-utils" },
+]
+sdist = { url = "https://files.pythonhosted.org/packages/a0/d8/7bdf30e4bfc5175609201806e399506a0a78a48e14367dc8b776a9b4c89c/langchain_core-1.2.29.tar.gz", hash = "sha256:cfb89c92bca81ad083eafcdfe6ec40f9803c9abf7dd166d0f8a8de1d2de03ca6", size = 846121, upload-time = "2026-04-14T20:44:58.117Z" }
+wheels = [
+    { url = "https://files.pythonhosted.org/packages/72/37/fed31f80436b1d7bb222f1f2345300a77a88215416acf8d1cb7c8fda7388/langchain_core-1.2.29-py3-none-any.whl", hash = "sha256:11f02e57ee1c24e6e0e6577acbd35df77b205d4692a3df956b03b5389cbe44a0", size = 508733, upload-time = "2026-04-14T20:44:56.712Z" },
+]
+
+[[package]]
+name = "langchain-openai"
```
jgarciao, line 1418: Can you update the lock file?

jgarciao: Mark saw that it's a ragas dependency

jgarciao (approved):
LGTM. Thanks a lot, @Bobbins228 !

jiripetrlik (approved):

PR #1417, opened 2026-04-17
fix: replace htpasswd CLI dependency with bcrypt

mynhardtburger:
## Summary
- Replace the `htpasswd` CLI call in `create_htpasswd_file()` with the `bcrypt` Python library, producing `$2b$` bcrypt hashes
- Removes the need to have `httpd-tools` / `apache2-utils` installed in the test environment
- Drops the `shlex` and `pyhelper_utils.shell.run_command` imports that were only used for the `htpasswd` subprocess call
- Adds `bcrypt>=4.0.0` as a project dependency
- OpenShift's HTPasswd IDP documentation uses htpasswd -B (bcrypt mode) in its examples per [official docs](https://docs.redhat.com/en/documentation/openshift_container_platform/4.18/html/authentication_and_authorization/configuring-identity-providers)
- replace bcrypt $2b$ prefix with $2y$ for OpenShift OAuth compatibility

## Summary by CodeRabbit

* **Refactor**
  * Password file generation now hashes credentials internally using bcrypt for a simpler, more reliable workflow.

* **Chores**
  * Removed an unnecessary system package from the Docker build.
  * Added a bcrypt runtime dependency to support the new hashing approach.


dbasunag (commented):
Please update the Dockerfile if we no longer need the htpasswd cli, generate the image and test it via jenkins.
1) tests that uses this code should be tested
2) smoke should pass

dbasunag:
/hold

jenkins run against this PR failed the test_mr_rbac.py tests. Please uphold it, once this test passes against your pr.

dbasunag (dismissed):

fege (dismissed):
/lgtm

dbasunag (approved):

fege (approved):
/lgtm

PR #1451, opened 2026-04-24
test: add AutoML smoke test for pipelines-components

MRGuziX:
Add basic E2E smoke test for the AutoGluon Tabular Training pipeline from pipelines-components. The test deploys a DSPA with built-in MinIO, uploads synthetic training data, submits the pipeline via the DSPA REST API, and polls the Argo Workflow status until completion.

- New test area: tests/pipelines_components/
- Pipeline YAML bundled from red-hat-data-services/pipelines-components
- Polls Argo Workflow directly (avoids persistence agent sync delays)
- New pytest marker: pipelines_components

JIRA: RHOAIENG-59399

Assisted-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

# Pull Request

## Related Issues

- Fixes: 
- JIRA: 

- [ ] Locally
- [ ] Jenkins

## Additional Requirements

## Summary by CodeRabbit

* **New Features**
  * End-to-end AutoML smoke test and AutoGluon tabular training pipeline with S3 support, model refit and leaderboard generation.
  * AutoRAG smoke tests with external Llama Stack integration and optional vector DB support.
  * New workflow resource support for pipeline status inspection.

* **Documentation**
  * Added a smoke-testing guide and environment configuration examples (.env.example) covering setup and run commands.

* **Tests**
  * Added test fixtures, utilities, and smoke test suites to deploy pipelines, upload data, run pipelines, poll statuses, and collect logs.

AutoML test run locally with:
`uv run pytest tests/pipelines_components/automl/ -m smoke -v -s --junitxml=results_xunit.xml`
image

AutoRAG test run locally with:
`pytest tests/pipelines_components/autorag/test_autorag_smoke.py -m smoke -v `
image


tests/pipelines_components/automl/pipelines/autogluon_tabular_training.yaml
```diff
@@ -0,0 +1,904 @@
+# PIPELINE DEFINITION
```
dbasunag: We don't use yaml like this. Can you please check if you can utilize https://github.com/RedHatQE/openshift-python-wrapper/blob/main/ocp_resources/pipeline.py instead?

tests/pipelines_components/autorag/pipelines/documents_rag_optimization.yaml
```diff
@@ -0,0 +1,1754 @@
...
+#    input_data_key: str
+#    input_data_secret_name: str
+#    llama_stack_secret_name: str
+#    llama_stack_vector_database_id: str
+#    optimization_max_rag_patterns: int [Default: 8.0]
+#    optimization_metric: str [Default: 'faithfulness']
+#    test_data_bucket_name: str
+#    test_data_key: str
+#    test_data_secret_name: str
+components:
```
dbasunag: Same comment here.

tests/pipelines_components/autorag/test_autorag_smoke.py
```diff
@@ -0,0 +1,118 @@
...
+    "AUTORAG_S3_SECRET_NAME": "S3 credentials secret name in the DSPA namespace",
+    "AUTORAG_LLAMA_STACK_URL": "Llama Stack server base URL",
+    "AUTORAG_LLAMA_STACK_API_KEY": "Llama Stack API key / bearer token",
+    "AUTORAG_EMBEDDINGS_MODEL": "Embedding model ID",
+    "AUTORAG_GENERATION_MODEL": "Generation model ID",
+}
+
+
+@pytest.mark.smoke
+@pytest.mark.pipelines_components
```
dbasunag: Your test is under tests/pipelines_component. We don't need marks like this `@pytest.mark.pipelines_components`

tests/pipelines_components/autorag/test_autorag_smoke.py
```diff
@@ -0,0 +1,118 @@
...
+class TestAutoRAGSmoke:
+    """AutoRAG pipeline smoke test using the Documents RAG Optimization Pipeline.
+
+    Requires a pre-existing DSPA with test data in S3 and an external Llama Stack server.
+    See tests/pipelines_components/README.md for required environment variables.
+    """
+
+    @pytest.fixture(autouse=True)
```
dbasunag: Please move the fixture to the contest.py

tests/pipelines_components/README.md
```diff
@@ -0,0 +1,171 @@
...
+Mount a valid `kubeconfig` and pass pytest arguments:
+
+```bash
+podman run --rm \
+  -v "${KUBECONFIG}:/home/odh/.kube/config:z" \
+  -e KUBECONFIG=/home/odh/.kube/config \
+  quay.io/opendatahub/opendatahub-tests:latest \
+  tests/pipelines_components/automl/ -m smoke -v -s \
+  --junitxml=/home/odh/opendatahub-tests/results_xunit.xml
+```
```
dbasunag: This is common for all the components. If missing in main readme/contributing/developer's guide, then can be added there.

tests/pipelines_components/README.md
```diff
@@ -0,0 +1,171 @@
...
+
+For AutoRAG, pass the required env vars or mount a `.env` file:
+
+```bash
```
dbasunag, line 113: Do you have a jira for testops for this?

tests/pipelines_components/README.md
```diff
@@ -0,0 +1,171 @@
...
+```
+
+### Monitoring a run
+
+Pipeline runs appear in the RHOAI/ODH dashboard as `automl-smoke-<namespace>` or `autorag-smoke`.
+
+## Test Markers
+
+- `@pytest.mark.smoke` - Quality gate marker
+- `@pytest.mark.pipelines_components` - Component ownership marker
```
dbasunag: This is not needed component ownership is decided by the name of the component under tests, unless multiple teams have ownership on a given component (e.g. llama stack)

tests/pipelines_components/utils.py
```diff
@@ -0,0 +1,129 @@
...
+
+from tests.pipelines_components.constants import PIPELINE_POLL_INTERVAL
+
+LOGGER = structlog.get_logger(name=__name__)
+
+WORKFLOW_SUCCEEDED: str = "Succeeded"
+WORKFLOW_TERMINAL_PHASES: set[str] = {"Succeeded", "Failed", "Error"}
+
+
+class Workflow(NamespacedResource):
```
dbasunag: Please generate this by using https://github.com/RedHatQE/openshift-python-wrapper/tree/main/class_generator and move it under utilities/resources.

pyproject.toml
```diff
@@ -78,6 +77,7 @@ dependencies = [
     "dictdiffer>=0.9.0",
     "pytest>=9.0.0",
     "huggingface-hub>=1.2.3",
+    "ipykernel>=7.2.0",
```
dbasunag: Can you please elaborate why we need this?

pytest.ini
```diff
@@ -55,6 +55,7 @@ markers =
     model_explainability
     llama_stack
     rag
+    pipelines_components: Pipelines Components team tests (AutoML, AutoRAG)
```
dbasunag: not needed. You only need to add if you are adding to a cross component area like llama_stack

dbasunag (changes requested):

tests/pipelines_components/README.md
MRGuziX: can you elaborate?

pyproject.toml
MRGuziX: added by mistake

pytest.ini
MRGuziX: done

tests/pipelines_components/utils.py
MRGuziX: done

tests/pipelines_components/README.md
MRGuziX: done

tests/pipelines_components/README.md
MRGuziX: done

tests/pipelines_components/autorag/test_autorag_smoke.py
MRGuziX: done

tests/pipelines_components/autorag/test_autorag_smoke.py
MRGuziX: doe

tests/pipelines_components/autorag/pipelines/documents_rag_optimization.yaml
MRGuziX: done

tests/pipelines_components/automl/pipelines/autogluon_tabular_training.yaml
MRGuziX: That's a Tekton Pipeline resource (tekton.dev), not a Kubeflow Pipeline. Our AutoML/AutoRAG tests use KFP v2 pipelines which are uploaded via the DSPA REST API                  
  (/apis/v2beta1/pipelines/upload) — not created as Kubernetes CRDs.                                                                                                               
                                                                                                                                                                                   
  The pipeline YAML we have is a compiled KFP pipeline definition that gets uploaded to the DSPA API server via HTTP POST. It's not a Kubernetes resource that you'd manage with   
  ocp_resources. The Pipeline class from openshift-python-wrapper is for Tekton pipelines, which is a different pipeline engine.                                                   
                                                                                                                                                                                   
  Our current approach — uploading the compiled YAML via REST API in utils.py:upload_pipeline() — is the correct way to handle KFP v2 pipelines. The YAML file is just a static    
  artifact, not a K8s manifest.
  
  so thats why i believe we need those yamls so it can work correctly. 
  If you have something else on mind, please ping me up :)

MRGuziX:
@coderabbitai review

tests/pipelines_components/automl/test_automl_smoke.py
```diff
@@ -0,0 +1,109 @@
...
+        pipelines_namespace: Namespace,
+        dspa_s3_credentials: Secret,
+        automl_train_data: str,
+    ) -> None:
+        """Given a DSPA with training data in S3, when an AutoML pipeline run is submitted, then it succeeds."""
+        pipeline_id = None
+        run_id = None
+
+        try:
+            pipeline_id = upload_pipeline(
```
dbasunag: This test should utilize fixtures for setup and clean up and only keep the validation (that the pipeline succeeds) in the test

tests/pipelines_components/utils.py
```diff
@@ -0,0 +1,210 @@
...
+            namespace=namespace,
+            run_id=run_id,
+        ):
+            LOGGER.info(f"Pipeline run {run_id}: {phase}")
+            if phase and phase in WORKFLOW_TERMINAL_PHASES:
+                return phase
+    except TimeoutExpiredError as err:
+        msg = f"Pipeline run {run_id} did not complete within {timeout}s"
+        LOGGER.error(msg)
+        raise TimeoutError(msg) from err
```
dbasunag: Why not raise the same error?

tests/pipelines_components/utils.py
```diff
@@ -0,0 +1,210 @@
...
+    resp = requests.delete(
+        url=f"{api_url}/apis/v2beta1/pipelines/{pipeline_id}",
+        headers=headers,
+        verify=ca_bundle,
+        timeout=60,
+    )
+    if resp.ok:
+        LOGGER.info(f"Deleted pipeline {pipeline_id}")
+    else:
+        LOGGER.warning(f"Failed to delete pipeline {pipeline_id}: {resp.status_code} {resp.text}")
```
dbasunag: Should we not fail? Just a warning would not raise awareness if there is a clean up issue.

tests/pipelines_components/utils.py
```diff
@@ -0,0 +1,210 @@
...
+    resp = requests.delete(
+        url=f"{api_url}/apis/v2beta1/runs/{run_id}",
+        headers=headers,
+        verify=ca_bundle,
+        timeout=60,
+    )
+    if resp.ok:
+        LOGGER.info(f"Deleted pipeline run {run_id}")
+    else:
+        LOGGER.warning(f"Failed to delete pipeline run {run_id}: {resp.status_code} {resp.text}")
```
dbasunag: Same comment here.

tests/pipelines_components/README.md
dbasunag: if you need to mount a .env file during execution or need to pass the required env variables for this run, you would need a ticket for testops to:
1) create the job in jenkins
2) to ensure the right steps were done before execution.

uv.lock
```diff
@@ -302,11 +302,11 @@ wheels = [
 
 [[package]]
 name = "certifi"
-version = "2026.4.22"
+version = "2026.2.25"
```
dbasunag: We should not downgrade.

tests/pipelines_components/autorag/conftest.py
```diff
@@ -0,0 +1,99 @@
...
+
+@pytest.fixture(scope="session", autouse=True)
+def _validate_autorag_env() -> None:
+    missing = [f"  {var}: {desc}" for var, desc in _AUTORAG_REQUIRED_ENV.items() if not os.environ.get(var)]
+    if missing:
+        pytest.skip("AutoRAG smoke test requires environment variables:\n" + "\n".join(missing))
+
+    import tests.pipelines_components.constants as _constants
+
+    _constants.AUTORAG_PIPELINE_YAML = resolve_pipeline_yaml(value=AUTORAG_PIPELINE_YAML)
```
dbasunag: Please always import at the top of the file.

tests/pipelines_components/autorag/test_autorag_smoke.py
```diff
@@ -0,0 +1,130 @@
...
+        autorag_llama_stack_secret: Secret,
+    ) -> None:
+        """Given a DSPA with documents and benchmark data in S3, when an AutoRAG pipeline run is submitted,
+        then it succeeds."""
+        run_suffix = uuid.uuid4().hex[:8]
+        pipeline_id = None
+        run_id = None
+
+        try:
+            pipeline_id = upload_pipeline(
```
dbasunag: Same comment. Please create a fixture that uploads and cleanup the pipeline.

tests/pipelines_components/.env.example
```diff
@@ -0,0 +1,41 @@
...
+AUTORAG_EMBEDDINGS_MODEL=
+AUTORAG_GENERATION_MODEL=
+
+# Vector database for AutoRAG (optional)
+AUTORAG_VECTOR_DB_ID=
+
+# Pipeline tuning (optional — defaults shown)
+#AUTORAG_MAX_RAG_PATTERNS=1
+#AUTORAG_OPTIMIZATION_METRIC=faithfulness
+#AUTORAG_PIPELINE_TIMEOUT=3600
```
dbasunag: If they are not needed, we can clean these up.

tests/pipelines_components/conftest.py
```diff
@@ -0,0 +1,271 @@
...
+                        "valueFrom": {"secretKeyRef": {"name": DSPA_S3_SECRET, "key": "secretkey"}},
+                    },
+                ],
+            }
+        ],
+        wait_for_resource=True,
+    ) as upload_pod:
+        try:
+            upload_pod.wait_for_status(status="Succeeded", timeout=120)
+        except Exception:
```
dbasunag: Can we limit this to only TimeoutExpiredError? I don't believe you can get anything else here.

tests/pipelines_components/.env.example
MRGuziX: all three are used:
                                                                                                                                                                                   
  - AUTORAG_MAX_RAG_PATTERNS → passed as optimization_max_rag_patterns pipeline parameter in autorag_run_id fixture (autorag/conftest.py)                                          
  - AUTORAG_OPTIMIZATION_METRIC → passed as optimization_metric pipeline parameter in the same fixture                                                                             
  - AUTORAG_PIPELINE_TIMEOUT → used in wait_for_pipeline_run in the test                                                                                                           
                                                                                                                                                                                   
  They're commented out in .env.example because they have defaults in constants.py (1, faithfulness, 3600). Users only uncomment them if they want to override

tests/pipelines_components/automl/test_automl_smoke.py
MRGuziX: done

tests/pipelines_components/utils.py
MRGuziX: done

tests/pipelines_components/utils.py
MRGuziX: done

tests/pipelines_components/utils.py
MRGuziX: done

uv.lock
MRGuziX: mistake during  rebasing

tests/pipelines_components/autorag/conftest.py
MRGuziX: done

tests/pipelines_components/autorag/test_autorag_smoke.py
MRGuziX: done

tests/pipelines_components/conftest.py
MRGuziX: done

Mateusz-Switala:
Unfortunately, we need to convert the PR to draft, some configuration is failing after implement setting up llamastack inside the tests.

mwaykole (commented):
## PR Review: Add AutoGluon/AutoRAG Pipeline Smoke Tests (RHOAIENG-59399)

**Reviewer:** Milind Waykole (@mwaykole)
**Risk Level:** 🔴 HIGH — Contains a regression that breaks existing llama_stack tests

---

## Summary

This PR adds a well-structured smoke test suite for AutoGluon Tabular Training (AutoML) and AutoRAG pipelines under `tests/pipelines_components/`. The architectural approach — deploying DSPA with built-in MinIO, polling Argo Workflows directly, and env-driven configuration — is sound. However, the PR has **one blocking regression** and several issues that should be addressed before merge.

---

## 🚨 BLOCKING: Existing `llama_stack` Tests Will Break at Import Time

**File:** `tests/llama_stack/utils.py` (lines removed) vs. `tests/llama_stack/conftest.py` (unchanged)

This PR removes three functions from `tests/llama_stack/utils.py` and moves them to the new `utilities/llama_stack_utils.py`:
- `create_llama_stack_distribution`
- `wait_for_unique_llama_stack_pod`
- `wait_for_llama_stack_client_ready`

However, `tests/llama_stack/conftest.py` still imports all three from `tests.llama_stack.utils`:

```python
# tests/llama_stack/conftest.py (NOT modified in this PR)
from tests.llama_stack.utils import (
    create_llama_stack_distribution,        # ← REMOVED from utils.py
    vector_store_upload_dataset,
    vector_store_upload_doc_sources,
    wait_for_llama_stack_client_ready,      # ← REMOVED from utils.py
    wait_for_unique_llama_stack_pod,        # ← REMOVED from utils.py
)
```

This will produce an `ImportError` at collection time, causing **all existing llama_stack tests to fail**. This is confirmed by the tox CI failure on this PR (`tox-tests` job → `Run tox environments` step → `uv run pytest --collect-only` fails).

**Fix:** Either update `tests/llama_stack/conftest.py` to import from `utilities.llama_stack_utils`, or keep the functions in `tests/llama_stack/utils.py` and re-export them as shims:

```python
# Option A: Update conftest.py imports (recommended — cleaner)
from utilities.llama_stack_utils import (
    create_llama_stack_distribution,
    wait_for_llama_stack_client_ready,
    wait_for_unique_llama_stack_pod,
)

# Option B: Keep backward compat shim in tests/llama_stack/utils.py
from utilities.llama_stack_utils import (
    create_llama_stack_distribution,
    wait_for_llama_stack_client_ready,
    wait_for_unique_llama_stack_pod,
)
__all__ = ["create_llama_stack_distribution", "wait_for_llama_stack_client_ready", "wait_for_unique_llama_stack_pod"]
```

---

## 🔴 HIGH: `pytest.ini` Marker Not Registered — `--strict-markers` Will Fail

The PR description mentions a new `pipelines_components` marker, but the `pytest.ini` uses `--strict-markers`. Looking at the PR diff, the only marker used in tests is `@pytest.mark.smoke` which **is** already registered.

However, if `pipelines_components` is intended as a component marker (consistent with `llama_stack`, `rag`, `model_explainability` in `pytest.ini`), it **must be added to `pytest.ini`** before Jenkins jobs can use `pytest -m pipelines_components` to select this test suite. Without it, any CI runner using that selector will error immediately.

```ini
# pytest.ini — add under component markers
pipelines_components
```

---

## 🟡 MEDIUM: `tests/llama_stack/constants.py` — Circular / Confusing Re-export via `__all__`

**File:** `tests/llama_stack/constants.py`

The PR imports symbols from `utilities.llama_stack_utils` into `tests/llama_stack/constants.py` and then declares them in `__all__`. This creates a confusing re-export pattern:

```python
# constants.py now does:
from utilities.llama_stack_utils import (
    LLAMA_STACK_DISTRIBUTION_SECRET_DATA,
    LLS_CLIENT_VERIFY_SSL,
    LLS_CORE_POD_FILTER,
    POSTGRES_IMAGE,
)

__all__ = [
    "LLAMA_STACK_DISTRIBUTION_SECRET_DATA",
    "LLS_CLIENT_VERIFY_SSL",
    "LLS_CORE_POD_FILTER",
    "POSTGRES_IMAGE",
]
```

`constants.py` is a module of **constants**, not a re-export hub. The `__all__` list is also non-standard in this codebase — no other `constants.py` in the repo uses `__all__`. Consumers should import from `utilities.llama_stack_utils` directly (as the new autorag conftest already does). The `__all__` block should be removed, and the 4 imported symbols should either stay in `constants.py` as definitions (pre-refactor style) or be fully removed from `constants.py` if all callers are updated.

---

## 🟡 MEDIUM: Shell Injection Risk in MinIO Uploader Command

**File:** `tests/pipelines_components/conftest.py`, `automl_train_data` fixture

```python
mc_setup = "export MC_CONFIG_DIR=/work/.mc && mc alias set dspa $MINIO_ENDPOINT $MINIO_ACCESS_KEY $MINIO_SECRET_KEY"
mc_upload = (
    f"cat <<'CSVEOF' > /work/train.csv\n{AUTOML_SMOKE_CSV}CSVEOF\n"
    f"mc cp /work/train.csv dspa/{shlex.quote(DSPA_S3_BUCKET)}/{shlex.quote(AUTOML_TRAIN_DATA_FILE_KEY)}"
)
```

`DSPA_S3_BUCKET` and `AUTOML_TRAIN_DATA_FILE_KEY` are `shlex.quote`d (good), but `AUTOML_SMOKE_CSV` is interpolated directly into a heredoc. While the CSV content is a hardcoded constant today, this pattern is unsafe if the constant ever references env vars. More importantly, `mc_setup` uses shell env var expansion (`$MINIO_ENDPOINT` etc.) which is correct — the actual creds come from the Pod env. This is okay as-is, but worth a note that if `AUTOML_SMOKE_CSV` ever becomes user-supplied, it must be sanitized.

The transient uploader Pod also has **no CPU/memory resource requests/limits** defined. On resource-constrained test clusters this can cause eviction or unpredictable scheduling.

```python
# Suggested: add resources to the container spec
"resources": {
    "requests": {"cpu": "100m", "memory": "128Mi"},
    "limits": {"cpu": "500m", "memory": "256Mi"},
},
```

---

## 🟡 MEDIUM: `dspa_s3_credentials` Fixture Modifies a Secret In-Place Without Rollback

**File:** `tests/pipelines_components/conftest.py`

```python
def dspa_s3_credentials(...) -> Secret:
    secret = Secret(...)
    secret.update(resource_dict={"stringData": {"AWS_ACCESS_KEY_ID": access_key, ...}})
    return secret  # ← no teardown, no context manager
```

This fixture patches the DSPA-managed S3 secret with additional AWS-style keys but **never reverts the changes**. If the DSPA operator later reconciles this secret and removes the extra fields, or if another test expects the original layout, this mutation will leak across test runs. The fixture should restore the original secret state on teardown, or — better — create a separate pipeline-component-specific secret rather than patching the DSPA-managed one.

---

## 🟡 MEDIUM: `resolve_pipeline_yaml` Downloads to a Temp File That Is Never Cleaned Up

**File:** `tests/pipelines_components/utils.py`

```python
with tempfile.NamedTemporaryFile(suffix=".yaml", delete=False) as tmp:
    tmp.write(resp.content)
LOGGER.info(f"Pipeline YAML downloaded to {tmp.name}")
return tmp.name  # ← caller never deletes this file
```

The `delete=False` combined with no cleanup means downloaded YAMLs accumulate in `/tmp` across test runs. Consider using `tempfile.mkstemp` with an explicit `atexit.register(os.unlink, tmp_path)` or convert the calling fixture to a context manager so the file is cleaned up on teardown.

---

## 🟡 MEDIUM: AutoRAG Test Prerequisites Are Very Heavy — Needs CI Skip Guard

**File:** `tests/pipelines_components/autorag/conftest.py`

The AutoRAG test deploys:
1. A vLLM inference InferenceService (8Gi RAM, 4 CPUs)
2. A vLLM embedding InferenceService (4Gi RAM, 2 CPUs)
3. A PostgreSQL Deployment
4. A LlamaStackDistribution CR

This is a very resource-intensive setup. The `_validate_autorag_env` fixture (session-scoped, autouse=True) correctly skips when env vars are missing. However, this fixture is **session-scoped**, meaning it runs even when only `automl` tests are collected. Consider making the `autouse` scope `False` and explicitly requiring it in the AutoRAG test class, or scoping it to the autorag package only.

Additionally, the LlamaStack distribution timeout is 600 seconds but the operator enablement (`autorag_llama_stack_operator`) has no explicit timeout guard — a failed DSC update could hang indefinitely.

---

## 🟢 LOW: `utilities/resources/workflow.py` — Missing `kind` Override

**File:** `utilities/resources/workflow.py`

```python
class Workflow(NamespacedResource):
    api_group: str = "argoproj.io"
    api_version: str = "v1alpha1"
```

The `openshift-python-wrapper` convention derives `kind` from the class name (`Workflow`). This is correct here since the Argo CRD kind is indeed `Workflow`. However, it's worth confirming that the API version on the target clusters is `v1alpha1` and not `v1` — some Argo distributions use `argoproj.io/v1`. A comment with the expected CRD FQDN would improve maintainability.

---

## 🟢 LOW: `__all__` in `tests/llama_stack/constants.py` Is Non-Standard

This is covered above in the MEDIUM finding. The pattern of placing `__all__` at the bottom of a constants file, after a `try/except` block, is fragile and non-idiomatic. No other `constants.py` in this repo uses this pattern.

---

## 🟢 LOW: `pytest_configure` Hook Placed in Non-Root Conftest

**File:** `tests/pipelines_components/conftest.py`

The `pytest_configure` hook in a sub-directory conftest is valid, but it only fires when pytest collects from that subtree. When running the full suite, this hook will run alongside the root `conftest.py`'s hooks. Ensure the `.env` file loading cannot conflict with other test areas that set the same env var keys (`AUTOML_PIPELINE_YAML`, etc.). The current implementation uses `os.environ.get` with precedence (shell > .env), which is correct.

---

## 🟢 LOW: Missing `AUTORAG_PIPELINE_TIMEOUT` Export in `.env.example`

The `.env.example` documents the optional `AUTORAG_PIPELINE_TIMEOUT` in a comment, but the README table references `AUTORAG_PIPELINE_TIMEOUT` as a tunable. Minor — no functional impact.

---

## ✅ What's Good

- **Architecture is clean**: Separating AutoML (self-contained with MinIO) from AutoRAG (pre-existing DSPA) is a sensible design.
- **Security context on uploader pod** is properly hardened: `allowPrivilegeEscalation: false`, `runAsNonRoot: true`, `seccompProfile: RuntimeDefault`, capabilities dropped.
- **MINIO_MC_IMAGE is digest-pinned** (`@sha256:`), mitigating supply-chain risk.
- **Sensitive env vars are masked** in logs via the `_mask_value` regex — good practice.
- **`AUTOML_SMOKE_CSV` has enough rows (30)** for stratified train/validation splits in AutoGluon binary classification.
- **`resolve_pipeline_yaml` supports both local paths and URLs** — developer-friendly.
- **Fixture teardown for `automl_pipeline_id` and `automl_run_id`** correctly deletes pipeline and run artifacts.
- **`collect_pipeline_pod_logs`** on failure is a good debugging pattern.
- **`utilities/llama_stack_utils.py`** centralizing shared LlamaStack utilities is the right direction — it just needs the conftest update to complete the migration.
- **`enabled_pipelines_in_dsc` fixture** properly uses `ResourceEditor` context manager for reversible DSC patching.

---

## Required Changes Before Merge

1. **[BLOCKING]** Update `tests/llama_stack/conftest.py` to import `create_llama_stack_distribution`, `wait_for_unique_llama_stack_pod`, and `wait_for_llama_stack_client_ready` from `utilities.llama_stack_utils` (or re-export them from `tests.llama_stack.utils` as shims).
2. **[HIGH]** Register `pipelines_components` marker in `pytest.ini` if it is intended to be used as a component selector in Jenkins.
3. **[MEDIUM]** Remove the `__all__` block from `tests/llama_stack/constants.py` — it's non-idiomatic and unnecessary.
4. **[MEDIUM]** Add resource limits to the MinIO uploader Pod in `automl_train_data` fixture.
5. **[MEDIUM]** Fix `dspa_s3_credentials` fixture to restore the secret on teardown (or use a separate secret).
6. **[MEDIUM]** Clean up the temp file in `resolve_pipeline_yaml` after it's consumed.

---

**Bottom line:** The feature work is solid and the test patterns are good. The blocking issue is purely a refactoring gap — the existing `llama_stack` tests will crash at import time because their conftest was not updated to match the moved symbols. Fix that, address the medium-severity items, and this is ready to merge.

/hold

MRGuziX:
### What was done:

  1. __all__ re-export in tests/llama_stack/constants.py — Removed the __all__ block and the unused import from utilities.llama_stack_utils. Updated the one consumer              
  (tests/llama_stack/operator/conftest.py) to import LLS_CORE_POD_FILTER from the canonical location.
  2. Broken imports in tests/llama_stack/conftest.py — Three functions (create_llama_stack_distribution, wait_for_unique_llama_stack_pod, wait_for_llama_stack_client_ready) and   
  three constants (LLAMA_STACK_DISTRIBUTION_SECRET_DATA, LLS_CLIENT_VERIFY_SSL, POSTGRES_IMAGE) were moved to utilities/llama_stack_utils.py but conftest still imported them from 
  the old locations. This caused ImportError at collection time, breaking all llama_stack tests. Updated the imports.
  3. Missing pipelines_components marker — Registered in pytest.ini and added @pytest.mark.pipelines_components to both TestAutoMLSmoke and TestAutoRAGSmoke. Without this,        
  --strict-markers would reject CI runs using -m pipelines_components.                                                                                                             
  4. Temp file leak in resolve_pipeline_yaml — Added atexit.register(os.unlink, tmp.name) so downloaded pipeline YAMLs are cleaned up on process exit.
                                                                                                                                                                                   
  Not fixed (no action needed)                                                                                                                                                     
                                                                                                                                                                                   
  - Workflow API version v1alpha1 — Correct. Argo Workflows has never graduated past v1alpha1.                                                                                     
  - pytest_configure in sub-directory conftest — Fine. All env var names are AUTOML_/AUTORAG_-prefixed, no collision risk. Shell precedence is already respected.
  - AUTORAG_PIPELINE_TIMEOUT missing from .env.example — It's already there as a comment on line 44. Consistent enough.                                                            
  - AutoRAG session-scoped _validate_autorag_env running for automl — Non-issue. It lives in autorag/conftest.py, which pytest only loads when collecting from that subdirectory.  
  - dspa_s3_credentials not reverting secret — Non-issue. The secret lives in an ephemeral namespace that's deleted on teardown.                                                   
  - Shell injection in AUTOML_SMOKE_CSV — Non-issue. It's a hardcoded constant, and the heredoc uses <<'CSVEOF' (single-quoted) which disables shell expansion.                    
  - README out of date — Real but separate from code correctness. Flagged in the initial analysis, not addressed in these fixes.

MRGuziX:
After introducing managed pipelines:
The AUTO ML TEST:
image

MRGuziX:
We would like to merge that PR with:
Running AutoML tests as they are proven to be working with current cluster of nightly builds.
https://jenkins-csb-rhods-opendatascience.dno.corp.redhat.com/job/devops/job/components-testing-config-checker
image


AutoRAG will be switch on when that issue will be fixed: https://redhat.atlassian.net/browse/RHOAIENG-68247

but the code and test itself it is working and it was prove on our cluster 
image

MRGuziX:
@mwaykole  @dbasunag  if you could take a look on a changes and my comment above

tests/pipelines_components/automl/conftest.py
```diff
@@ -0,0 +1,150 @@
...
+    dspa: DataSciencePipelinesApplication,
+    dspa_api_url: str,
+    dspa_auth_headers: dict[str, str],
+    dspa_ca_bundle_file: str,
+) -> dict[str, str] | None:
+    """Discovered managed pipeline info, or None in legacy mode."""
+    if not use_managed_pipelines(yaml_env_value=AUTOML_PIPELINE_YAML):
+        return None
+    LOGGER.info(f"DSPA ready; sleeping {DSPA_READY_BUFFER_SECONDS}s for managed pipeline registration")
+    time.sleep(DSPA_READY_BUFFER_SECONDS)
```
mwaykole: possible to use TimeoutSampler

tests/pipelines_components/autorag/conftest.py
```diff
@@ -0,0 +1,1138 @@
...
+
+@pytest.fixture(scope="session", autouse=True)
+def _validate_autorag_env() -> None:
+    if AUTORAG_PIPELINE_YAML:
+        LOGGER.info("AUTORAG_PIPELINE_YAML is set — using legacy YAML upload mode")
+    else:
+        LOGGER.info("AUTORAG_PIPELINE_YAML is not set — using managed pipeline mode")
+    missing = [f"  {var}: {desc}" for var, desc in _AUTORAG_REQUIRED_ENV.items() if not os.environ.get(var)]
+    if missing:
+        pytest.skip("AutoRAG smoke test requires environment variables:\n" + "\n".join(missing))
```
dbasunag: Please fail here, instead of skips. Skips are considered release blockers now.

tests/pipelines_components/autorag/conftest.py
```diff
@@ -0,0 +1,1138 @@
...
+# Override parent namespace fixture with a shorter name to stay under the
+# 63-char DNS label limit for KServe predictor hostnames.
+# ---------------------------------------------------------------------------
+
+
+@pytest.fixture(scope="class")
+def pipelines_namespace(admin_client: DynamicClient) -> Generator[Namespace, Any, Any]:  # noqa: UFN001
+    from ocp_resources.exceptions import ResourceTeardownError
+
+    from utilities.infra import create_ns
```
dbasunag: Please move them to the top of this file.

tests/pipelines_components/autorag/conftest.py
```diff
@@ -0,0 +1,1138 @@
...
+            yield namespace
+    except ResourceTeardownError:
+        LOGGER.warning("Namespace %s teardown timed out, force-removing finalizers", namespace.name)
+        try:
+            ns_json = namespace.instance.to_dict()
+            ns_json["spec"]["finalizers"] = []
+            admin_client.request("PUT", f"/api/v1/namespaces/{namespace.name}/finalize", body=ns_json)
+            LOGGER.info("Force-finalized namespace %s", namespace.name)
+        except Exception:  # noqa: BLE001
+            LOGGER.warning("Could not force-finalize namespace %s — delete manually", namespace.name)
```
dbasunag: Should not we triage why this happened and fail? This can hide bug (either in automation clean up order or in your product code.).

tests/pipelines_components/autorag/conftest.py
```diff
@@ -0,0 +1,1138 @@
...
+        model_env_variables=[{"name": "VLLM_CPU_KVCACHE_SPACE", "value": "2"}],
+        argument=["--served-model-name", served_model_name, "--max-model-len", "4096"],
+    ) as isvc:
+        yield isvc
+
+
+@pytest.fixture(scope="class")
+def autorag_inference_url(autorag_inference_service: InferenceService, pipelines_namespace: Namespace) -> str:
+    try:
+        return f"{autorag_inference_service.instance.status.address.url}/v1"
```
dbasunag: Would this be a bug?

tests/pipelines_components/autorag/conftest.py
```diff
@@ -0,0 +1,1138 @@
...
+            AUTORAG_EMBEDDING_MAX_MODEL_LEN,
+        ],
+    ) as isvc:
+        yield isvc
+
+
+@pytest.fixture(scope="class")
+def autorag_embedding_url(autorag_embedding_service: InferenceService, pipelines_namespace: Namespace) -> str:
+    try:
+        return f"{autorag_embedding_service.instance.status.address.url}/v1"
```
dbasunag: Same question, would it not be a bug?

tests/pipelines_components/autorag/conftest.py
```diff
@@ -0,0 +1,1138 @@
...
+        min_ready_seconds=10,
+    )
+    deployment.timeout_seconds = 240
+    deployment.wait(timeout=240)
+    deployment.wait_for_replicas()
+    pod = _wait_for_unique_ogx_pod(client=admin_client, namespace=autorag_ogx_server.namespace)
+
+    try:
+        logs = pod.log()
+        LOGGER.info("OGX pod startup log", logs=logs[-8000:] if len(logs) > 8000 else logs)
```
dbasunag: This can flood the console. Why do we need it?

tests/pipelines_components/autorag/conftest.py
```diff
@@ -0,0 +1,1138 @@
...
+    registered_ids = {m.id for m in models.data}
+    LOGGER.info(
+        "OGX registered models",
+        models=[
+            {
+                "id": m.id,
+                "model_type": str(getattr(m, "model_type", "?")),
+                "custom_metadata": getattr(m, "custom_metadata", {}),
+            }
+            for m in models.data
```
dbasunag: can you please use model instead of `m`? We prefer words vs letter in variable names in this repo.

tests/pipelines_components/autorag/conftest.py
```diff
@@ -0,0 +1,1138 @@
...
+    except Exception as exc:  # noqa: BLE001
+        LOGGER.info(
+            "vLLM URL not reachable from test runner (cluster-internal); skipping readiness wait",
+            url=vllm_base_url,
+            model=model_name,
+            reason=str(exc),
+        )
+        return
+
+    deadline = time.monotonic() + timeout
```
dbasunag: Please use timeout sampler

tests/pipelines_components/conftest.py
```diff
@@ -0,0 +1,312 @@
...
+    if env_file.is_file():
+        loaded = {}
+        for line in env_file.read_text().splitlines():
+            line = line.strip()
+            if not line or line.startswith("#") or "=" not in line:
+                continue
+            key, _, value = line.partition("=")
+            loaded[key.strip()] = os.environ.get(key.strip(), value.strip())
+
+        logger = structlog.get_logger(name=__name__)
```
dbasunag: Should be at the top of this file.

tests/pipelines_components/conftest.py
```diff
@@ -0,0 +1,312 @@
...
+        logger.info(  # noqa: FCN001
+            "Loaded .env file",
+            path=str(env_file),
+            variables={k: _mask_value(key=k, value=v) for k, v in loaded.items()},
```
dbasunag: same comment about k, v here.

tests/pipelines_components/conftest.py
```diff
@@ -0,0 +1,312 @@
...
+        )
+
+    if config.option.collectonly:
+        return
+
+    try:
+        existing = get_rhods_subscription()
+    except Exception:  # noqa: BLE001
+        LOGGER.debug("No cluster connection available — skipping RHOAI operator check")
+        return
```
dbasunag: if no cluster connection is available, should not we fail? If we are not using collect-only, then the only other use case would be running test.

tests/pipelines_components/conftest.py
```diff
@@ -0,0 +1,312 @@
...
+
+    if existing:
+        installed_csv = existing.instance.status.get("installedCSV", "unknown")
+        LOGGER.info(f"RHOAI operator installed: {existing.name} (CSV: {installed_csv})")
+    else:
+        LOGGER.info("RHOAI operator is not installed on this cluster")
```
dbasunag: Same comment

tests/pipelines_components/conftest.py
```diff
@@ -0,0 +1,312 @@
...
+def _skip_teardown_if_requested():
+    """When SKIP_TEARDOWN=true, prevent all ocp_resources from being deleted on teardown.
+
+    This keeps the namespace, pods, routes, and all resources alive after the test
+    so that teams can inspect logs and state.
+    Delete the namespace manually when done: oc delete ns <namespace>
+    """
+    if os.getenv("SKIP_TEARDOWN", "").lower() not in ("true", "1", "yes"):
+        yield
+        return
```
dbasunag: We should not have to use yield and return together.  Instead of just if, please use if/else

tests/pipelines_components/conftest.py
```diff
@@ -0,0 +1,312 @@
...
+    LOGGER.warning("SKIP_TEARDOWN is set — all resources will be kept after test completion")
+    from ocp_resources.resource import Resource
```
dbasunag: Always have the imports at the top of the file.

tests/pipelines_components/conftest.py
```diff
@@ -0,0 +1,312 @@
...
+                "spec": {
+                    "components": {
+                        "aipipelines": {"managementState": "Managed"},
+                    }
+                }
+            }
+        }
+    ):
+        wait_for_dsc_status_ready(dsc_resource=dsc_resource)
+        yield dsc_resource
```
dbasunag, line 137: you should also have wait_for_dsc_status_ready(dsc_resource=dsc_resource) out side of yield.

pyproject.toml
```diff
@@ -111,6 +110,9 @@ dependencies = [
     "openai>=2.36",
     "ragas>=0.4",
     "langchain-community<0.4.2",
+    "pyrefly>=1.0",
```
dbasunag: Please undo this one

dbasunag (changes requested):

tests/pipelines_components/autorag/conftest.py
MRGuziX: Removed. The _wait_for_unique_ogx_pod still validates the pod exists — we just don't dump its logs anymore.

tests/pipelines_components/autorag/conftest.py
MRGuziX: You're right. Silently force-finalizing masks real issues stuck finalizers usually mean something went wrong with teardown order or a product bug. Better to let it fail so we investigate.

tests/pipelines_components/autorag/conftest.py
MRGuziX: Yes, it could be. If status.address.url is None (ISVC not ready, or RawDeployment mode where the status URL isn't populated), you'd get None/v1 as a string instead of a real URL

pyproject.toml
```diff
@@ -56,7 +56,6 @@ dev = [
     "ipdb>=0.13.13",
     "ipython>=8.12.3",
     "pre-commit>=4.3.0",
-    "pyrefly>=1.0",
```
dbasunag, line 59: pyrefly needs to be added back. It is a dev requirement.

tests/pipelines_components/utils.py
```diff
@@ -0,0 +1,328 @@
...
+
+
+def collect_pipeline_pod_logs(
+    admin_client: DynamicClient,
+    namespace: str,
+    run_id: str,
+) -> None:
+    """Log failed workflow node messages for post-failure debugging."""
+    found = False
+    for workflow in Workflow.get(client=admin_client, namespace=namespace):
```
dbasunag: Workflow(client=admin_client, namespace=namespace, label_selector="pipeline/runid={run_id}") you get you a list with matching workflow.

tests/pipelines_components/utils.py
```diff
@@ -0,0 +1,328 @@
...
+def get_workflow_phase(
+    admin_client: DynamicClient,
+    namespace: str,
+    run_id: str,
+) -> str | None:
+    """Get the phase of the Argo Workflow associated with a pipeline run ID."""
+    for workflow in Workflow.get(client=admin_client, namespace=namespace):
+        labels = workflow.instance.metadata.get("labels", {})
+        if labels.get("pipeline/runid") == run_id:
+            return workflow.instance.get("status", {}).get("phase")
```
dbasunag: Workflow(client=admin_client, namespace=namespace, label_selector="pipeline/runid={run_id}") you get you a list with matching workflow.

tests/pipelines_components/autorag/test_autorag_smoke.py
dbasunag: It is still present. Can we please remove these?

tests/pipelines_components/utils.py
MRGuziX: fair

tests/pipelines_components/utils.py
MRGuziX: fair

pyproject.toml
MRGuziX: added back

dbasunag (approved):
I see merge conflict though.

pyproject.toml
dbasunag: Ok, this undoes the work done in https://github.com/opendatahub-io/opendatahub-tests/pull/1838. Can you please take this change out?

dbasunag (changes requested):
You would need to remove the change that undos https://github.com/opendatahub-io/opendatahub-tests/pull/1838

pytest.ini
```diff
@@ -58,6 +58,7 @@ markers =
     # multiple subfolders at /tests/ogx
     ai_safety
     ogx
+    pipelines_components
```
dbasunag: This needs to be removed

pyproject.toml
```diff
@@ -111,6 +111,8 @@ dependencies = [
     "openai>=2.36",
     "ragas>=0.4",
     "langchain-community<0.4.2",
+    "ipykernel>=7.3.0",
```
dbasunag, line 114: This should cause a uv.lock to be generated.

dbasunag (approved):

dbasunag:
Merging as requested by @MRGuziX

PR #1452, opened 2026-04-24
fix: Add image validation for async-job image

dbasunag:
# Pull Request

## Related Issues

- Fixes: 
- JIRA: 

- [x] Locally
- [ ] Jenkins

## Additional Requirements

## Summary by CodeRabbit

* **Tests**
  * Added validation coverage for async upload job container images, including format and related-image reference checks.
  * Added fixtures to deploy and inspect ephemeral async jobs to streamline image validation across test scenarios.
* **Refactor**
  * Consolidated job-pod lookup into a shared test utility to centralize and reuse pod retrieval logic.

### DIFF
--- a/tests/model_registry/image_validation/conftest.py
+++ b/tests/model_registry/image_validation/conftest.py
@@ -3,10 +3,13 @@ from typing import Any
 
 import pytest
 from kubernetes.dynamic import DynamicClient
+from ocp_resources.config_map import ConfigMap
 from ocp_resources.pod import Pod
 from pytest import FixtureRequest
+from pytest_testconfig import config as py_config
 
 from utilities.general import wait_for_pods_by_labels
+from utilities.infra import ResourceNotFoundError
 
 
 @pytest.fixture(scope="class")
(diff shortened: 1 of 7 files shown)


tests/model_registry/image_validation/conftest.py
```diff
@@ -32,3 +35,21 @@ def resource_pods(request: FixtureRequest, admin_client: DynamicClient) -> list[
     label_selector = request.param.get("label_selector")
     assert namespace
     return list(Pod.get(namespace=namespace, label_selector=label_selector, client=admin_client))
+
+
+@pytest.fixture(scope="session")
+def async_upload_image(admin_client: DynamicClient) -> str:
```
lugi0: The current tests spin up pods and then get the list of container images from the pod in order to validate them, I would expect we do the same for the async job (i.e. create pod/job, check its YAML for the image it is using, validate that image).
I don't know how much runtime that would add vs. checking the image defined in the configmap, but this way we would be 100% sure that the image rhoai is actually using is "correct"/validated.

lugi0 (changes requested):

tests/model_registry/image_validation/conftest.py
dbasunag: With the change it is adding a few seconds. Not significant enough. Also with your suggestion it is now more aligned with existing automation in this area.

lugi0 (approved):

sheltoncyril (approved):
/lgtm

PR #1458, opened 2026-04-26
feat(evalhub): add ServiceMonitor and NetworkPolicy monitoring tests

ruivieira:
## Summary

Implements integration tests for the EvalHub Prometheus monitoring feature introduced in [trustyai-service-operator#713](https://github.com/trustyai-explainability/trustyai-service-operator/pull/713), covering the acceptance criteria defined in [RHOAIENG-58857](https://redhat.atlassian.net/browse/RHOAIENG-58857).

## Type

- [ ] feat
- [ ] fix
- [ ] docs
- [ ] refactor / chore
- [x] test / ci

### `test_evalhub_servicemonitor.py`

| Test | Class | Purpose |
|---|---|---|
| `test_servicemonitor_created` | `TestEvalHubServiceMonitor` | SM exists after EvalHub deploy |
| `test_servicemonitor_name` | `TestEvalHubServiceMonitor` | Name follows `{evalhub}-metrics` convention |
| `test_servicemonitor_labels` | `TestEvalHubServiceMonitor` | Carries `app`, `component`, `instance` labels |
| `test_servicemonitor_single_endpoint` | `TestEvalHubServiceMonitor` | Exactly one scrape endpoint defined |
| `test_servicemonitor_endpoint_path` | `TestEvalHubServiceMonitor` | Scrape path is `/metrics` |
| `test_servicemonitor_endpoint_scheme` | `TestEvalHubServiceMonitor` | Scheme is `https` |
| `test_servicemonitor_endpoint_port` | `TestEvalHubServiceMonitor` | Port name is `https` |
| `test_servicemonitor_scrape_interval` | `TestEvalHubServiceMonitor` | Interval is `30s` |
| `test_servicemonitor_honor_labels` | `TestEvalHubServiceMonitor` | `honorLabels` is enabled |
| `test_servicemonitor_tls_configured` | `TestEvalHubServiceMonitor` | TLS config is present |
| `test_servicemonitor_tls_server_name` | `TestEvalHubServiceMonitor` | TLS serverName matches service DNS |
| `test_servicemonitor_tls_ca_configmap` | `TestEvalHubServiceMonitor` | CA sourced from `{evalhub}-service-ca` ConfigMap |
| `test_servicemonitor_tls_no_insecure_skip` | `TestEvalHubServiceMonitor` | `insecureSkipVerify` is not set |
| `test_servicemonitor_namespace_selector` | `TestEvalHubServiceMonitor` | NamespaceSelector scoped to EvalHub namespace |
[...]

PR #1464, opened 2026-04-27
test: Add EvalHub ServiceMonitor automated tests (RHAISTRAT-1507)

fege:
Implements 19/24 test cases (5 marked manual)
  - ServiceMonitor lifecycle
  - Prometheus scraping validation
  - NetworkPolicy + RBAC security
  - Upgrade scenarios
  - E2E workflows

  Manual tests (TC-NEG-*, TC-UPGRADE-002/003) special cluster setup:
  - CRD removal, NetworkPolicy manipulation,
  - RBAC modification, operator rollback

  Test plan: https://github.com/fege/collection-tests/pull/5
  Strategy: https://redhat.atlassian.net/browse/RHAISTRAT-1507

  Co-Authored-By: Claude Sonnet 4.5 (1M context) <noreply@anthropic.com>

# Pull Request

## Related Issues

- Fixes: 
- JIRA: 

- [ ] Locally
- [ ] Jenkins

## Additional Requirements

### DIFF
--- a/tests/model_explainability/evalhub/conftest.py
+++ b/tests/model_explainability/evalhub/conftest.py
@@ -19,10 +19,12 @@ from ocp_resources.role_binding import RoleBinding
 from ocp_resources.route import Route
 from ocp_resources.secret import Secret
 from ocp_resources.service_account import ServiceAccount
+from ocp_resources.service_monitor import ServiceMonitor
 from pytest_testconfig import config as py_config
 from timeout_sampler import TimeoutExpiredError, TimeoutSampler
 
 from tests.model_explainability.evalhub.constants import (
+    EVALHUB_APP_LABEL,
     EVALHUB_JOB_CONFIG_CLUSTERROLE,
     EVALHUB_JOB_SA_PREFIX,
     EVALHUB_JOB_SA_SUFFIX,
@@ -101,6 +103,26 @@ def evalhub_ca_bundle_file(
 ) -> str:
     """Create a CA bundle file for verifying the EvalHub route TLS certificate."""
     return create_ca_bundle_file(client=admin_client)
+
+
+@pytest.fixture(scope="class")
+def evalhub_service_monitor(
+    admin_client: DynamicClient,
+    model_namespace: Namespace,
+    evalhub_deployment: Deployment,
+) -> ServiceMonitor:
+    """Get the ServiceMonitor created by the TrustyAI operator for EvalHub metrics scraping."""
+    service_monitors = list(
+        ServiceMonitor.get(
+            client=admin_client,
+            namespace=model_namespace.name,
+            label_selector=f"app={EVALHUB_APP_LABEL}",
+        )
+    )
+    assert service_monitors, (
+        f"No ServiceMonitor found in namespace '{model_namespace.name}' with label app={EVALHUB_APP_LABEL}"
+    )
+    return service_monitors[0]
 
 
 # ---------------------------------------------------------------------------
--- a/tests/model_explainability/evalhub/test_evalhub_servicemonitor_e2e.py
+++ b/tests/model_explainability/evalhub/test_evalhub_servicemonitor_e2e.py
@@ -1,0 +1,236 @@
+"""EvalHub ServiceMonitor E2E workflow tests.
+
+Test Plan Reference: RHAISTRAT-1507
+Test Cases: TC-E2E-001, TC-E2E-002, TC-E2E-003
+"""
+
+import json
+import time
+
+import pytest
+import structlog
+from kubernetes.dynamic import DynamicClient
+from ocp_resources.deployment import Deployment
+from ocp_resources.evalhub import EvalHub
+from ocp_resources.namespace import Namespace
+from ocp_resources.network_policy import NetworkPolicy
+from ocp_resources.pod import Pod
+from ocp_resources.service_monitor import ServiceMonitor
+from timeout_sampler import TimeoutExpiredError, TimeoutSampler
+
+from tests.model_explainability.evalhub.constants import EVALHUB_APP_LABEL
+from utilities.constants import Timeout
+
+LOGGER = structlog.get_logger(name=__name__)
+
+
+@pytest.mark.parametrize(
+    "model_namespace",
+    [
+        pytest.param(
+            {"name": "test-evalhub-e2e-happy-path"},
+        ),
+    ],
+    indirect=True,
+)
+@pytest.mark.smoke
+@pytest.mark.model_explainability
+class TestEvalHubE2EHappyPath:
+    """E2E: Deploy EvalHub → ServiceMonitor → Prometheus scrapes → metrics queryable (RHAISTRAT-1507)."""
+
+    def test_deploy_evalhub_to_queryable_metrics(
+        self,
+        admin_client: DynamicClient,
+        model_namespace: Namespace,
+        evalhub_cr: EvalHub,
+        evalhub_deployment: Deployment,
+        evalhub_service_monitor: ServiceMonitor,
+    ) -> None:
+        """TC-E2E-001: Validate complete happy path from EvalHub deployment to queryable metrics."""
+        assert evalhub_deployment.instance.status.availableReplicas > 0, (
+            f"EvalHub deployment '{evalhub_deployment.name}' has no available replicas"
+        )
+
+        endpoints = evalhub_service_monitor.instance.spec.endpoints
+        endpoint = endpoints[0]
+        assert endpoint.port == "https" and endpoint.path == "/metrics" and endpoint.interval == "30s"
+
+        owner_refs = evalhub_service_monitor.instance.metadata.ownerReferences
+        evalhub_owner_ref = next((ref for ref in owner_refs if ref.kind == "EvalHub"), None)
+        assert evalhub_owner_ref and evalhub_owner_ref.name == evalhub_cr.name
+
+        prometheus_pods = list(
+            Pod.get(
+                client=admin_client,
+                namespace="openshift-user-workload-monitoring",
+                label_selector="app.kubernetes.io/name=prometheus",
+            )
+        )
+        if not prometheus_pods:
+            pytest.skip("User workload monitoring not enabled - cannot verify Prometheus scraping")
+
+        prometheus_pod = prometheus_pods[0]
+        evalhub_up_query = f'up{{namespace="{model_namespace.name}", job=~".*evalhub.*"}}'
+
+        def _check_metric_up() -> bool:
+            try:
+                exec_result = prometheus_pod.execute(
+                    command=[
+                        "curl",
+                        "-s",
+                        "--data-urlencode",
+                        f"query={evalhub_up_query}",
+                        "http://localhost:9090/api/v1/query",
+                    ],
+                    container="prometheus",
+                )
+                if not exec_result:
+                    return False
+                response = json.loads(exec_result)
+                results = response.get("data", {}).get("result", [])
+                return any(r.get("value", [None, "0"])[1] == "1" for r in results)
+            except json.JSONDecodeError, KeyError, IndexError, TypeError:
+                return False
+
+        try:
+            for _ in TimeoutSampler(wait_timeout=90, sleep=5, func=_check_metric_up):
+                if _check_metric_up():
+                    LOGGER.info("E2E happy path successful: EvalHub → ServiceMonitor → Prometheus → queryable")
+                    break
+        except TimeoutExpiredError as err:
+            raise AssertionError("Prometheus did not scrape EvalHub within 90 seconds") from err
+
+
+@pytest.mark.parametrize(
+    "model_namespace",
+    [
+        pytest.param(
+            {"name": "test-evalhub-e2e-network-policy"},
+        ),
+    ],
+    indirect=True,
+)
+@pytest.mark.smoke
+@pytest.mark.model_explainability
+class TestEvalHubE2ENetworkPolicy:
+    """E2E: ServiceMonitor and NetworkPolicy enable scraping in default-deny namespace (RHAISTRAT-1507)."""
+
+    def test_servicemonitor_and_networkpolicy_enable_scraping(
+        self,
+        admin_client: DynamicClient,
+        model_namespace: Namespace,
+    ) -> None:
+        """TC-E2E-002: Validate ServiceMonitor and NetworkPolicy created atomically for default-deny namespace."""
+        with (
+            NetworkPolicy(
+                client=admin_client,
+                name="default-deny-ingress",
+                namespace=model_namespace.name,
+                pod_selector={},
+                policy_types=["Ingress"],
+                ingress=[],
+                wait_for_resource=True,
+            ),
+            EvalHub(
+                client=admin_client,
+                name="evalhub-netpol-test",
+                namespace=model_namespace.name,
+                database={"type": "sqlite"},
+                wait_for_resource=True,
+            ) as evalhub_cr,
+        ):
+            deployment = Deployment(
+                client=admin_client,
+                name=evalhub_cr.name,
+                namespace=model_namespace.name,
+            )
+            deployment.wait_for_replicas(timeout=Timeout.TIMEOUT_5MIN)
+
+            time.sleep(10)
+
+            service_monitors = list(
+                ServiceMonitor.get(
+                    client=admin_client,
+                    namespace=model_namespace.name,
+                    label_selector=f"app={EVALHUB_APP_LABEL}",
+                )
+            )
+            assert any(evalhub_cr.name in sm.name for sm in service_monitors)
+
+            network_policies = list(NetworkPolicy.get(client=admin_client, namespace=model_namespace.name))
+            evalhub_netpol = next(
+                (np for np in network_policies if "evalhub" in np.name.lower() and np.name != "default-deny-ingress"),
+                None,
+            )
+            assert evalhub_netpol is not None, "NetworkPolicy not created for EvalHub metrics scraping"
+
+            ingress_rules = evalhub_netpol.instance.spec.get("ingress", [])
+            assert ingress_rules, f"NetworkPolicy '{evalhub_netpol.name}' has no ingress rules"
+
+            LOGGER.info("E2E NetworkPolicy test successful: Both ServiceMonitor and NetworkPolicy created")
+
+
+@pytest.mark.parametrize(
+    "model_namespace",
+    [
+        pytest.param(
+            {"name": "test-evalhub-e2e-cleanup"},
+        ),
+    ],
+    indirect=True,
+)
+@pytest.mark.smoke
+@pytest.mark.model_explainability
+class TestEvalHubE2ECleanup:
+    """E2E: Deploy EvalHub → delete EvalHub → ServiceMonitor cleaned up (RHAISTRAT-1507)."""
+
+    def test_evalhub_deletion_triggers_servicemonitor_cleanup(
+        self,
+        admin_client: DynamicClient,
+        model_namespace: Namespace,
+    ) -> None:
+        """TC-E2E-003: Validate complete lifecycle with automatic cleanup via garbage collection."""
+        evalhub_name = "evalhub-cleanup-test"
+        with EvalHub(
+            client=admin_client,
+            name=evalhub_name,
+            namespace=model_namespace.name,
+            database={"type": "sqlite"},
+            wait_for_resource=True,
+        ) as evalhub_cr:
+            deployment = Deployment(
+                client=admin_client,
+                name=evalhub_cr.name,
+                namespace=model_namespace.name,
+            )
+            deployment.wait_for_replicas(timeout=Timeout.TIMEOUT_5MIN)
+
+            service_monitors = list(
+                ServiceMonitor.get(
+                    client=admin_client,
+                    namespace=model_namespace.name,
+                    label_selector=f"app={EVALHUB_APP_LABEL}",
+                )
+            )
+            evalhub_sm = next((sm for sm in service_monitors if evalhub_name in sm.name), None)
+            assert evalhub_sm is not None, f"ServiceMonitor not created for EvalHub '{evalhub_name}'"
+            sm_name = evalhub_sm.name
+
+        # EvalHub deleted, wait for ServiceMonitor GC
+        try:
+            for _ in TimeoutSampler(
+                wait_timeout=60,
+                sleep=2,
+                func=lambda: (
+                    not ServiceMonitor(
+                        client=admin_client,
+                        name=sm_name,
+                        namespace=model_namespace.name,
+                    ).exists
+                ),
+            ):
+                if not ServiceMonitor(client=admin_client, name=sm_name, namespace=model_namespace.name).exists:
+                    LOGGER.info(f"E2E cleanup successful: ServiceMonitor '{sm_name}' removed via garbage collection")
+                    break
+        except TimeoutExpiredError as err:
+            raise AssertionError(f"ServiceMonitor '{sm_name}' not GC'd within 60 seconds") from err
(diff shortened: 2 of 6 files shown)


ruivieira:
@fege My feedback on this PR:

There are syntax errors that prevent collection: [`test_evalhub_servicemonitor_e2e.py`](https://github.com/opendatahub-io/opendatahub-tests/pull/1464/files#diff-e36b6694c5f7a3e10be03cb0ce4c7e96a9c7c0af1b31e8b42a7bffef4a7c0a57) and [`test_evalhub_servicemonitor_prometheus.py`](https://github.com/opendatahub-io/opendatahub-tests/pull/1464/files#diff-e8a8b7e9d5d3e0b6f2b1c4a7e3a5c8f9b2d6e1a4c7f0b3e6d9a2c5f8b1e4d7) both use Python 2-style bare-tuple `except` syntax:

```python
except json.JSONDecodeError, KeyError, IndexError, TypeError:
```

But this is a `SyntaxError` in Python 3, IIUC. The module will fail to import and no tests in it will run. The right way is:

```python
except (json.JSONDecodeError, KeyError, IndexError, TypeError):
```

The NetworkPolicy tests are testing the wrong thing. The PR trustyai-service-operator#713 _removed_ the operator-managed NetworkPolicy. [PR #1458](https://github.com/opendatahub-io/opendatahub-tests/pull/1458) has this correctly with `test_operator_does_not_create_networkpolicy`, to asserts the absence of any operator-created NetworkPolicy.

This PR goes the other way: [`TC-NP-001`](https://github.com/opendatahub-io/opendatahub-tests/pull/1464/files#diff-a9e3b0c1d7f8e2a5b4c6d1e0f3a7b2c5d8e1f4a7b0c3d6e9f2a5b8c1d4e7f0) asserts `evalhub_netpol is not None`, and [`TC-E2E-002`](https://github.com/opendatahub-io/opendatahub-tests/pull/1464/files#diff-e36b6694c5f7a3e10be03cb0ce4c7e96a9c7c0af1b31e8b42a7bffef4a7c0a57) does the same. On a real deployment these tests will always fail. The right thing to test here is what #1458 tests: that no NetworkPolicy blocks port 8443 from the monitoring namespace, and that the operator no longer creates one at all.

The TLS assertion is backwards: [`TC-CFG-001`](https://github.com/opendatahub-io/opendatahub-tests/pull/1464/files#diff-b8c2d5e7a1f4b7c0e3d6a9f2b5c8e1d4f7a0b3c6e9d2a5f8b1c4e7d0a3f6b9) asserts:

```python
assert tls_config.insecureSkipVerify is True
```

The ServiceMonitor uses an OpenShift service-CA ConfigMap (`{evalhub}-service-ca`) for full certificate validation, which means `insecureSkipVerify` should explicitly **not** be set. This is what [#1458's `test_servicemonitor_tls_no_insecure_skip`](https://github.com/opendatahub-io/opendatahub-tests/pull/1458/files#diff-f3a7b0c1d4e7f0a3b6c9d2e5f8a1b4c7d0e3f6a9b2c5d8e1f4a7b0c3d6e9f2) validates. As it is, TC-CFG-001 would pass on a misconfigured ServiceMonitor and fail on a correct one.

TC-RBAC-002 isn't doing what it claims: [The test body](https://github.com/opendatahub-io/opendatahub-tests/pull/1464/files#diff-a9e3b0c1d7f8e2a5b4c6d1e0f3a7b2c5d8e1f4a7b0c3d6e9f2a5b8c1d4e7f0) even says so in its own comments: it only checks that `ownerReferences` is set and that the namespace matches, which are covered elsewhere. The comment block at the end describes the test that would actually validate restricted user access (`oc auth can-i`, impersonation), but none of that is implemented. Either implement it or drop the test; leaving it in gives wrong impression.

The ServiceMonitor deletion lifecycle is covered three separate times across `test_evalhub_servicemonitor_lifecycle.py`, `test_evalhub_servicemonitor_e2e.py`, and indirectly in the cleanup E2E test , this could be together. Also, a couple of places use `time.sleep(10)` as a reconciliation wait, which might be fragile. `TimeoutSampler` is already used elsewhere in the file and would be more reliable here. The `evalhub_service_monitor` fixture in `conftest.py` also uses a label-selector scan and returns `service_monitors[0]`, which is fine but a little loose: [#1458 constructs it by deterministic name](https://github.com/opendatahub-io/opendatahub-tests/pull/1458/files#diff-f3a7b0c1d4e7f0a3b6c9d2e5f8a1b4c7d0e3f6a9b2c5d8e1f4a7b0c3d6e9f2) (`f"{evalhub_cr.name}{EVALHUB_SERVICE_MONITOR_SUFFIX}"`), which is harder to accidentally match on the wrong object.

dbasunag:
> @fege My feedback on this PR:
> 
> There are syntax errors that prevent collection: [`test_evalhub_servicemonitor_e2e.py`](https://github.com/opendatahub-io/opendatahub-tests/pull/1464/files#diff-e36b6694c5f7a3e10be03cb0ce4c7e96a9c7c0af1b31e8b42a7bffef4a7c0a57) and [`test_evalhub_servicemonitor_prometheus.py`](https://github.com/opendatahub-io/opendatahub-tests/pull/1464/files#diff-e8a8b7e9d5d3e0b6f2b1c4a7e3a5c8f9b2d6e1a4c7f0b3e6d9a2c5f8b1e4d7) both use Python 2-style bare-tuple `except` syntax:
> 
> ```python
> except json.JSONDecodeError, KeyError, IndexError, TypeError:
> ```
> 
> But this is a `SyntaxError` in Python 3, IIUC. The module will fail to import and no tests in it will run. The right way is:
> 
> ```python
> except (json.JSONDecodeError, KeyError, IndexError, TypeError):
> ```
> 
> The NetworkPolicy tests are testing the wrong thing. The PR trustyai-service-operator#713 _removed_ the operator-managed NetworkPolicy. [PR #1458](https://github.com/opendatahub-io/opendatahub-tests/pull/1458) has this correctly with `test_operator_does_not_create_networkpolicy`, to asserts the absence of any operator-created NetworkPolicy.
> 
> This PR goes the other way: [`TC-NP-001`](https://github.com/opendatahub-io/opendatahub-tests/pull/1464/files#diff-a9e3b0c1d7f8e2a5b4c6d1e0f3a7b2c5d8e1f4a7b0c3d6e9f2a5b8c1d4e7f0) asserts `evalhub_netpol is not None`, and [`TC-E2E-002`](https://github.com/opendatahub-io/opendatahub-tests/pull/1464/files#diff-e36b6694c5f7a3e10be03cb0ce4c7e96a9c7c0af1b31e8b42a7bffef4a7c0a57) does the same. On a real deployment these tests will always fail. The right thing to test here is what #1458 tests: that no NetworkPolicy blocks port 8443 from the monitoring namespace, and that the operator no longer creates one at all.
> 
> The TLS assertion is backwards: [`TC-CFG-001`](https://github.com/opendatahub-io/opendatahub-tests/pull/1464/files#diff-b8c2d5e7a1f4b7c0e3d6a9f2b5c8e1d4f7a0b3c6e9d2a5f8b1c4e7d0a3f6b9) asserts:
> 
> ```python
> assert tls_config.insecureSkipVerify is True
> ```
> 
> The ServiceMonitor uses an OpenShift service-CA ConfigMap (`{evalhub}-service-ca`) for full certificate validation, which means `insecureSkipVerify` should explicitly **not** be set. This is what [#1458's `test_servicemonitor_tls_no_insecure_skip`](https://github.com/opendatahub-io/opendatahub-tests/pull/1458/files#diff-f3a7b0c1d4e7f0a3b6c9d2e5f8a1b4c7d0e3f6a9b2c5d8e1f4a7b0c3d6e9f2) validates. As it is, TC-CFG-001 would pass on a misconfigured ServiceMonitor and fail on a correct one.
> 
> TC-RBAC-002 isn't doing what it claims: [The test body](https://github.com/opendatahub-io/opendatahub-tests/pull/1464/files#diff-a9e3b0c1d7f8e2a5b4c6d1e0f3a7b2c5d8e1f4a7b0c3d6e9f2a5b8c1d4e7f0) even says so in its own comments: it only checks that `ownerReferences` is set and that the namespace matches, which are covered elsewhere. The comment block at the end describes the test that would actually validate restricted user access (`oc auth can-i`, impersonation), but none of that is implemented. Either implement it or drop the test; leaving it in gives wrong impression.
> 
> The ServiceMonitor deletion lifecycle is covered three separate times across `test_evalhub_servicemonitor_lifecycle.py`, `test_evalhub_servicemonitor_e2e.py`, and indirectly in the cleanup E2E test , this could be together. Also, a couple of places use `time.sleep(10)` as a reconciliation wait, which might be fragile. `TimeoutSampler` is already used elsewhere in the file and would be more reliable here. The `evalhub_service_monitor` fixture in `conftest.py` also uses a label-selector scan and returns `service_monitors[0]`, which is fine but a little loose: [#1458 constructs it by deterministic name](https://github.com/opendatahub-io/opendatahub-tests/pull/1458/files#diff-f3a7b0c1d4e7f0a3b6c9d2e5f8a1b4c7d0e3f6a9b2c5d8e1f4a7b0c3d6e9f2) (`f"{evalhub_cr.name}{EVALHUB_SERVICE_MONITOR_SUFFIX}"`), which is harder to accidentally match on the wrong object.



> @fege My feedback on this PR:
> 
> There are syntax errors that prevent collection: [`test_evalhub_servicemonitor_e2e.py`](https://github.com/opendatahub-io/opendatahub-tests/pull/1464/files#diff-e36b6694c5f7a3e10be03cb0ce4c7e96a9c7c0af1b31e8b42a7bffef4a7c0a57) and [`test_evalhub_servicemonitor_prometheus.py`](https://github.com/opendatahub-io/opendatahub-tests/pull/1464/files#diff-e8a8b7e9d5d3e0b6f2b1c4a7e3a5c8f9b2d6e1a4c7f0b3e6d9a2c5f8b1e4d7) both use Python 2-style bare-tuple `except` syntax:
> 
> ```python
> except json.JSONDecodeError, KeyError, IndexError, TypeError:
> ```
> 
> But this is a `SyntaxError` in Python 3, IIUC. The module will fail to import and no tests in it will run. The right way is:
> 
> ```python
> except (json.JSONDecodeError, KeyError, IndexError, TypeError):
> ```
> 
> The NetworkPolicy tests are testing the wrong thing. The PR trustyai-service-operator#713 _removed_ the operator-managed NetworkPolicy. [PR #1458](https://github.com/opendatahub-io/opendatahub-tests/pull/1458) has this correctly with `test_operator_does_not_create_networkpolicy`, to asserts the absence of any operator-created NetworkPolicy.
> 
> This PR goes the other way: [`TC-NP-001`](https://github.com/opendatahub-io/opendatahub-tests/pull/1464/files#diff-a9e3b0c1d7f8e2a5b4c6d1e0f3a7b2c5d8e1f4a7b0c3d6e9f2a5b8c1d4e7f0) asserts `evalhub_netpol is not None`, and [`TC-E2E-002`](https://github.com/opendatahub-io/opendatahub-tests/pull/1464/files#diff-e36b6694c5f7a3e10be03cb0ce4c7e96a9c7c0af1b31e8b42a7bffef4a7c0a57) does the same. On a real deployment these tests will always fail. The right thing to test here is what #1458 tests: that no NetworkPolicy blocks port 8443 from the monitoring namespace, and that the operator no longer creates one at all.
> 
> The TLS assertion is backwards: [`TC-CFG-001`](https://github.com/opendatahub-io/opendatahub-tests/pull/1464/files#diff-b8c2d5e7a1f4b7c0e3d6a9f2b5c8e1d4f7a0b3c6e9d2a5f8b1c4e7d0a3f6b9) asserts:
> 
> ```python
> assert tls_config.insecureSkipVerify is True
> ```
> 
> The ServiceMonitor uses an OpenShift service-CA ConfigMap (`{evalhub}-service-ca`) for full certificate validation, which means `insecureSkipVerify` should explicitly **not** be set. This is what [#1458's `test_servicemonitor_tls_no_insecure_skip`](https://github.com/opendatahub-io/opendatahub-tests/pull/1458/files#diff-f3a7b0c1d4e7f0a3b6c9d2e5f8a1b4c7d0e3f6a9b2c5d8e1f4a7b0c3d6e9f2) validates. As it is, TC-CFG-001 would pass on a misconfigured ServiceMonitor and fail on a correct one.
> 
> TC-RBAC-002 isn't doing what it claims: [The test body](https://github.com/opendatahub-io/opendatahub-tests/pull/1464/files#diff-a9e3b0c1d7f8e2a5b4c6d1e0f3a7b2c5d8e1f4a7b0c3d6e9f2a5b8c1d4e7f0) even says so in its own comments: it only checks that `ownerReferences` is set and that the namespace matches, which are covered elsewhere. The comment block at the end describes the test that would actually validate restricted user access (`oc auth can-i`, impersonation), but none of that is implemented. Either implement it or drop the test; leaving it in gives wrong impression.
> 
> The ServiceMonitor deletion lifecycle is covered three separate times across `test_evalhub_servicemonitor_lifecycle.py`, `test_evalhub_servicemonitor_e2e.py`, and indirectly in the cleanup E2E test , this could be together. Also, a couple of places use `time.sleep(10)` as a reconciliation wait, which might be fragile. `TimeoutSampler` is already used elsewhere in the file and would be more reliable here. The `evalhub_service_monitor` fixture in `conftest.py` also uses a label-selector scan and returns `service_monitors[0]`, which is fine but a little loose: [#1458 constructs it by deterministic name](https://github.com/opendatahub-io/opendatahub-tests/pull/1458/files#diff-f3a7b0c1d4e7f0a3b6c9d2e5f8a1b4c7d0e3f6a9b2c5d8e1f4a7b0c3d6e9f2) (`f"{evalhub_cr.name}{EVALHUB_SERVICE_MONITOR_SUFFIX}"`), which is harder to accidentally match on the wrong object.

@ruivieira we use python 3.14, and it supports this exception syntax: https://peps.python.org/pep-0758/

mwaykole (commented):
## PR Review: EvalHub ServiceMonitor Automated Tests (RHAISTRAT-1507)

**Risk Level: 🔴 HIGH — Multiple blockers that will cause hard failures on any real cluster; the PR cannot be merged in its current state.**

Thanks for the contribution, @fege — the test plan coverage is broad and the structure is well-organized. That said, there are several critical correctness issues that need to be addressed before this can merge. I'm flagging them in priority order. @ruivieira and @dbasunag have already called out most of the blockers; this review adds additional detail and a few new findings.

---

### 🔴 BLOCKER 1 — Python 2-style `except` syntax (SyntaxError in Python 3)

**Files:** `test_evalhub_servicemonitor_e2e.py` (line ~142), `test_evalhub_servicemonitor_prometheus.py`

```python
# Current (invalid Python 3 — SyntaxError at import time, ALL tests in file fail to collect)
except json.JSONDecodeError, KeyError, IndexError, TypeError:

# Required
except (json.JSONDecodeError, KeyError, IndexError, TypeError):
```

This is a hard syntax error. Python 3 does not allow comma-separated exception types without parentheses. The module fails to import entirely, so **no test in either file will run** — not just the affected test function. This was already noted by @ruivieira; it still needs to be fixed.

---

### 🔴 BLOCKER 2 — NetworkPolicy tests assert the wrong behavior (inverted logic)

**Files:** `test_evalhub_servicemonitor_security.py` (`TC-NP-001`), `test_evalhub_servicemonitor_e2e.py` (`TC-E2E-002`)

The upstream operator PR [trustyai-service-operator#713](https://github.com/trustyai-explainability/trustyai-service-operator/pull/713) **removed** operator-managed NetworkPolicy. These tests assert the opposite — that the operator *creates* a NetworkPolicy:

```python
# TC-NP-001 (security.py) — will always fail on a correct deployment
assert evalhub_network_policy is not None, "NetworkPolicy not created for EvalHub metrics scraping"

# TC-E2E-002 (e2e.py) — same inverted logic
assert evalhub_netpol is not None, "NetworkPolicy not created for EvalHub metrics scraping"
```

Compare with what PR #1458 correctly tests:
```python
# Correct — asserts absence of operator-managed NetworkPolicy
def test_operator_does_not_create_networkpolicy(...)
def test_no_networkpolicy_blocks_metrics_port_from_monitoring(...)
```

The `TC-NP-002` ("Prometheus can scrape in default-deny namespace") test is also flawed: it asserts `assert network_policies, ...` — i.e., it requires a NetworkPolicy to exist for the test to be "meaningful." But the correct behavior after #713 is that no operator-managed NetworkPolicy should exist. The test then falls back to checking ServiceMonitor config only (which is covered elsewhere), making `TC-NP-002` redundant.

**Suggested fix:** Replace `TC-NP-001` and `TC-E2E-002` with:
1. An assertion that no operator-created NetworkPolicy exists (matching `evalhub` in name).
2. An assertion that nothing blocks port 8443 from `openshift-user-workload-monitoring` (what #1458's `test_no_networkpolicy_blocks_metrics_port_from_monitoring` already does correctly).

---

### 🔴 BLOCKER 3 — TLS assertion is backwards (`TC-CFG-001`)

**File:** `test_evalhub_servicemonitor_lifecycle.py`

```python
# Current — PASSES on misconfigured ServiceMonitor, FAILS on correct one
assert tls_config.insecureSkipVerify is True
```

The ServiceMonitor uses an OpenShift service-CA ConfigMap (`{evalhub-name}-service-ca`) for full certificate validation. `insecureSkipVerify` should **not** be set. This is precisely what PR #1458's `test_servicemonitor_tls_no_insecure_skip` validates:

```python
# Correct (from #1458)
insecure = tls.get("insecureSkipVerify") if isinstance(tls, dict) else getattr(tls, "insecureSkipVerify", None)
assert not insecure, "insecureSkipVerify must not be set — EvalHub uses service CA for TLS validation"
```

There is also a missing assertion: `TC-CFG-001` should verify the `caFile`/`ca.configMap` points to the `{evalhub-name}-service-ca` ConfigMap, which is the actual security contract. See `test_servicemonitor_tls_ca_configmap` in #1458.

---

### 🔴 BLOCKER 4 — `TC-RBAC-002` is a no-op stub that gives a false sense of coverage

**File:** `test_evalhub_servicemonitor_security.py`

The test's own docstring says it:

> "NOTE: Full RBAC validation would require: 1. Creating a test user with limited permissions... 2. Using `oc auth can-i`... 3. Attempting actual delete/patch operations..."

What the test *actually* does is re-check `ownerReferences` and namespace equality — both of which are already covered by `TC-SM-002` and `TC-RBAC-003`. This test provides zero additional signal and actively misleads readers into thinking RBAC restriction has been validated.

**Options:**
- Implement actual RBAC validation using `oc auth can-i --as=<serviceaccount>` (preferred, since this is an E2E test suite against a real cluster).
- Drop the test entirely. A stub that appears to test something but doesn't is worse than no test.

---

### 🟡 CONCERN 5 — `evalhub_service_monitor` fixture uses loose label-selector lookup

**File:** `conftest.py`

```python
# Current — returns service_monitors[0] which could be any EvalHub's SM in a noisy namespace
service_monitors = list(ServiceMonitor.get(..., label_selector=f"app={EVALHUB_APP_LABEL}"))
return service_monitors[0]
```

PR #1458 constructs the ServiceMonitor by deterministic name:
```python
name=f"{evalhub_cr.name}{EVALHUB_SERVICE_MONITOR_SUFFIX}"
```

This is strictly safer. If other EvalHub instances exist in the namespace (as some tests in this very PR create them), `service_monitors[0]` may return the wrong ServiceMonitor. Consider adopting the naming-based approach and importing `EVALHUB_SERVICE_MONITOR_SUFFIX` from constants (which PR #1458 adds).

---

### 🟡 CONCERN 6 — `time.sleep()` used for reconciliation waits

**Files:** `test_evalhub_servicemonitor_e2e.py` (line ~199), `test_evalhub_servicemonitor_lifecycle.py` (TC-SM-005)

```python
time.sleep(10)  # TC-E2E-002 post-deployment wait
time.sleep(10)  # TC-SM-005 post-deployment wait
```

`TimeoutSampler` is already used in this file and is the project-standard pattern for polling. A fixed 10-second sleep is both fragile (too short on a slow cluster) and wasteful (always waits 10s even when the resource is already ready). Replace with:
```python
for _ in TimeoutSampler(wait_timeout=30, sleep=2, func=lambda: bool(list(ServiceMonitor.get(...)))):
    if list(ServiceMonitor.get(...)):
        break
```

---

### 🟡 CONCERN 7 — Duplication of ServiceMonitor deletion lifecycle across three test files

The GC/deletion lifecycle is tested in:
- `TC-SM-003` in `test_evalhub_servicemonitor_lifecycle.py`
- `TC-E2E-003` in `test_evalhub_servicemonitor_e2e.py`  
- (Implicitly) in the cleanup step of `TC-E2E-002`

These are essentially the same assertion. Consider collapsing the deletion tests into a single location and having the E2E test focus on end-to-end Prometheus scrape validation rather than repeating GC coverage.

---

### 🟡 CONCERN 8 — `TC-SM-004` may not actually test what it claims

**File:** `test_evalhub_servicemonitor_lifecycle.py`

```python
# Trigger reconcile by adding a metadata label
evalhub_cr.update(resource_dict={"metadata": {"labels": current_labels}})
# Then waits for ServiceMonitor resourceVersion to change
```

Adding a label to `metadata` of the EvalHub CR may or may not trigger the operator to update the ServiceMonitor, depending on operator implementation. The operator likely reconciles on Spec changes, not arbitrary metadata changes. If the operator only watches `.spec`, this test will reliably time out. Verify against the operator source that this label trigger actually causes a ServiceMonitor update, or use a `.spec` field change instead.

---

### 🟡 CONCERN 9 — `TC-SCRAPE-003` uses `time.sleep(35)` to validate scrape interval

**File:** `test_evalhub_servicemonitor_prometheus.py`

```python
time.sleep(35)  # Wait for next scrape
assert 25 <= scrape_delta_seconds <= 40
```

This adds 35 seconds of unconditional wall time to every CI run that includes this test. Given the test is `@pytest.mark.tier1`, it will run frequently. The tolerance window (±5s) is also quite tight for a loaded CI cluster. Consider using a wider window (±10s) or checking `lastScrapeError` is empty rather than timing the interval precisely.

---

### 🟢 POSITIVE OBSERVATIONS

- Good use of `TimeoutSampler` throughout for polling (except where `time.sleep` slips in).
- Test docstrings are detailed and reference the test plan IDs — easy to trace back to the strategy.
- `TC-SM-001`, `TC-SM-002` endpoint config assertions are well-structured with individual `assert` statements and clear messages.
- `TC-RBAC-001` (ClusterRole RBAC check) is solid and correctly validates required verbs.
- `TC-RBAC-003` (namespace isolation via two concurrent namespaces) is a well-designed test.
- The upgrade pre/post test structure (`@pytest.mark.pre_upgrade` / `@pytest.mark.post_upgrade`) follows established project conventions.
- The E2E happy path (`TC-E2E-001`) gracefully skips when user workload monitoring is not enabled, which is the right degradation behavior.

---

### Summary of Required Changes Before Merge

| # | Severity | Change Required |
|---|----------|-----------------|
| 1 | 🔴 Blocker | Fix Python 2 `except` syntax → `except (A, B, C):` |
| 2 | 🔴 Blocker | Invert NetworkPolicy tests — assert absence of operator-managed NetworkPolicy |
| 3 | 🔴 Blocker | Invert TLS assertion — `insecureSkipVerify` must NOT be `True` |
| 4 | 🔴 Blocker | Remove or fully implement `TC-RBAC-002` — current body doesn't test claimed behavior |
| 5 | 🟡 Concern | Use deterministic ServiceMonitor name in fixture |
| 6 | 🟡 Concern | Replace `time.sleep(10)` waits with `TimeoutSampler` |
| 7 | 🟡 Concern | Consolidate duplicated GC/deletion tests |
| 8 | 🟡 Concern | Verify that metadata label change actually triggers operator reconcile for TC-SM-004 |
| 9 | 🟡 Concern | Widen the scrape interval timing window in TC-SCRAPE-003 |

Also note: PR #1458 covers overlapping test cases. It would be worth coordinating with @dbasunag/@ruivieira to decide which PR carries which test cases, to avoid duplicate coverage landing on `main`.

PR #1470, opened 2026-04-28
fix: use python3 instead of python in OVMS smoke pod command

rpancham:
Addresses : https://redhat.atlassian.net/browse/RHOAIENG-60144

## Release Notes

* **Tests**
  * Updated smoke test execution environment to use Python 3.


Raghul-M (approved):

brettmthompson (approved):
/lgtm

PR #1476, opened 2026-04-28
feat: add upstream image registry support to image validation

sheltoncyril:
## Summary
- Add `upstream` parameter to `validate_image_format` and `validate_container_images` in `utilities/general.py` to support `quay.io/opendatahub` registry validation for ODH deployments
- Update `validate_trustyai_operator_image` to pass distribution flag through to image format validation
- All changes are backward-compatible — existing callers default to `upstream=False` (registry.redhat.io)

## Test plan
- [ ] Verify downstream (RHOAI) clusters still validate against `registry.redhat.io`
- [ ] Verify upstream (ODH) clusters validate against `quay.io/opendatahub`
- [ ] Verify existing callers of `validate_image_format` and `validate_container_images` are unaffected

## Summary by CodeRabbit

* **Tests**
  * Enhanced test infrastructure to support validation of container images against multiple registry sources, enabling more flexible testing configurations.

* **Chores**
  * Updated image validation utilities to support configurable registry validation paths for improved deployment flexibility.

### DIFF
--- a/tests/model_explainability/trustyai_operator/utils.py
+++ b/tests/model_explainability/trustyai_operator/utils.py
@@ -4,18 +4,22 @@ from utilities.general import validate_image_format
 
 
 def validate_trustyai_operator_image(
-    related_images_refs: set[str], tai_operator_configmap_data: dict[str, str], tai_operator_deployment: Deployment
+    related_images_refs: set[str],
+    tai_operator_configmap_data: dict[str, str],
+    tai_operator_deployment: Deployment,
+    upstream: bool = False,
 ) -> None:
     """Validates the TrustyAI operator image.
     Checks if:
         - container image matches that of the operator configmap.
         - image is present in relatedImages of CSV.
-        - image complies with OpenShift AI requirements i.e. sourced from registry.redhat.io and pinned w/o tags.
+        - image is sourced from the expected registry and pinned w/o tags.
 
         Args:
-            related_images_refs (set[str]): set of related image refs from the RHOAI CSV
+            related_images_refs (set[str]): set of related image refs from the CSV
             tai_operator_configmap_data (dict[str, str]): TrustyAI configmap data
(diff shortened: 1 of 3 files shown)


kpunwatk (approved):

tests/model_explainability/trustyai_operator/utils.py
```diff
@@ -26,5 +30,5 @@ def validate_trustyai_operator_image(
     tai_operator_image = tai_operator_deployment.instance.spec.template.spec.containers[0].image
     assert tai_operator_image == tai_operator_configmap_data["trustyaiOperatorImage"]
     assert tai_operator_image in related_images_refs
-    image_valid, error_message = validate_image_format(image=tai_operator_image)
+    image_valid, error_message = validate_image_format(image=tai_operator_image, upstream=upstream)
```
dbasunag, line 33: Why not make it downstream_only instead? 
Since the images could be either quay or registry.redhat, it does not make sense to run this test against odh.

dbasunag (changes requested):

tests/model_explainability/trustyai_operator/utils.py
sheltoncyril: I did not think about that, thanks for bringing it up but for ODH, I was under the assumption that we have to fully source all images from quay.io/opendatahub as there's a separate Konflux for ODH, right? are we allowed to have images pulled from registry? I do think this PR might be unnecessary incase someone decides to have their own repo on docker or ghcr. I can close this in that case @dbasunag

dbasunag: If a RHOAI build has quay image references that would be a blocker bug - it must always have registry.redhat references, hence we added the image validation tests across components. However for odh I am not sure if having a non-quay reference is considered bug. They typically should have quay image references. 

I agree this PR is not necessary.

PR #1480, opened 2026-04-29
fix: Set static vector store chunking strategy

jgarciao:
This change is needed in order to use the embedding model nomic-ai/nomic-embed-text-v2-moe, as this model has a maximum sequence length of 512 tokens, which is smaller than the default chunking size used in auto mode (800 tokens)


jiripetrlik (approved):

Bobbins228 (approved):
/lgtm

PR #1495, opened 2026-04-30
fix: set chunking strategy to 384/64 and modify finance dataset

jgarciao:
Some vector store tests were failing after modifying the embedding model to Nomic-embed-text-v2-moe. These changes increase the test success ratio

## Summary by CodeRabbit

* **Tests**
  * Refined financial Q&A test data by shortening and generalizing question phrasing across quarterly entries to improve clarity and consistency.
  * Optimized vector storage chunking parameters used in test processing to improve efficiency of test data handling and indexing.

PR #1497, opened 2026-04-30
fix: set chunking strategy to 384/64 and modify keyword questions

jgarciao:
Some vector store tests were failing after modifying the embedding model to Nomic-embed-text-v2-moe. These changes increase the test success ratio

This is a backport of #1495


Bobbins228 (commented):
/lgtm

Ygnas (approved):

PR #1511, opened 2026-05-07
fix: use dedicated model_with_benchmark_metadata to reduce test flakiness

dbasunag:
# Pull Request

## Summary
Once in a while a random model without benchmark data is getting picked up by randomly_picked_model_from_catalog_api_by_source and causing test failure

## Related Issues

- Fixes: 
- JIRA: 

- [x] Locally
- [ ] Jenkins

## Additional Requirements

## Summary by CodeRabbit

* **Tests**
  * Added a shared class-scoped fixture to select and validate catalog models that include benchmark/metadata, improving test coverage and logging of discovery events.
  * Updated a metadata validation test to use the new fixture for more reliable inputs.
  * Improved benchmark metadata path resolution to handle case-insensitive model name segments, reducing test flakiness when locating metadata.

### DIFF
--- a/tests/model_registry/model_catalog/metadata/conftest.py
+++ b/tests/model_registry/model_catalog/metadata/conftest.py
@@ -29,6 +29,8 @@ def expected_labels_by_asset_type(
     return [label for label in all_labels if label.get("assetType") == asset_type]
 
 
+# TODO: RHOAIENG-62057 - Remove this fixture and use randomly_picked_model_from_catalog_api_by_source
+# once the bug is fixed.
 @pytest.fixture(scope="class")
 def model_with_benchmark_metadata(
     model_catalog_pod: Pod,
--- a/tests/model_registry/model_catalog/metadata/conftest.py
+++ b/tests/model_registry/model_catalog/metadata/conftest.py
@@ -1,9 +1,20 @@
+import random
 from typing import Any
 
 import pytest
+import structlog
 from kubernetes.dynamic import DynamicClient
+from ocp_resources.pod import Pod
 
+from tests.model_registry.model_catalog.constants import (
+    CATALOG_CONTAINER,
+    PERFORMANCE_DATA_DIR,
+    VALIDATED_CATALOG_ID,
+)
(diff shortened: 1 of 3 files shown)


tests/model_registry/model_catalog/metadata/conftest.py
```diff
@@ -16,3 +27,43 @@ def expected_labels_by_asset_type(
     asset_type = request.param
     all_labels = get_labels_from_configmaps(admin_client=admin_client, namespace=model_registry_namespace)
     return [label for label in all_labels if label.get("assetType") == asset_type]
+
+
+@pytest.fixture(scope="class")
+def model_with_benchmark_metadata(
```
fege, line 35: probably I missed something, but I thought that all the downstream models have this metadata. is something changed? if my supposition is correct we could simply use `@dowstream_only` on the test

lugi0: Only the validated models, but I think this could be simplified to just pick from that source?

dbasunag: Right now all the validated models don't have benchmark data.

tests/model_registry/model_catalog/metadata/utils.py
```diff
@@ -273,6 +273,25 @@ def validate_custom_properties_match_metadata(api_custom_properties: dict[str, s
     return True
 
 
+def get_benchmark_path(model_catalog_pod: Pod, model_name: str) -> str:
+    """Resolve the metadata.json path with case-insensitive directory matching."""
+    base_dir = "/shared-benchmark-data"
```
fege, line 278: what do you think about adding this as input param?

lugi0 (approved):

tests/model_registry/model_catalog/metadata/utils.py
dbasunag: Since this is the base directory, I don't see a point of making it a input param, which would them become customizable and this path is fixed.

fege (approved):
/lgtm

PR #1514, opened 2026-05-08
Add tests for MaaS controller bundle cleanup

SB159:
# Pull Request

## Summary

Add tests for MaaS controller bundle cleanup

## Related Issues

- Fixes: 
- JIRA: 

- [ ] Locally
- [ ] Jenkins

## Additional Requirements

## Summary by CodeRabbit

* **Tests**
  * Added a new test suite validating MaaS disable cleanup behavior.
  * Verifies the MaaS controller deployment is removed after disable.
  * Confirms cluster-scoped RBAC and service account for the controller are removed.
  * Asserts no controller bundle resources remain present after cleanup.


tests/model_serving/maas_billing/maas_cleanup/conftest.py
```diff
@@ -0,0 +1,29 @@
...
+
+    Depends on maas_controller_enabled_latest so MaaS is guaranteed to be running
+    before the disable is triggered, matching the real admin workflow that the bug
+    reproduces.
+    """
+    component_patch = {
+        DscComponents.KSERVE: {"modelsAsService": {"managementState": DscComponents.ManagementState.REMOVED}}
+    }
+    with ResourceEditor(patches={dsc_resource: {"spec": {"components": component_patch}}}):
+        dsc_resource.wait_for_condition(condition="Ready", status="True", timeout=600)
```
dbasunag: You need to add a check here so that you don't call dsc ready, before the changes are actually applied. So I would suggest for check for the time it was updated last, before the with block, then wait for ready state and transition time(or similar, I don't recall the name) should be > previously saved time

tests/model_serving/maas_billing/maas_cleanup/test_maas_disable_cleanup.py
```diff
@@ -0,0 +1,95 @@
...
+        applications_namespace = py_config["applications_namespace"]
+        controller_deployment = Deployment(
+            client=admin_client,
+            name=MAAS_CONTROLLER_RESOURCE_NAME,
+            namespace=applications_namespace,
+        )
+        assert controller_deployment.wait_deleted(timeout=600), (
+            f"maas-controller Deployment still exists in namespace '{applications_namespace}'"
+            f" after managementState set to Removed"
+        )
```
dbasunag: This might be flaky. I would suggest check if the deployment exist before waiting.

tests/model_serving/maas_billing/maas_cleanup/test_maas_disable_cleanup.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def test_disable_maas_removes_crds(
+        self,
+        admin_client: DynamicClient,
+    ) -> None:
+        """Verify all MaaS CRDs are removed after managementState is set to Removed."""
+        surviving_crds = [
+            crd_name for crd_name in MAAS_CRDS if CustomResourceDefinition(client=admin_client, name=crd_name).exists
+        ]
+        assert not surviving_crds, (
```
dbasunag: you should use all(). surviving_crds would be a list of strings or something. It may never be empty

tests/model_serving/maas_billing/maas_cleanup/test_maas_disable_cleanup.py
```diff
@@ -0,0 +1,95 @@
...
+            f"ServiceAccount/{MAAS_CONTROLLER_RESOURCE_NAME}": ServiceAccount(
+                client=admin_client,
+                name=MAAS_CONTROLLER_RESOURCE_NAME,
+                namespace=applications_namespace,
+            ),
+        }
+        surviving_rbac = [label for label, resource in rbac_resources.items() if resource.exists]
+        assert not surviving_rbac, (
+            f"MaaS RBAC resources still present after managementState set to Removed: {', '.join(surviving_rbac)}"
+        )
```
dbasunag, line 58: Same comment.

dbasunag (changes requested):

tests/model_serving/maas_billing/maas_cleanup/conftest.py
SB159: Done.

tests/model_serving/maas_billing/maas_cleanup/test_maas_disable_cleanup.py
SB159: Done

tests/model_serving/maas_billing/maas_cleanup/test_maas_disable_cleanup.py
SB159: Done

tests/model_serving/maas_billing/maas_cleanup/test_maas_disable_cleanup.py
SB159: Done

dbasunag (dismissed):

mwaykole (commented):
## PR Review — `feat/rhoaieng-61344-maas-disable-cleanup`

**Reviewer:** mwaykole (autonomous dev-engineer agent)
**Risk Level:** 🟡 MEDIUM — functional gaps could cause false-positive test results and incomplete cleanup verification

---

### Summary

This PR adds a new test suite (`TestMaaSDisableCleanup`) to verify that setting `managementState: Removed` on `kserve.modelsAsService` in the DSC triggers full teardown of the MaaS controller bundle. The approach and fixture design are generally sound, but there are several correctness issues — one of which would cause a test to silently pass even when the resource was never cleaned up.

---

### 🔴 Critical Issues

#### 1. `test_disable_maas_removes_controller_deployment` — Silent Pass on Missing Deployment

**File:** `test_maas_disable_cleanup.py`, lines 30–43

```python
# CURRENT — silently passes if deployment never existed
if controller_deployment.exists:
    assert controller_deployment.wait_deleted(timeout=600), ...
```

If the `maas-controller` Deployment does **not** exist when the test runs (e.g. it was never deployed, already deleted by a previous test, or DSC never reconciled far enough to create it), this test passes with zero assertions. A cleanup test that validates removal must first **assert that the resource existed** before checking deletion. Consider:

```python
# RECOMMENDED — assert it is gone; failure reveals whether the deployment existed at all
assert not controller_deployment.exists, (
    f"maas-controller Deployment still exists in namespace '{applications_namespace}' "
    f"after managementState set to Removed"
)
```

If the resource might legitimately be absent (e.g. because upstream cleanup races the test), add a `wait_deleted` unconditionally or document the race explicitly with a skip/xfail.

---

#### 2. Missing CRD — `externalmodels.maas.opendatahub.io`

**File:** `test_maas_disable_cleanup.py`, lines 16–21

```python
MAAS_CRDS: tuple[str, ...] = (
    f"maasmodelrefs.{ApiGroups.MAAS_IO}",
    f"maasauthpolicies.{ApiGroups.MAAS_IO}",
    f"maassubscriptions.{ApiGroups.MAAS_IO}",
    f"tenants.{ApiGroups.MAAS_IO}",
)
```

Cross-referencing against the actual MaaS bundle at `repos/midstream/models-as-a-service/deployment/base/maas-controller/crd/bases/`, the bundle ships **five** CRDs:
- `maas.opendatahub.io_externalmodels.yaml` ← **not in MAAS_CRDS**
- `maas.opendatahub.io_maasauthpolicies.yaml` ✓
- `maas.opendatahub.io_maasmodelrefs.yaml` ✓
- `maas.opendatahub.io_maassubscriptions.yaml` ✓
- `maas.opendatahub.io_tenants.yaml` ✓

The `externalmodels.maas.opendatahub.io` CRD is silently missing from the cleanup check. If the operator fails to delete it, this PR's tests will not catch the regression. Please add:

```python
MAAS_CRDS: tuple[str, ...] = (
    f"externalmodels.{ApiGroups.MAAS_IO}",    # ← add this
    f"maasmodelrefs.{ApiGroups.MAAS_IO}",
    f"maasauthpolicies.{ApiGroups.MAAS_IO}",
    f"maassubscriptions.{ApiGroups.MAAS_IO}",
    f"tenants.{ApiGroups.MAAS_IO}",
)
```

---

#### 3. Missing RBAC Names — Cluster & Namespaced Resources

**File:** `test_maas_disable_cleanup.py`, lines 55–76 and `utils.py`

The test and utility check only for a ClusterRole and ClusterRoleBinding named `maas-controller`. However, the actual bundle (`rbac/clusterrole.yaml`, `rbac/clusterrole_binding.yaml`) names them **`maas-controller-role`** and **`maas-controller-rolebinding`** respectively. If the hardcoded name `"maas-controller"` does not match the live cluster resources, the `.exists` checks will always return `False` — and all assertions will silently pass.

Additionally, the bundle includes **two additional aggregate ClusterRoles** not checked anywhere in this PR:
- `maas-owner-role` (`rbac/owner_role.yaml`)
- `maas-viewer-role` (`rbac/viewer_role.yaml`)

And **two namespaced RBAC resources** (missing entirely from both the test and `get_surviving_maas_controller_resources()`):
- `Role/maas-controller-leader-election-role` (namespace-scoped)
- `RoleBinding/maas-controller-leader-election-rolebinding` (namespace-scoped)

**Action required:**
1. Confirm the actual resource names deployed by RHOAI on a live cluster and align `MAAS_CONTROLLER_RESOURCE_NAME` (or expand to multiple constants).
2. Either add the missing aggregate ClusterRoles and namespaced Role/RoleBinding to `get_surviving_maas_controller_resources()`, or explicitly document why they are intentionally excluded (e.g. RHOAI operator leaves them behind by design).

---

### 🟡 Medium Issues

#### 4. TOCTOU Race in CRD Assertion vs Error Message

**File:** `test_maas_disable_cleanup.py`, lines 50–55

```python
assert all(not CustomResourceDefinition(client=admin_client, name=crd_name).exists for crd_name in MAAS_CRDS), (
    f"MaaS CRDs still present after managementState set to Removed: "
    f"{', '.join(crd for crd in MAAS_CRDS if CustomResourceDefinition(client=admin_client, name=crd).exists)}"
)
```

The assertion and the error message each call `.exists` separately, creating a TOCTOU window. A CRD deleted between the two calls will appear in the assertion as gone but not in the error string (or vice versa), producing a confusing failure message. Capture results once:

```python
surviving_crds = [crd for crd in MAAS_CRDS if CustomResourceDefinition(client=admin_client, name=crd).exists]
assert not surviving_crds, (
    f"MaaS CRDs still present after managementState set to Removed: {', '.join(surviving_crds)}"
)
```

The same pattern exists in `test_disable_maas_removes_cluster_rbac` (lines 58–76) — apply the same fix there.

---

#### 5. Missing `__init__.py` for New Package

**File:** `tests/model_serving/maas_billing/maas_cleanup/` (new directory)

All sibling test packages (`component_health/`, `external_model/`, `maas_api_key/`, `maas_subscription/`, `oidc_tests/`) include an `__init__.py`. The new `maas_cleanup/` package is missing one. While pytest may discover tests without it, the missing `__init__.py` breaks the project convention and can cause import conflicts if two test directories have a module with the same name.

---

#### 6. Fixture Teardown May Hang Without Timeout on Conditions

**File:** `conftest.py`, lines 26–28

```python
    dsc_resource.wait_for_condition(condition="ModelsAsServiceReady", status="True", timeout=900)
    dsc_resource.wait_for_condition(condition="Ready", status="True", timeout=600)
```

These are called in teardown (after the `with ResourceEditor` block exits). If the cluster is in a degraded state or the gateway dependencies (`maas_gateway_api`, `maas_request_ratelimit_policy`) are torn down before this fixture completes, `ModelsAsServiceReady=True` may never be reached and the teardown will block until timeout. Consider wrapping the teardown conditions in a `try/except` with structured logging to avoid masking the root failure when teardown itself fails:

```python
try:
    dsc_resource.wait_for_condition(condition="ModelsAsServiceReady", status="True", timeout=900)
    dsc_resource.wait_for_condition(condition="Ready", status="True", timeout=600)
except Exception as exc:  
    LOGGER.warning(f"DSC did not return to healthy state after MaaS re-enable: {exc}")
```

---

#### 7. Jira Ticket Missing from PR Description

**File:** PR description

The JIRA field in the PR template is unfilled (`<!-- Jira information -->`). The branch name `feat/rhoaieng-61344-maas-disable-cleanup` contains the Jira key `RHOAIENG-61344`. Please link it in the PR description so CI tooling and changelog automation can track it:

```markdown
- JIRA: [RHOAIENG-61344](https://issues.redhat.com/browse/RHOAIENG-61344)
```

---

### 🟢 What's Done Well

- The fixture pattern using `ResourceEditor` + `wait_for_dsc_reconciliation` + `baseline_time` correctly avoids false-positive "Ready" states immediately after a patch — a subtle and important correctness detail.
- Separating concerns between `conftest.py` (lifecycle), `test_*.py` (assertions), and `utils.py` (resource collection) is clean and follows the project's structure.
- Using `scope="class"` for the fixture is appropriate for destructive/lifecycle tests.
- `get_surviving_maas_controller_resources()` returning a list of `Kind/name` strings gives human-readable failure messages.
- Tier-3 marker is correct for destructive/negative tests.

---

### Required Before Merge

| # | Severity | Action |
|---|----------|--------|
| 1 | 🔴 | Fix silent-pass in `test_disable_maas_removes_controller_deployment` — unconditionally assert the resource is gone |
| 2 | 🔴 | Add `externalmodels.maas.opendatahub.io` to `MAAS_CRDS` |
| 3 | 🔴 | Confirm actual ClusterRole/ClusterRoleBinding names on cluster; add missing `maas-owner-role`, `maas-viewer-role`, and namespaced `Role`/`RoleBinding` to coverage |
| 4 | 🟡 | Fix TOCTOU in CRD + RBAC assert/error-message double-query |
| 5 | 🟡 | Add `__init__.py` to `tests/model_serving/maas_billing/maas_cleanup/` |
| 6 | 🟡 | Guard teardown `wait_for_condition` calls with `try/except` + logging |
| 7 | 🟡 | Fill in the Jira ticket number in the PR description |

mwaykole (commented):
## PR Review — `feat/rhoaieng-61344-maas-disable-cleanup` (Follow-Up)

**Reviewer:** mwaykole (autonomous dev-engineer agent)
**Risk Level:** 🟡 MEDIUM — two critical correctness issues remain unaddressed from my previous review; the PR should not merge until they are resolved
**Previous Review Posted:** 2026-05-14 (state: COMMENTED)
**PR Status:** `dbasunag` APPROVED on 2026-05-11 (before my review was posted)

---

### What's Been Fixed Since the First Pass ✅

1. **`wait_for_dsc_reconciliation` baseline-time guard** — `dbasunag`'s original comment addressed: `baseline_time` is now captured before the `with ResourceEditor` block, preventing false-positive "Ready" hits before reconciliation starts. ✅
2. **Deployment existence guard** — `dbasunag`'s original comment addressed: the deployment check now uses `if controller_deployment.exists` before calling `wait_deleted`. ✅
3. **CRD list now uses `all()`** — `dbasunag`'s original comment addressed. ✅

---

### 🔴 Critical Issues Still Open (from my 2026-05-14 review)

#### Issue 1 — `test_disable_maas_removes_controller_deployment` Can Silently Pass

**File:** `test_maas_disable_cleanup.py`, lines 36–43

```python
if controller_deployment.exists:
    assert controller_deployment.wait_deleted(timeout=600), (...)
```

This test passes with **zero assertions** if the `maas-controller` Deployment doesn't exist at the time the check runs — whether because:
- DSC reconciliation completed before the test ran (race condition), or
- The Deployment was never created in the first place, or
- A previous test in the same class already checked it

A cleanup test MUST unconditionally assert the resource is absent — not conditionally wait for deletion only if the resource happens to still be there at snapshot time. Replace with:

```python
assert not controller_deployment.exists, (
    f"maas-controller Deployment still exists in namespace '{applications_namespace}' "
    f"after managementState set to Removed"
)
```

This is the correct pattern since `wait_for_dsc_reconciliation` (called in the fixture before `yield`) already ensures the DSC has fully reconciled **before** any test method runs. At that point, the Deployment should already be gone — if it is not, the test must fail.

---

#### Issue 2 — TOCTOU in `test_disable_maas_removes_crds` (Double `.exists` Query)

**File:** `test_maas_disable_cleanup.py`, lines 48–54

```python
assert all(not CustomResourceDefinition(client=admin_client, name=crd_name).exists for crd_name in MAAS_CRDS), (
    f"MaaS CRDs still present after managementState set to Removed: "
    f"{', '.join(crd for crd in MAAS_CRDS if CustomResourceDefinition(client=admin_client, name=crd).exists)}"
)
```

The assertion generator and the error-message generator each make **separate API calls** to `.exists`. In a fast-deleting cluster, a CRD can disappear between the two calls, causing:
- The `all()` check to fail (CRD was present during the assertion pass), but
- The CRD name to be absent from the error string (already deleted by the time the message string is built)

This produces a confusing "MaaS CRDs still present: " (empty list) failure. Fix by capturing results once:

```python
surviving_crds = [crd for crd in MAAS_CRDS if CustomResourceDefinition(client=admin_client, name=crd).exists]
assert not surviving_crds, (
    f"MaaS CRDs still present after managementState set to Removed: {', '.join(surviving_crds)}"
)
```

The identical TOCTOU pattern exists in `test_disable_maas_removes_cluster_rbac` (the `all(not resource.exists ...)` + error-message loop) — apply the same fix there.

---

### 🟡 Medium Issues Still Open

#### Issue 3 — Missing `__init__.py`

**Directory:** `tests/model_serving/maas_billing/maas_cleanup/`

Every sibling package has one:
```
tests/model_serving/maas_billing/component_health/__init__.py  ✅
tests/model_serving/maas_billing/external_model/__init__.py    ✅
tests/model_serving/maas_billing/maas_api_key/__init__.py      ✅
tests/model_serving/maas_billing/maas_subscription/__init__.py ✅
tests/model_serving/maas_billing/oidc_tests/__init__.py        ✅
tests/model_serving/maas_billing/maas_cleanup/__init__.py      ❌ MISSING
```

Add an empty `__init__.py` to follow project convention and prevent potential import path conflicts.

---

#### Issue 4 — Fixture Name Violates Naming Convention (CodeRabbit also flagged this)

**File:** `conftest.py`, line 12

`maas_disabled_for_cleanup_test` is an **action-style name** (describes what the test does), not a **noun describing the state/resource** (as required by the project's coding guidelines: *"Fixture names MUST be nouns: `storage_secret` not `create_secret`"*).

Rename to `dsc_with_maas_disabled` (or similar) and update `@pytest.mark.usefixtures(...)` in the test class.

---

#### Issue 5 — Jira Ticket Not Linked in PR Description

The PR description template still has `<!-- Jira information -->` unfilled. The branch name contains `RHOAIENG-61344`. Please add:

```markdown
- JIRA: [RHOAIENG-61344](https://issues.redhat.com/browse/RHOAIENG-61344)
```

---

#### Issue 6 — `maas_controller_enabled_latest` Fixture Scope Mismatch

**File:** `conftest.py`, line 13 (fixture dependency chain)

`maas_disabled_for_cleanup_test` is `scope="class"`, but its dependency `maas_controller_enabled_latest` (defined in `tests/model_serving/maas_billing/conftest.py`) is `scope="session"`. A `class`-scoped fixture depending on a `session`-scoped fixture is valid pytest behavior, but **there is no `maas_gateway_api` or `maas_request_ratelimit_policy` guarding the MaaS infrastructure** in the new fixture — those were the two required prerequisites for `maas_controller_enabled_latest` to work reliably. If this test runs in isolation (e.g. on a cluster where no gateway is pre-configured), the `maas_controller_enabled_latest` fixture may not have properly bootstrapped MaaS before this fixture tries to set it to `Removed`, causing confusing failures.

At minimum, document in the fixture docstring that the test requires a pre-configured MaaS gateway, or add the gateway readiness check as a dependency.

---

### 🟢 What's Done Well (unchanged from previous review)

- `ResourceEditor` + `wait_for_dsc_reconciliation` + `baseline_time` approach is correct and avoids false-positive "Ready" states — this is a subtle and important detail done right.
- Separation of `conftest.py` (lifecycle), `test_*.py` (assertions), and `utils.py` (resource collection) is clean.
- `scope="class"` for a destructive fixture is appropriate.
- `get_surviving_maas_controller_resources()` returning `Kind/name` strings makes failures human-readable.
- Tier-3 marker is correct for destructive/negative lifecycle tests.

---

### Action Table

| # | Severity | File | Status | Required Action |
|---|----------|------|--------|-----------------|
| 1 | 🔴 Critical | `test_maas_disable_cleanup.py:36-43` | 🔴 Open | Remove `if exists` guard; unconditionally assert `not deployment.exists` |
| 2 | 🔴 Critical | `test_maas_disable_cleanup.py:48-76` | 🔴 Open | Fix TOCTOU double-query in CRD and RBAC assertion/error-message |
| 3 | 🟡 Medium | `maas_cleanup/__init__.py` | 🔴 Open | Add missing `__init__.py` |
| 4 | 🟡 Medium | `conftest.py:12` | 🔴 Open | Rename fixture to noun: `dsc_with_maas_disabled` |
| 5 | 🟡 Medium | PR description | 🔴 Open | Link `RHOAIENG-61344` Jira ticket |
| 6 | 🟡 Medium | `conftest.py:13` | 🔴 Open | Document or guard MaaS gateway prerequisite |

dbasunag (approved):

threcc (approved):

PR #1472, opened 2026-04-28
test: add TrustyAI DB storage upgrade tests

ssaleem-rh:
# Pull Request

## Summary

Adds upgrade tests for TrustyAI Service with database storage to verify data persistence across upgrades from 2.25 to 3.3.

### Test Coverage
- **TestPreUpgradeTrustyAIServiceDB**: Creates TrustyAI Service with MariaDB storage, sends inferences, uploads training data, and schedules drift metrics before upgrade  
- **TestPostUpgradeTrustyAIServiceDB**: Validates database credentials after upgrade, verifies pre-upgrade metrics persist, schedules new metrics, and confirms DB functionality post-upgrade  

### Fixture Updates
Modified 10 fixtures to support pre/post-upgrade lifecycle:
- **Pre-upgrade**: Creates resources with `teardown=False` to persist across upgrade  
- **Post-upgrade**: Reuses existing resources with `ensure_exists=True`, then performs cleanup  

### Utilities
- Added `validate_db_credentials_secret()` to verify required MariaDB credential keys  
- Updated `create_isvc_getter_*` functions to support a `teardown` parameter for flexible cleanup  

## Related Issues

- Fixes: 
- JIRA: (https://redhat.atlassian.net/browse/RHOAIENG-54756)

## Please review and indicate how it has been tested

- [x] Locally
- [ ] Jenkins

pre-upgrade(2.25):
image

post-upgrade(3.3):
image

## Additional Requirements

- [ ] If this PR introduces a new test image, did you create a PR to mirror it in disconnected environment?
- [ ] If this PR introduces new marker(s)/adds a new component, was relevant ticket created to update relevant Jenkins job?

## Summary by CodeRabbit

* **Tests**
  * Improved upgrade tests to verify DB-backed service state, credential persistence, and metric scheduling/cleanup across upgrades.
  * Added explicit validation of database credentials and stronger checks for required credential fields.
  * Ensured test resources are cleaned up deterministically after post-upgrade scenarios.
  * Expanded raw-deployment coverage to exercise DB-backed operations before and after upgrades.

PR #1527, opened 2026-05-12
Backport #1472: TrustyAI DB storage upgrade tests to 3.4

ssaleem-rh:
# Pull Request

## Summary

Backport of #1472 to the 3.4 branch.

Includes:

- TrustyAIService DB storage upgrade tests
- Database credential validation checks
- Pre/post-upgrade lifecycle handling updates
- Security improvements (dynamic DB password generation)

Manually resolved conflict in `conftest.py` when cherry-picking from main branch (commit 1d2e5a7).

## Related Issues

- Fixes: 
- JIRA: https://redhat.atlassian.net/browse/RHOAIENG-54756

 Tested locally on:
 
- pre-upgrade: 2.25
- post-upgrade: 3.3

- [x] Locally
- [ ] Jenkins

## Additional Requirements

### DIFF
--- a/tests/model_explainability/trustyai_service/conftest.py
+++ b/tests/model_explainability/trustyai_service/conftest.py
@@ -1,4 +1,5 @@
 import json
+import secrets
 from collections.abc import Generator
 from typing import Any
 
@@ -200,6 +201,7 @@ def db_credentials_secret(
         yield secret
         secret.clean_up()
     else:
+        db_password = secrets.token_urlsafe(nbytes=24)
         with Secret(
             client=admin_client,
             name=DB_CREDENTIALS_SECRET_NAME,
@@ -208,7 +210,7 @@ def db_credentials_secret(
                 "databaseKind": MARIADB,
                 "databaseName": DB_NAME,
                 "databaseUsername": DB_USERNAME,
-                "databasePassword": DB_PASSWORD,
+                "databasePassword": db_password,
                 "databaseService": MARIADB,
                 "databasePort": "3306",
                 "databaseGeneration": "update",
--- a/tests/model_explainability/trustyai_service/conftest.py
+++ b/tests/model_explainability/trustyai_service/conftest.py
@@ -411,14 +411,14 @@ def isvc_getter_service_account(
     cleanup.
     """
     if pytestconfig.option.post_upgrade:
-        sa = ServiceAccount(
+        service_account = ServiceAccount(
             client=admin_client,
             name=ISVC_GETTER,
             namespace=model_namespace.name,
             ensure_exists=True,
         )
-        yield sa
-        sa.clean_up()
+        yield service_account
+        service_account.clean_up()
     else:
         with create_isvc_getter_service_account(
             client=admin_client, namespace=model_namespace, name=ISVC_GETTER, teardown=teardown_resources
@@ -475,14 +475,14 @@ def isvc_getter_role_binding(
     and manages cleanup.
     """
     if pytestconfig.option.post_upgrade:
-        rb = RoleBinding(
+        role_binding = RoleBinding(
             client=admin_client,
             name=ISVC_GETTER,
             namespace=model_namespace.name,
             ensure_exists=True,
         )
-        yield rb
-        rb.clean_up()
+        yield role_binding
+        role_binding.clean_up()
     else:
         with create_isvc_getter_role_binding(
             client=admin_client,
--- a/tests/model_explainability/trustyai_service/conftest.py
+++ b/tests/model_explainability/trustyai_service/conftest.py
@@ -79,11 +79,17 @@ def trustyai_service(
     user_workload_monitoring_config: ConfigMap,
     teardown_resources: bool,
 ) -> Generator[TrustyAIService, Any, Any]:
+    """Provides a TrustyAIService instance for testing.
+
+    In post-upgrade mode, references the existing TrustyAIService created during pre-upgrade tests
+    and cleans it up after post-upgrade tests complete.
+
+    In pre-upgrade mode (or when no upgrade flag is set), creates a new TrustyAIService with either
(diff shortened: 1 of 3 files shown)


dbasunag (approved):

sheltoncyril (approved):

mwaykole (changes requested):
## PR Review — Backport #1472: TrustyAI DB Storage Upgrade Tests to 3.4

**Reviewer**: mwaykole (Autonomous Dev Engineer)
**Risk Level**: 🟡 MEDIUM — Test-only PR, no production code changed, but carries meaningful logic risks in fixture lifecycle and upgrade state assumptions.

---

### Summary

This is a clean, well-structured backport of #1472 (merged 2026-05-11) to the `3.4` branch. The author correctly cherry-picked the commit, manually resolved a `conftest.py` conflict, and the CI pipeline (pre-commit + tox + image build against `3.4`) passed green. Two project members (`dbasunag`, `sheltoncyril`) have already approved.

The PR adds:
- `TestPreUpgradeTrustyAIServiceDB` — pre-upgrade test class for TrustyAI with MariaDB storage
- `TestPostUpgradeTrustyAIServiceDB` — post-upgrade persistence validation class
- `validate_db_credentials_secret()` utility in `utils.py`
- `teardown` parameter propagation across all `create_isvc_getter_*` helper functions

The overall design is sound and follows the existing upgrade test pattern. Below are specific findings.

---

### 🔴 MUST-FIX — Blockers

#### 1. Dead constant `DB_PASSWORD` still present in `conftest.py`

The old code used `DB_PASSWORD: str = "trustyai_password"`. This PR correctly replaces it with `secrets.token_urlsafe(nbytes=24)` — a good security improvement. However, the constant `DB_PASSWORD` is **still defined at line 70 of `conftest.py`** and is now completely unused (no reference anywhere in the file).

This dead code will mislead future developers into thinking the hardcoded value is still active.

**Fix**: Remove the `DB_PASSWORD: str = "trustyai_password"` constant from `conftest.py`.

---

#### 2. `validate_db_credentials_secret()` checks raw base64, not decoded values

In `utils.py`, `secret.instance.data` returns **base64-encoded** values (standard Kubernetes Secret behavior). The emptiness check:

```python
empty_keys = {key for key in required_keys if not secret_data.get(key)}
```

...detects a missing/`null` key but **cannot detect a key whose actual decoded value is an empty string or whitespace**. For example, `" "` (a single space) base64-encodes to `"IA=="`, which is truthy — so a whitespace-only password would silently pass validation.

**Fix**: Decode before checking:
```python
import base64

empty_keys = {
    key for key in required_keys
    if not base64.b64decode(secret_data.get(key, "")).decode("utf-8").strip()
}
```
This ensures the actual credential value (not just its base64 payload) is non-empty and usable.

---

### 🟡 SHOULD-FIX — Meaningful Gaps

#### 3. No cross-class dependency chain between pre- and post-upgrade DB tests

`test_trustyai_service_db_pre_upgrade_drift_metric_schedule_meanshift` has a `depends` but **no `name`** tag:

```python
@pytest.mark.dependency(depends=["db_pre_upgrade_data_upload"])
# ↑ no name= here
```

This means `test_trustyai_service_db_post_upgrade_preexisting_metric_can_be_deleted` cannot declare a dependency on whether the metric was actually scheduled. If the pre-upgrade drift test fails, the post-upgrade test will attempt to delete a metric that was never scheduled, producing a confusing and misleading failure.

**Recommendation**: Add `name="db_pre_upgrade_drift_scheduled"` to the pre-upgrade drift test and wire up `depends=["db_pre_upgrade_drift_scheduled"]` on the post-upgrade deletion test, OR add error-handling for the case where no metric exists.

#### 4. Namespace sharing between test classes needs a documentation comment

Both `TestPreUpgradeTrustyAIServiceDB` and `TestPostUpgradeTrustyAIServiceDB` use the same hardcoded namespace `"test-trustyaiservice-db-upgrade"`. This is correct by design (post-upgrade must reuse pre-upgrade resources), but there is no guard or comment preventing both classes from running in the same pytest session. Please add a comment at the `pytest.param` level explaining the intentional sharing and the constraint that these classes must never run in the same session.

#### 5. Post-upgrade fixture teardown order — add a scope comment

The `db_credentials_secret` post-upgrade path calls `secret.clean_up()` in the fixture teardown. The TrustyAI service requires this secret to remain alive for all three post-upgrade tests. Since all fixtures are `scope="class"`, pytest guarantees cleanup runs after all class tests finish — but this is an implicit contract. Add a comment on the fixture noting:
> `# cleanup is intentionally deferred: class-scope teardown runs after all tests in TestPostUpgradeTrustyAIServiceDB complete`

---

### 🟢 What's Done Well

1. **Security improvement is solid**: Dynamic password via `secrets.token_urlsafe(nbytes=24)` with `rh-pre-commit.check-secrets: ENABLED` confirmed in commit metadata.
2. **Docstrings on all 10 modified fixtures**: A significant readability/maintainability improvement.
3. **`teardown` propagation is backward-compatible**: Default `teardown=True` on all helper functions means zero impact on existing callers outside upgrade tests.
4. **`ensure_exists=True` pattern**: Correct — post-upgrade fixtures fail fast if the pre-upgrade run was incomplete, rather than silently re-creating resources.
5. **`@pytest.mark.dependency` chaining within each class**: The intra-class dependency chain is correct and will skip dependent tests on failure.
6. **DCO sign-off**: All three author commits carry valid `Signed-off-by: Shehan Saleem <ssaleem@redhat.com>`.
7. **CI green**: pre-commit, tox, and `3.4` image build all passed.

---

### Decision

Requesting changes on items **1** (dead constant — 1 line delete) and **2** (base64 decode in validator — 3 line fix). Both are low-effort and high-value. Items 3–5 are suggestions that would meaningfully improve robustness but are not hard blockers for a test-only backport.

Happy to re-review once those two are addressed. 🙌

PR #1550, opened 2026-05-14
added fixes for model registry test cases

diwakarsharma2:
# Pull Request

## Summary

1. multiple model catalog postgres pods were getting created with same suffix and at tear down, that sufffix was being used to delete the pods which was causing deletion of cluster managed model registry pods.
2.  Extra condition is added to teardown to make sure cluster managed model registry pods are not getting deleted.
3. Downgraded python vesion to support s390X plugins.

- [x] Locally
- [ ] Jenkins

## Additional Requirements

## Summary by CodeRabbit

* **Chores**
  * Broadened supported Python versions to >=3.12.
  * Modernized exception-handling syntax across tests and utilities.
  * Adjusted a type annotation in core logging utilities.

* **Bug Fixes**
  * Improved PostgreSQL handling in test infra so "postgres" and "postgresql" are treated equivalently.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,7 @@ dev = [
 ]
 
 [project]
-requires-python = "==3.14.*"
+requires-python = ">=3.12"
 name = "opendatahub-tests"
 version = "0.1.0"
 description = "Tests repository for Open Data Hub (ODH)"
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -936,7 +936,7 @@ def gpu_count_on_cluster(nodes: list[Any]) -> int:
             if key in allowed_exact or any(key.startswith(p) for p in allowed_prefixes):
                 try:
                     total_gpus += int(val)
-                except ValueError, TypeError:
+                except (ValueError, TypeError):
                     LOGGER.debug(f"Skipping non-integer allocatable for {key} on {node.name}: {val!r}")
                     continue
     return total_gpus
--- a/tests/model_registry/conftest.py
+++ b/tests/model_registry/conftest.py
@@ -1,11 +1,11 @@
 import os
+import uuid
 from collections.abc import Generator
 from contextlib import ExitStack
 from typing import Any
 
 import pytest
 import structlog
-import uuid
 from kubernetes.dynamic import DynamicClient
 from ocp_resources.config_map import ConfigMap
 from ocp_resources.data_science_cluster import DataScienceCluster
@@ -372,7 +372,9 @@ def model_registry_instance(
                 if registry.name == "model-registry0" or "model-registry0-postgres" or "db-model-registry0":
                     continue
                 else:
-                    wait_for_default_resource_cleanedup(admin_client=admin_client, namespace_name=model_registry_namespace)
+                    wait_for_default_resource_cleanedup(
+                        admin_client=admin_client, namespace_name=model_registry_namespace
+                    )
 
 
 @pytest.fixture(scope="class")
--- a/tests/model_registry/conftest.py
+++ b/tests/model_registry/conftest.py
@@ -5,6 +5,7 @@ from typing import Any
 
 import pytest
 import structlog
+import uuid
 from kubernetes.dynamic import DynamicClient
 from ocp_resources.config_map import ConfigMap
 from ocp_resources.data_science_cluster import DataScienceCluster
@@ -344,11 +345,14 @@ def model_registry_instance(
         yield [mr_instance]
         mr_instance.delete(wait=True)
     else:
-        db_name = param.get("db_name", "mysql")
+        db_name = param.get("db_name", "postgres")
+        unique_suffix = uuid.uuid4().hex[:6]
+        unique_base_name = f"{MR_INSTANCE_BASE_NAME}-{unique_suffix}"
+
         mr_objects = get_model_registry_objects(
             client=admin_client,
             namespace=model_registry_namespace,
-            base_name=MR_INSTANCE_BASE_NAME,
+            base_name=unique_base_name,
             num=param.get("num_resources", 1),
             teardown_resources=teardown_resources,
             params=param,
@@ -364,7 +368,11 @@ def model_registry_instance(
                 )
             yield mr_instances
         if db_name == "default":
-            wait_for_default_resource_cleanedup(admin_client=admin_client, namespace_name=model_registry_namespace)
+            for registry in ModelRegistry.get(namespace=model_registry_namespace, client=admin_client):
+                if registry.name == "model-registry0" or "model-registry0-postgres" or "db-model-registry0":
+                    continue
+                else:
+                    wait_for_default_resource_cleanedup(admin_client=admin_client, namespace_name=model_registry_namespace)
 
 
 @pytest.fixture(scope="class")
@@ -376,7 +384,7 @@ def model_registry_metadata_db_resources(
     model_registry_namespace: str,
 ) -> Generator[dict[Any, Any]]:
     num_resources = getattr(request, "param", {}).get("num_resources", 1)
-    db_backend = getattr(request, "param", {}).get("db_name", "mysql")
+    db_backend = getattr(request, "param", {}).get("db_name") or "postgres"
 
     if pytestconfig.option.post_upgrade:
         resources = {
--- a/tests/model_registry/model_catalog/search/utils.py
+++ b/tests/model_registry/model_catalog/search/utils.py
@@ -406,7 +406,7 @@ def _validate_single_criterion(
         else:
             LOGGER.warning(f"Unknown key_type: {key_type}")
             return False, f"{key_name}: unknown type {key_type}"
-    except ValueError, TypeError:
+    except (ValueError, TypeError):
         return False, f"{key_name}: conversion error"
 
     # Perform comparison based on type
--- a/tests/model_serving/maas_billing/utils.py
+++ b/tests/model_serving/maas_billing/utils.py
@@ -100,7 +100,7 @@ def mint_token(
     )
     try:
         body = resp.json()
-    except JSONDecodeError, ValueError:
+    except (JSONDecodeError, ValueError):
         body = {}
     return resp, body
 
@@ -401,7 +401,7 @@ def get_total_tokens(resp: Response, *, fail_if_missing: bool = False) -> int | 
     if header_val is not None:
         try:
             return int(header_val)
-        except TypeError, ValueError:
+        except (TypeError, ValueError):
             if fail_if_missing:
                 raise AssertionError(
                     f"Token usage header is not parseable as int; headers={dict(resp.headers)} body={resp.text[:500]}"
--- a/tests/model_serving/model_server/kserve/inference_graph/conftest.py
+++ b/tests/model_serving/model_server/kserve/inference_graph/conftest.py
@@ -128,13 +128,13 @@ def dog_breed_inference_graph(
 
     try:
         name = request.param["name"]
-    except AttributeError, KeyError:
+    except (AttributeError, KeyError):
         name = "dog-breed-pipeline"
 
     try:
         if not request.param["external-route"]:
             labels[networking_label] = "cluster-local"
-    except AttributeError, KeyError:
+    except (AttributeError, KeyError):
         pass
 
     with InferenceGraph(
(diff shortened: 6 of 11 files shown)


tests/model_registry/model_catalog/search/utils.py
```diff
@@ -406,7 +406,7 @@ def _validate_single_criterion(
         else:
             LOGGER.warning(f"Unknown key_type: {key_type}")
             return False, f"{key_name}: unknown type {key_type}"
-    except ValueError, TypeError:
+    except (ValueError, TypeError):
```
dbasunag, line 409: Please undo this change. With python 3.14, the original syntax was right.

tests/model_registry/conftest.py
```diff
@@ -344,11 +345,14 @@ def model_registry_instance(
         yield [mr_instance]
         mr_instance.delete(wait=True)
     else:
-        db_name = param.get("db_name", "mysql")
+        db_name = param.get("db_name", "postgres")
+        unique_suffix = uuid.uuid4().hex[:6]
+        unique_base_name = f"{MR_INSTANCE_BASE_NAME}-{unique_suffix}"
```
dbasunag, line 350: This might be breaking BYOIDC and upgrade flow. Why this change is needed can you explain please?

tests/model_registry/conftest.py
```diff
@@ -364,7 +368,13 @@ def model_registry_instance(
...
             yield mr_instances
         if db_name == "default":
-            wait_for_default_resource_cleanedup(admin_client=admin_client, namespace_name=model_registry_namespace)
+            for registry in ModelRegistry.get(namespace=model_registry_namespace, client=admin_client):
+                if registry.name == "model-registry0" or "model-registry0-postgres" or "db-model-registry0":
+                    continue
+                else:
+                    wait_for_default_resource_cleanedup(
+                        admin_client=admin_client, namespace_name=model_registry_namespace
+                    )
```
dbasunag, line 377: Can you please explain why this change is needed?

tests/model_serving/maas_billing/utils.py
```diff
@@ -100,7 +100,7 @@ def mint_token(
     )
     try:
         body = resp.json()
-    except JSONDecodeError, ValueError:
+    except (JSONDecodeError, ValueError):
```
dbasunag, line 103: Please undo this.

tests/model_serving/model_server/kserve/inference_graph/conftest.py
```diff
@@ -128,13 +128,13 @@ def dog_breed_inference_graph(
 
     try:
         name = request.param["name"]
-    except AttributeError, KeyError:
+    except (AttributeError, KeyError):
```
dbasunag, line 131: ditto

tests/model_serving/model_server/kserve/inference_graph/conftest.py
```diff
@@ -256,7 +256,7 @@ def bare_service_account(
     try:
         if request.param["name"]:
             name = request.param["name"]
-    except AttributeError, KeyError:
+    except (AttributeError, KeyError):
```
dbasunag, line 259: ditto.

tests/conftest.py
```diff
@@ -936,7 +936,7 @@ def gpu_count_on_cluster(nodes: list[Any]) -> int:
             if key in allowed_exact or any(key.startswith(p) for p in allowed_prefixes):
                 try:
                     total_gpus += int(val)
-                except ValueError, TypeError:
+                except (ValueError, TypeError):
```
dbasunag, line 939: ditto

utilities/plugins/openai_plugin.py
```diff
@@ -137,7 +137,7 @@ def get_request_http(host: str, endpoint: str) -> Any:
             if data:
                 data = OpenAIClient._remove_keys(data, keys_to_remove)
             return data
-        except requests.exceptions.RequestException, json.JSONDecodeError:
+        except (requests.exceptions.RequestException, json.JSONDecodeError):
```
dbasunag, line 140: ditto.

utilities/infra.py
```diff
@@ -984,7 +984,7 @@ def wait_for_serverless_pods_deletion(resource: Project | Namespace, admin_clien
                 LOGGER.info(f"Waiting for {KServeDeploymentType.SERVERLESS} pod {pod.name} to be deleted")
                 pod.wait_deleted(timeout=Timeout.TIMEOUT_1MIN)
 
-        except ResourceNotFoundError, NotFoundError:
+        except (ResourceNotFoundError, NotFoundError):
```
dbasunag, line 987: ditto

pyproject.toml
```diff
@@ -36,7 +36,7 @@ dev = [
 ]
 
 [project]
-requires-python = "==3.14.*"
+requires-python = ">=3.12"
```
dbasunag, line 39: We can't take this change.

dbasunag (changes requested):

mwaykole (changes requested):
## 🔍 PR Review — `added fixes for model registry test cases`

**Reviewer:** @mwaykole
**Risk Level:** 🟠 **MEDIUM** — Two logic bugs that will cause test behaviour opposite to the stated intent, plus a project-wide Python version constraint change that needs cross-team alignment.

---

## Summary

This PR bundles four distinct concerns:
1. Fix duplicate postgres pod creation in model-registry tests (**the stated goal**)
2. Fix accidental deletion of cluster-managed model-registry pods at teardown (**the stated goal**)
3. Downgrade Python requirement from `==3.14.*` → `>=3.12` for s390X plugin compatibility
4. Modernise comma-style `except` to tuple syntax across multiple files

Concerns 3 and 4 are straightforward and largely correct. Concerns 1 and 2 contain **critical logic bugs** that invert or nullify the intended fix.

---

## 🔴 Critical Issues

### 1. `model_registry/conftest.py` — Boolean short-circuit makes the guard always-True (lines 370-374)

**This is the most important bug in the PR.** The teardown guard intended to skip cluster-managed registries is broken by Python's operator precedence.

**Current code:**
```python
if registry.name == "model-registry0" or "model-registry0-postgres" or "db-model-registry0":
    continue
```

**What Python actually evaluates:**
```python
if (registry.name == "model-registry0") or ("model-registry0-postgres") or ("db-model-registry0"):
    continue
```

The non-empty string literals `"model-registry0-postgres"` and `"db-model-registry0"` are **always truthy**. This means **every** `ModelRegistry` object will hit `continue` — `wait_for_default_resource_cleanedup` is **never called** for any registry. The teardown is completely skipped.

**Fix required:**
```python
MANAGED_REGISTRY_NAMES = {"model-registry0", "model-registry0-postgres", "db-model-registry0"}
for registry in ModelRegistry.get(namespace=model_registry_namespace, client=admin_client):
    if registry.name in MANAGED_REGISTRY_NAMES:
        continue
    wait_for_default_resource_cleanedup(admin_client=admin_client, namespace_name=model_registry_namespace)
```

Also note the hardcoded names (`model-registry0`, `model-registry0-postgres`, `db-model-registry0`) are already defined as constants in `tests/model_registry/constants.py` (`MR_INSTANCE_NAME`, `MR_POSTGRES_DB_OBJECT`, `DB_RESOURCE_NAME`). Please use those constants instead of raw strings to avoid future drift.

---

### 2. `model_registry/utils.py` — Partial alias creates a split-brain backend configuration (line 149)

```python
# Only this function handles 'postgresql'
if db_backend in ["postgres", "postgresql"]:
```

But the entire sibling API in `utils.py` still only handles `"postgres"` as a strict equality check:
- `get_database_volumes()` → `if db_backend == "postgres"`
- `get_database_volume_mounts()` → `if db_backend == "postgres"`
- `get_database_image()` → `if db_backend == "postgres"`
- `get_database_env_vars()` (not shown in diff but same pattern)
- Port mapping in `PORT_MAP` constant only has `"postgres"` as the key

If any caller ever passes `"postgresql"`, it would receive Postgres health probes but MySQL volumes, MySQL images, MySQL env-vars, and port 3306. That's a guaranteed crashloop that is deceptively hard to diagnose.

**Either:**
- **(Preferred)** Normalize at the entry point: add `if db_backend == "postgresql": db_backend = "postgres"` at the top of `get_model_registry_deployment_template_dict()`, and revert this change.
- **(Acceptable)** Drop the alias here (revert to `== "postgres"`) since no caller in this PR actually passes `"postgresql"`.

---

## 🟡 Moderate Issues

### 3. `pyproject.toml` — Python version constraint loosened without CI matrix update

```toml
# Before
requires-python = "==3.14.*"
# After  
requires-python = ">=3.12"
```

This is a **project-wide change** that affects everyone, not just s390X. Concerns:
- The `.pre-commit-config.yaml` still hard-codes `default_language_version: python: python3.14`. If the project targets 3.12+, the pre-commit config should be updated to match, or you'll see failures on contributors running Python 3.12/3.13.
- This is why **pre-commit.ci is currently failing** on this PR.
- `>=3.12` is very broad — does the project intend to support 3.12, 3.13, and 3.14? This should be a deliberate, team-wide decision. Consider `>=3.12,<3.15` or a more specific range after alignment.
- The PR description mentions this is for s390X plugin support. Please document which specific package on s390X requires 3.12 compatibility so reviewers can assess the risk.

### 4. `tests/model_registry/conftest.py` — UUID suffix applied only to `model_registry_instance` but not `model_registry_metadata_db_resources`

The UUID-suffixed `unique_base_name` is generated inside `model_registry_instance` but the DB resources fixture (`model_registry_metadata_db_resources`) still uses `DB_BASE_RESOURCES_NAME` directly (constant, no suffix). If the root cause of pod collision was non-unique naming, this might be incomplete — the postgres DB deployment spawned by `model_registry_metadata_db_resources` may still collide. Please verify both fixtures are aligned.

### 5. `utilities/plugins/openai_plugin.py` — `get_request_http` silently swallows exceptions (line 140)

```python
except (requests.exceptions.RequestException, json.JSONDecodeError):
    LOGGER.exception("Request error")
    # No raise! Returns None implicitly
```

Callers expecting a `dict` will receive `None`, which defers the failure to a later `NoneType has no attribute` error — making root-cause analysis much harder. The sibling `streaming_request_http` correctly re-raises. This inconsistency was present before this PR but since you're touching this function, please fix it:
```python
except (requests.exceptions.RequestException, json.JSONDecodeError):
    LOGGER.exception("Request error")
    raise
```

---

## 🟢 Positives

- ✅ **Exception syntax fixes** across 8 files are correct and long overdue. The comma-separated form (`except ValueError, TypeError:`) was invalid Python 3 syntax and it is good to have this cleaned up. (Note: Python 3 would have been silently treating the second exception as the `as` binding variable name, which could cause subtle issues.)
- ✅ **UUID suffix for unique naming** is the right approach to prevent pod name collisions — it just needs the logic bug above fixed.
- ✅ **`utilities/logger.py`** — Quoting `"RedactedString"` as a forward reference is correct for older type-checker configs and is a safe change.
- ✅ **DB backend default change** from `"mysql"` → `"postgres"` is consistent with the broader model registry postgres migration direction.

---

## 🚫 CI Status

| Check | Status |
|-------|--------|
| DCO | ❌ FAIL |
| pre-commit.ci | ❌ FAIL |
| tox-tests | ✅ Pass |
| build | ✅ Pass |

**Action required:**
1. DCO sign-off is missing. Add `Signed-off-by: Your Name <your@email.com>` to all commits, or use `git commit --amend --signoff` + force push.
2. pre-commit failure is likely caused by the Python version mismatch between `pyproject.toml` (now `>=3.12`) and `.pre-commit-config.yaml` (`python3.14`). These need to be aligned.

---

## Required Changes Before Merge

| # | File | Issue | Severity |
|---|------|-------|----------|
| 1 | `tests/model_registry/conftest.py:370-374` | Boolean short-circuit makes teardown guard always-True | 🔴 Blocker |
| 2 | `tests/model_registry/utils.py:149` | Partial `postgresql` alias creates inconsistent backend config | 🔴 Blocker |
| 3 | `pyproject.toml` | `.pre-commit-config.yaml` must be updated to match new Python constraint | 🟠 Required |
| 4 | All commits | DCO sign-off missing | 🟠 Required |
| 5 | `utilities/plugins/openai_plugin.py:140` | `get_request_http` silently swallows exceptions | 🟡 Recommended |
| 6 | `tests/model_registry/conftest.py:370-374` | Use constants instead of hardcoded registry name strings | 🟡 Recommended |
| 7 | `tests/model_registry/conftest.py:349` | Log `unique_base_name` so CI failures are debuggable | 🟡 Recommended |

Happy to discuss any of these. The direction of this PR is right — the UUID naming fix and exception syntax modernization are good ideas — but the teardown guard logic bug would actively make the problem worse than the original code.

diwakarsharma2:
closing the PR as it is not required now.

PR #1552, opened 2026-05-14
feat: Added basic tests for eval-hub kueue integration

kpunwatk:
This PR adds baisc E2E tests for kueue integration with eval-hub.

# Pull Request

## Related Issues

- Fixes: 
- JIRA: 

- [ ] Locally
- [ ] Jenkins

## Additional Requirements

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive EvalHub↔Kueue integration tests validating workload admission, lifecycle, quota reservation, inadmissibility, and queue capacity scenarios; includes vLLM emulator, RBAC/user token setup, and namespace fixtures.
* **Utilities**
  * Added Kueue Workload support and updated Kueue resource API version handling.
* **Chores**
  * Centralized Kueue test constants (quota values and pinned vLLM emulator image).


tests/model_explainability/evalhub/kueue/conftest.py
```diff
@@ -0,0 +1,382 @@
...
+# ---------------------------------------------------------------------------
+
+
+def _is_kueue_operator_installed(admin_client: DynamicClient) -> bool:
+    """Check if the Kueue operator is installed and ready."""
+    try:
+        csvs = list(
+            ClusterServiceVersion.get(
+                client=admin_client,
+                namespace=py_config.get("applications_namespace", "openshift-operators"),
```
dbasunag: The default value would never apply, because https://github.com/opendatahub-io/opendatahub-tests/blob/main/tests/global_config.py#L12 exists. So get() is not needed here.

tests/model_explainability/evalhub/kueue/conftest.py
```diff
@@ -0,0 +1,382 @@
...
+    except ResourceNotFoundError:
+        return False
+
+
+@pytest.fixture(scope="session")
+def kueue_unmanaged_dsc(admin_client: DynamicClient, dsc_resource: DataScienceCluster) -> Generator[None, Any, Any]:
+    """Set DSC Kueue to Unmanaged and wait for CRDs to be available."""
+    try:
+        if not _is_kueue_operator_installed(admin_client):
+            pytest.skip("Kueue operator is not installed, skipping Kueue tests")
```
dbasunag: We should fail here. There is a mandate to remove all silent skips. If trusty needs kueue operator installed, then if it is not present we should fail.

tests/model_explainability/evalhub/kueue/conftest.py
```diff
@@ -0,0 +1,382 @@
...
+                wait_for_dsc_reconciliation(dsc=dsc_resource, baseline_time=pre_patch_time)
+            else:
+                LOGGER.info("Kueue already Unmanaged, no patch needed")
+
+            # Always wait for Kueue CRDs and controller pods (regardless of patch)
+            wait_for_kueue_crds_available(client=admin_client)
+            yield
+
+    except (AttributeError, KeyError) as e:
+        pytest.skip(f"Kueue component not found in DSC: {e}")
```
dbasunag: Same comment. No skips please. We should fail if kueue component is missing.

tests/model_explainability/evalhub/kueue/conftest.py
```diff
@@ -0,0 +1,382 @@
...
+        client=admin_client,
+    ) as resource_flavor:
+        yield resource_flavor
+
+
+@pytest.fixture(scope="class")
+def evalhub_kueue_multi_job_cluster_queue(
+    admin_client: DynamicClient,
+    evalhub_kueue_multi_job_resource_flavor: ResourceFlavor,
+    kueue_unmanaged_dsc,
```
dbasunag: missing types.

tests/model_explainability/evalhub/kueue/conftest.py
```diff
@@ -0,0 +1,382 @@
...
+            sleep=5,
+            func=tenant_rbac_ready,
+            admin_client=admin_client,
+            namespace=evalhub_kueue_namespace.name,
+        ):
+            if ready:
+                LOGGER.info(f"Operator RBAC provisioned in {evalhub_kueue_namespace.name}")
+                return
+    except TimeoutExpiredError as err:
+        raise RuntimeError(f"Operator RBAC not provisioned in {evalhub_kueue_namespace.name}") from err
```
dbasunag: Why not raise the same exception?

tests/model_explainability/evalhub/kueue/test_evalhub_kueue_basic.py
```diff
@@ -0,0 +1,252 @@
...
+    validate_evalhub_job_completed,
+    wait_for_evalhub_job,
+    wait_for_evalhub_job_workload_admitted,
+    wait_for_evalhub_job_workload_inadmissible,
+)
+from utilities.kueue_utils import LocalQueue, check_gated_pods_and_running_pods
+
+
+@pytest.mark.tier1
+@pytest.mark.kueue
```
dbasunag: We don't need this marker, unless they would be executed separately from trusty lanes.

tests/model_explainability/evalhub/kueue/test_evalhub_kueue_basic.py
```diff
@@ -0,0 +1,252 @@
...
+@pytest.mark.model_explainability
```
dbasunag: `@pytest.mark.model_explainability` should be removed. The tests are already under tests/model_explainability.

tests/model_explainability/evalhub/kueue/conftest.py
kpunwatk: Done, thanks for suggestion

nbs-rh (approved):
LGTM

mwaykole (commented):
## PR Review — EvalHub × Kueue E2E Tests

**Reviewer:** mwaykole (Dev Engineer)
**Risk Level:** 🟡 MEDIUM — new test infrastructure, no production code changed, but several design issues that could cause flaky tests or mask real failures in CI.

---

### ✅ What's Good

- **Security posture is correct**: The vLLM emulator container spec has `allowPrivilegeEscalation: false`, `runAsNonRoot: true`, `capabilities.drop: ["ALL"]`, and `seccompProfile: RuntimeDefault`. Well done — this matches project security standards.
- **DSC patch with baseline timestamp**: Reading `lastTransitionTime` *before* applying the patch to `kueue_unmanaged_dsc` is a subtle but important correctness detail. Good defensive coding.
- **ExitStack for conditional cleanup**: Using `ExitStack` to only register the `ResourceEditor` context when a patch is actually applied is the right approach.
- **Workload class placement**: Adding `Workload` to `utilities/kueue_utils.py` alongside `ResourceFlavor`, `LocalQueue`, and `ClusterQueue` is consistent with the module's intent.
- **Test scenario coverage**: The four test cases together (created → lifecycle → status-reflects-state → capacity-exhaustion) provide a solid progression of assertions.

---

### 🔴 Must-Fix Issues

#### 1. Mutable image tag — pinned digest required
**File:** `tests/model_explainability/evalhub/kueue/constants.py:9`

```python
# Current — WRONG
VLLM_EMULATOR_IMAGE = "quay.io/opendatahub/vllm-openai-ubi9:fast"
```

The tag `fast` is a mutable floating tag. If the image is updated, tests will silently run against a different image version, leading to non-reproducible CI results. The rest of the codebase pins by digest (e.g., `multitenancy/conftest.py` uses `quay.io/trustyai_testing/vllm_emulator@sha256:c4bdd5bb93...`). Please follow the same pattern:

```python
# Correct — pin by digest
VLLM_EMULATOR_IMAGE = "quay.io/opendatahub/vllm-openai-ubi9@sha256:<digest>"
```

Also note: the multitenancy tests already use `quay.io/trustyai_testing/vllm_emulator` from a testing-specific registry. Using a different image from a different registry here is an inconsistency worth explaining in the PR description. Was this intentional? If so, please document why.

---

#### 2. Broad label selector in capacity-exhaustion test — will produce false positives
**File:** `tests/model_explainability/evalhub/kueue/test_evalhub_kueue_basic.py:723-731`

```python
# Current — matches ALL evalhub jobs in the namespace, not just the two submitted
all_labels = "app=evalhub,component=evaluation-job"
running_pods, gated_pods = check_gated_pods_and_running_pods(
    labels=[all_labels],
    namespace=evalhub_kueue_namespace.name,
    admin_client=admin_client,
)
assert running_pods == 1, ...
assert gated_pods == 1, ...
```

This selector matches *every* evaluation-job pod in `test-evalhub-kueue`. If any pod from `test_evalhub_job_workload_created` or `test_evalhub_job_lifecycle_with_kueue` is still Running at this point (test class is shared), the count will be wrong.

**Fix:** Build the selector using the two submitted job IDs:

```python
selector1 = evalhub_runtime_label_selector(evalhub_job_id=job1_id)
selector2 = evalhub_runtime_label_selector(evalhub_job_id=job2_id)
running_pods, gated_pods = check_gated_pods_and_running_pods(
    labels=[selector1, selector2],
    namespace=evalhub_kueue_namespace.name,
    admin_client=admin_client,
)
```

---

#### 3. Missing `@pytest.mark.kueue` marker — test cannot be selected by CI label
**File:** `tests/model_explainability/evalhub/kueue/test_evalhub_kueue_basic.py:22`

`pytest.ini` explicitly registers a `kueue` marker. The test class only has `@pytest.mark.tier1`. Jenkins jobs likely select Kueue tests with `-m kueue`. Without the marker, these tests will never run in the dedicated Kueue CI job.

```python
@pytest.mark.tier1
@pytest.mark.kueue   # ← add this
class TestEvalHubKueueBasic:
```

---

#### 4. `evalhub_kueue_namespace` does not depend on `kueue_unmanaged_dsc` — ordering hazard
**File:** `tests/model_explainability/evalhub/kueue/conftest.py`

The namespace fixture is created independently of `kueue_unmanaged_dsc`. If the namespace is labeled as an EvalHub tenant namespace *before* Kueue is set to Unmanaged, the operator may auto-provision resources into a partially-initialized state. All class-scoped queue fixtures correctly depend on `kueue_unmanaged_dsc`, but `evalhub_kueue_namespace` does not.

```python
def evalhub_kueue_namespace(
    admin_client: DynamicClient,
    kueue_unmanaged_dsc: None,  # ← add this dependency
) -> Generator[Namespace, Any, Any]:
```

---

### 🟡 Should-Fix Issues

#### 5. Incomplete PR description — typo in title and all template fields empty
The PR body says *"This PR adds baisc E2E tests..."* (typo: "baisc") and leaves all template fields empty — no Jira link, no related issue, no checkbox ticked. CodeRabbit also flagged this (Description check: ❌). Please fill it in. This is a blocking concern for reviewers who need to trace this back to a Jira ticket and understand how it was tested.

---

#### 6. Bare `for` loop over `TimeoutSampler` — no timeout error handler
**File:** `tests/model_explainability/evalhub/kueue/test_evalhub_kueue_basic.py` (test_evalhub_status_reflects_kueue_state)

```python
for sample in TimeoutSampler(wait_timeout=180, sleep=10, func=_get_job_status, ...):
    state = sample.get("status", {}).get("state", "")
    if state == "running":
        break
# ← no except TimeoutExpiredError
```

`TimeoutSampler` raises `TimeoutExpiredError` when the timeout expires. If the job never reaches `running` state, the exception propagates as an unhandled error rather than a clean test failure with a meaningful message. Wrap it:

```python
try:
    for sample in TimeoutSampler(wait_timeout=180, sleep=10, func=_get_job_status, ...):
        if sample.get("status", {}).get("state") == "running":
            break
except TimeoutExpiredError:
    pytest.fail("EvalHub job did not reach 'running' state within 180s")
```

---

#### 7. Hardcoded namespace name `test-evalhub-kueue` — parallel run collision risk
**File:** `tests/model_explainability/evalhub/kueue/conftest.py:175`

All resource names are hardcoded (`evalhub-multi-flavor`, `evalhub-single-cluster-queue`, etc.). If two runs of this test class run concurrently (different Jenkins pipelines sharing a cluster), they will collide. Consider a `uuid` suffix, similar to how other tests in the suite generate namespace names dynamically.

---

#### 8. `Workload` class is missing explicit `api_version` — may default to wrong version
**File:** `utilities/kueue_utils.py:98-101`

```python
class Workload(NamespacedResource):
    """Kueue Workload resource (kueue.x-k8s.io/v1beta1)."""
    api_group: str = "kueue.x-k8s.io"
```

The docstring says `v1beta1` but no `api_version` is declared. The `ocp-resources` framework will infer the version from the cluster's API discovery, which may return a different preferred version if Kueue upgrades. Explicitly pin it:

```python
class Workload(NamespacedResource):
    """Kueue Workload resource (kueue.x-k8s.io/v1beta1)."""
    api_group: str = "kueue.x-k8s.io"
    api_version: str = "v1beta1"
```

---

#### 9. Private function `_get_job_status` imported directly in test file
**File:** `tests/model_explainability/evalhub/kueue/test_evalhub_kueue_basic.py:10`

```python
from tests.model_explainability.evalhub.utils import (
    _get_job_status,
    ...
)
```

Functions prefixed with `_` are by convention private to their module. Either rename it to `get_job_status` to signal it is part of the public test utility API, or replace the direct call in the test with the existing `wait_for_evalhub_job` helper and remove this import.

---

#### 10. `_get_evalhub_job_workload` — label propagation assumption undocumented
**File:** `tests/model_explainability/evalhub/utils.py` (new Kueue section)

The lookup uses `label_selector` and the comment says Kueue uses an *owner reference*, but the actual mechanism relied on is Kueue's *label propagation* from the parent batch Job onto the Workload. This works today but could silently break if Kueue changes propagation behavior. The comment should state this explicitly so future maintainers understand the contract.

---

### 📋 Checklist Before Merge

- [ ] **CRITICAL** — Pin `VLLM_EMULATOR_IMAGE` to an immutable digest
- [ ] **CRITICAL** — Add `@pytest.mark.kueue` to `TestEvalHubKueueBasic`
- [ ] **CRITICAL** — Fix broad label selector in `test_queue_capacity_exhaustion` to scope to submitted job IDs
- [ ] **SHOULD** — Add `kueue_unmanaged_dsc` as dependency of `evalhub_kueue_namespace`
- [ ] **SHOULD** — Wrap `TimeoutSampler` loop in `try/except TimeoutExpiredError` with `pytest.fail()`
- [ ] **SHOULD** — Add `api_version = "v1beta1"` to `Workload` class
- [ ] **NICE** — Rename `_get_job_status` → `get_job_status` or stop importing private functions in tests
- [ ] **REQUIRED** — Fill in PR description: Jira ticket, test evidence, testing checkboxes
- [ ] **REQUIRED** — Fix typo "baisc" → "basic" in PR title/description
- [ ] **CLARIFY** — Explain why a different vLLM image + registry is used vs. the one in multitenancy tests

---

Overall this is a good first pass at Kueue integration coverage for EvalHub. The test scenario selection is solid, fixture teardown logic is mostly correct, and the security context on the emulator deployment is properly hardened. The 🔴 must-fix items are all about CI reliability (mutable image tag, missing marker, broad selector) rather than fundamental design problems. Addressing those three will get this to a mergeable state.

tests/model_explainability/evalhub/kueue/constants.py
```diff
@@ -0,0 +1,9 @@
+# Kueue Configuration
+SINGLE_JOB_CPU_QUOTA = "2"
+SINGLE_JOB_MEMORY_QUOTA = "4Gi"
+MULTI_JOB_CPU_QUOTA = "8"
+MULTI_JOB_MEMORY_QUOTA = "16Gi"
+
+# vLLM emulator configuration
+VLLM_EMULATOR = "vllm-emulator"
+VLLM_EMULATOR_IMAGE = "quay.io/opendatahub/vllm-openai-ubi9:fast"
```
mwaykole: 🔴 **MUST FIX — Mutable floating tag**

The `fast` tag is mutable and will silently change between CI runs, making tests non-reproducible. The rest of the codebase (see `multitenancy/conftest.py`) pins images by SHA256 digest.

```python
# Replace with an immutable digest pin:
VLLM_EMULATOR_IMAGE = "quay.io/opendatahub/vllm-openai-ubi9@sha256:<digest>"
```

Also, the existing vLLM emulator tests use `quay.io/trustyai_testing/vllm_emulator` — please clarify in the PR description why a different image/registry is used here.

tests/model_explainability/evalhub/kueue/test_evalhub_kueue_basic.py
```diff
@@ -0,0 +1,250 @@
...
+    evalhub_runtime_label_selector,
+    submit_evalhub_job,
+    validate_evalhub_job_completed,
+    wait_for_evalhub_job,
+    wait_for_evalhub_job_workload_admitted,
+    wait_for_evalhub_job_workload_inadmissible,
+)
+from utilities.kueue_utils import LocalQueue, check_gated_pods_and_running_pods
+
+
```
mwaykole, line 22: 🔴 **MUST FIX — Missing `@pytest.mark.kueue` marker**

`pytest.ini` registers a `kueue` marker specifically for selecting Kueue tests. Without it, Jenkins CI jobs that run with `-m kueue` will never execute this test class.

```python
@pytest.mark.tier1
@pytest.mark.kueue  # ← add this
class TestEvalHubKueueBasic:
```

tests/model_explainability/evalhub/kueue/test_evalhub_kueue_basic.py
```diff
@@ -0,0 +1,250 @@
...
+        for sample in TimeoutSampler(
+            wait_timeout=180,
+            sleep=10,
+            func=_get_job_status,
+            host=evalhub_mt_route.host,
+            token=evalhub_kueue_user_token,
+            ca_bundle_file=evalhub_mt_ca_bundle_file,
+            tenant=evalhub_kueue_namespace.name,
+            job_id=job_id,
+        ):
```
mwaykole: 🔴 **MUST FIX — Broad label selector produces incorrect pod counts**

`"app=evalhub,component=evaluation-job"` matches **all** evalhub evaluation pods in the namespace — not just the two jobs submitted in this test. If pods from earlier tests in this class are still Running, `assert running_pods == 1` will fail non-deterministically.

Fix by scoping to the two submitted job IDs:

```python
selector1 = evalhub_runtime_label_selector(evalhub_job_id=job1_id)
selector2 = evalhub_runtime_label_selector(evalhub_job_id=job2_id)
running_pods, gated_pods = check_gated_pods_and_running_pods(
    labels=[selector1, selector2],
    namespace=evalhub_kueue_namespace.name,
    admin_client=admin_client,
)
```

tests/model_explainability/evalhub/kueue/conftest.py
```diff
@@ -0,0 +1,447 @@
...
+                            DscComponents.KUEUE: {"managementState": DscComponents.ManagementState.UNMANAGED}
+                        }
+                    }
+                }
+                stack.enter_context(cm=ResourceEditor(patches={dsc_resource: dsc_dict}))
+
+                # Wait for DSC to reconcile the patch
+                wait_for_dsc_reconciliation(dsc=dsc_resource, baseline_time=pre_patch_time)
+            else:
+                LOGGER.info("Kueue already Unmanaged, no patch needed")
```
mwaykole: 🟡 **SHOULD FIX — Namespace created before Kueue is ready**

This fixture has no dependency on `kueue_unmanaged_dsc`. If pytest creates the namespace before Kueue is in Unmanaged state, the operator may reconcile tenant RBAC against a partially-initialized Kueue. Add the dependency:

```python
def evalhub_kueue_namespace(
    admin_client: DynamicClient,
    kueue_unmanaged_dsc: None,  # ← add
) -> Generator[Namespace, Any, Any]:
```

utilities/kueue_utils.py
```diff
@@ -95,6 +95,12 @@ def to_dict(self) -> None:
                 _spec["resourceGroups"] = self.resource_groups
 
 
+class Workload(NamespacedResource):
```
mwaykole, line 107: 🟡 **SHOULD FIX — Missing explicit `api_version`**

The docstring says `v1beta1` but `api_version` is not declared. The `ocp-resources` library will infer the preferred version from API discovery, which may return `v1beta2` after a Kueue operator upgrade.

```python
class Workload(NamespacedResource):
    """Kueue Workload resource (kueue.x-k8s.io/v1beta1)."""
    api_group: str = "kueue.x-k8s.io"
    api_version: str = "v1beta1"  # ← add this
```

tests/model_explainability/evalhub/kueue/test_evalhub_kueue_basic.py
```diff
@@ -0,0 +1,250 @@
...
+        evalhub_kueue_single_job_local_queue: LocalQueue,
+        evalhub_kueue_user_token: str,
+        evalhub_kueue_vllm_service: Service,
+        evalhub_mt_route: Route,
+        evalhub_mt_ca_bundle_file: str,
+    ) -> None:
+        """Validate failure behavior: job rejection when queue capacity is exhausted.
+
+        Submit 2 jobs with quota for 1, verify second is gated.
+        """
```
mwaykole, line 200: 🟡 **SHOULD FIX — Unhandled `TimeoutExpiredError`**

If the job never reaches `running` state, `TimeoutSampler` raises `TimeoutExpiredError` which will appear as an exception rather than a clean test failure. Wrap it:

```python
from timeout_sampler import TimeoutExpiredError

try:
    for sample in TimeoutSampler(wait_timeout=180, sleep=10, func=_get_job_status, ...):
        if sample.get("status", {}).get("state") == "running":
            break
except TimeoutExpiredError:
    pytest.fail("EvalHub job did not reach 'running' state within 180s")
```

mwaykole (dismissed):
Requesting changes on the three critical blocking issues: (1) mutable image tag, (2) missing `@pytest.mark.kueue` marker, and (3) broad label selector in capacity exhaustion test. Details in the general review comment above.

tests/model_explainability/evalhub/kueue/test_evalhub_kueue_basic.py
kpunwatk: @dbasunag suggested to remove this as "We don't need this marker, unless they would be executed separately from trusty lanes."

tests/model_explainability/evalhub/kueue/test_evalhub_kueue_basic.py
kpunwatk: This fix were pushed in commit 29f9957 and are currently on the kueue_integration branch.

tests/model_explainability/evalhub/kueue/test_evalhub_kueue_basic.py
kpunwatk: This fix were pushed in commit 29f9957 and are currently on the kueue_integration branch.

dbasunag (approved):

sheltoncyril (approved):
/lgtm

PR #1554, opened 2026-05-14
test: add tests for validatedTasks and ServingConfig

dbasunag:
# Pull Request

## Related Issues

- Fixes: 
- JIRA: 

- [ ] Locally
- [ ] Jenkins

## Additional Requirements

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end tests for models with and without serving config and validated tasks, verifying returned metadata, searchable validated-task terms, and behavior when fields are omitted.
  * Applied session-scoped fixtures to multiple test modules to ensure a consistent test environment across suites.
* **Bug Fixes**
  * Excluded serving_config from the catalog API's filter options.


mwaykole (commented):
## 🔍 PR Review — `test: add tests for validatedTasks and ServingConfig`

**Reviewer:** @mwaykole (Autonomous Dev Engineer)
**Risk Level:** 🟡 LOW-MEDIUM — pure test addition, no production code changed
**Verdict:** Changes are solid but have a few correctness, robustness, and structural issues that should be addressed before merge.

---

### ✅ What's Good

- **Clear intent**: The test class name, docstrings, and parametrize ID all clearly communicate what's being tested.
- **Good coverage breadth**: Four distinct test scenarios — model with fields, model without fields, filter-options exclusion, and search-by-validated-task — is exactly the right decomposition.
- **Correct fixture usage**: `updated_catalog_config_map` + `model_registry_namespace` + `@pytest.mark.tier1` follow the established patterns of `test_custom_model_catalog.py` well.
- **YAML builder helper**: Extracting `_build_custom_yaml_with_serving_config()` keeps the test body clean — matches how `get_sample_yaml_str()` is used elsewhere.
- `db_constants.py` change is correct and well-placed: `serving_config` belongs in `API_EXCLUDED_FILTER_FIELDS` alongside `source_id`, `logo`, and `license_link`.

---

### 🚨 Issues — Please Address

#### 1. **[BUG] `test_validated_tasks_searchable_via_q_parameter` — search may silently pass on wrong data**

The test assumes `q=tool-calling` returns the model because `validatedTasks` contains `'tool-calling'`. **But look at `SEARCH_MODELS_DB_QUERY` in `db_constants.py`** — the backend search query matches against:
```sql
name, description, provider, libraryName, tasks
```
Notice: **`validatedTasks` is NOT in the search index** — the backend searches `tasks`, not `validatedTasks`. Your custom model YAML also sets `tasks: [text-generation, tool-calling]`, which means the search _will_ match — but it's matching on `tasks`, not `validatedTasks`.

This is a **semantic gap**: the test claims it validates that `validatedTasks` is searchable, but what's actually being tested is that `tasks` is searchable (the model happens to have matching values in both fields).

**Recommendation**: Either:
- (a) Update the docstring to correctly state what is being searched (the `tasks` field, not `validatedTasks`), OR
- (b) Create a model where `tasks` and `validatedTasks` differ so the test truly isolates searchability of `validatedTasks`

#### 2. **[BUG] Missing `wait_for_model_catalog_api` / readiness check between ConfigMap update and API calls**

Every other test class using `updated_catalog_config_map` in this package relies on `catalog_config/conftest.py`'s `updated_catalog_config_map` fixture which already calls `wait_for_model_catalog_pod_ready_after_deletion` + `wait_for_model_catalog_api`. So the fixture chain handles it.

However, this test class mixes fixture setup at class level (`scope='class'`) with individual test-level API calls that depend on the catalog being ready. The problem is that `test_serving_config_excluded_from_filter_options` calls `filter_options` — which queries the **global catalog state**, not a source-specific endpoint. If the pod hasn't synced the new custom data to the DB by the time this test runs, the test may pass trivially (the field isn't there simply because the catalog hasn't been loaded yet, not because it's intentionally excluded). 

**Recommendation**: Add an explicit assertion that the custom source `CUSTOM_CATALOG_ID1` is visible via `/sources` before running `filter_options`, to confirm the catalog loaded successfully.

#### 3. **[CORRECTNESS] `EXPECTED_REQUIRED_ARGS` format mismatch**

In the YAML:
```yaml
requiredArgs:
  - "--config_format granite"
```

But the constant is defined as:
```python
EXPECTED_REQUIRED_ARGS = ["--config_format granite"]
```

This looks fine at first glance, but the assertion is:
```python
assert tool_calling.get("requiredArgs") == EXPECTED_REQUIRED_ARGS
```

The risk is that the catalog backend may parse and deserialize `requiredArgs` differently depending on how the YAML `servingConfig` field is stored in the DB. If `requiredArgs` comes back as a JSON string or a comma-joined string rather than a Python list, this assertion will fail with a confusing error. 

**Recommendation**: Add a type assertion before the equality check:
```python
assert isinstance(tool_calling.get('requiredArgs'), list),     f"requiredArgs should be a list, got: {type(tool_calling.get('requiredArgs'))}"
assert tool_calling.get("requiredArgs") == EXPECTED_REQUIRED_ARGS
```

#### 4. **[STRUCTURE] Class scope `updated_catalog_config_map` with function-scope fixture mismatch**

The `updated_catalog_config_map` fixture in `model_catalog/conftest.py` is `scope='class'` — meaning the ConfigMap patch is applied once for all four tests in the class. This is correct and intended.

However, **none of the four test methods** use `updated_catalog_config_map` as an argument — they only list it in `@pytest.mark.usefixtures`. This means if `updated_catalog_config_map` fails (e.g., the catalog pod doesn't come back ready), the tests still run against whatever the cluster's current state is, with no assertion failure at setup time.

**Recommendation**: Consider having at least the first test method explicitly receive `updated_catalog_config_map: ConfigMap` as an argument and assert it `ensure_exists`, consistent with how `test_custom_model_catalog.py` does it.

#### 5. **[MINOR] Missing Jira/GitHub issue link in PR description and test docstring**

The test class docstring references `RHOAIENG-60668` but:
- The PR description has all template placeholders blank (CodeRabbit already flagged this ⚠️)
- No related Jira ticket is linked

**Recommendation**: Fill in the PR description — at minimum: summary of what this tests, JIRA ticket, and check the "Locally" testing box.

---

### 💡 Suggestions (Non-blocking)

#### S1. Namespace the custom model names more defensively

```python
MODEL_WITH_SERVING_CONFIG = "test/model-with-serving-config"
MODEL_WITHOUT_SERVING_CONFIG = "test/model-without-serving-config"
```

These names look innocuous but could collide if another test (running concurrently or in the same namespace) also injects a model named `test/model-with-serving-config` into `CUSTOM_CATALOG_ID1`. Consider adding a unique suffix (e.g., based on module name or a UUID4 prefix) as `_build_custom_yaml_with_serving_config` already uses `current_time` for epochs — the names themselves would be safer with a suffix like `-sc` or a test-file-specific prefix.

#### S2. Use `sorted()` consistently for list comparisons

Test 1 uses:
```python
assert sorted(validated_tasks) == sorted(EXPECTED_VALIDATED_TASKS)
```
Good call — order-independent comparison. Consider applying the same to `requiredArgs` if you don't want to assert order:
```python
assert sorted(tool_calling.get("requiredArgs", [])) == sorted(EXPECTED_REQUIRED_ARGS)
```
(Only if order isn't semantically significant for `requiredArgs`.)

#### S3. `_build_custom_yaml_with_serving_config` could be a module-level constant

Since the YAML content doesn't change between test runs (the `current_time` trick is already applied for epoch fields, but the model names and serving config are static), this could be computed once at import time rather than being called per-parametrize invocation. The current approach works fine but may surprise readers who expect module-level helpers to be constant.

#### S4. Add `@pytest.mark.ModelRegistry` label to match PR labels

The PR is labeled `ModelRegistry` in GitHub — the corresponding pytest marker should be applied to the class so Jenkins/CI can correctly bucket this test, consistent with the marker taxonomy in `pytest.ini`.

---

### 📋 Summary Table

| # | Issue | Severity | Required? |
|---|-------|----------|-----------|
| 1 | `q` search test validates wrong field (tasks vs validatedTasks) | 🔴 BUG | Yes |
| 2 | filter_options test may pass trivially if catalog not loaded | 🟡 CORRECTNESS | Yes |
| 3 | `requiredArgs` type not validated before equality check | 🟡 CORRECTNESS | Yes |
| 4 | `updated_catalog_config_map` not in method signatures | 🟡 STRUCTURE | Recommended |
| 5 | PR description template blank, JIRA not linked | 🟠 PROCESS | Yes |
| S1 | Model name collision risk | 🟢 SUGGESTION | No |
| S2 | Inconsistent sorted() for requiredArgs | 🟢 SUGGESTION | No |
| S3 | YAML builder as constant vs function | 🟢 SUGGESTION | No |
| S4 | Missing `@pytest.mark.ModelRegistry` | 🟢 SUGGESTION | No |

---

Good work on the overall structure and coverage intent, @dbasunag. The critical fix is item #1 — the search test is testing `tasks` searchability, not `validatedTasks` — the docstring and/or test design needs to be corrected so the intent is clear and not accidentally passing for the wrong reason. Happy to discuss the right approach for #1 if you need a second opinion on the design.

tests/model_registry/model_catalog/catalog_config/test_serving_config_validated_tasks.py
```diff
@@ -0,0 +1,180 @@
...
+
+    def test_validated_tasks_searchable_via_q_parameter(
+        self: Self,
+        model_catalog_rest_url: list[str],
+        model_registry_rest_headers: dict[str, str],
+    ):
+        """Given a custom catalog model with validatedTasks containing 'tool-calling'
+        When searching with q=tool-calling scoped to the custom source
+        Then the model with that validated task is returned in results
+        """
```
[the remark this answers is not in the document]
dbasunag: For tests we don't do explicit return typing.

tests/model_registry/model_catalog/catalog_config/test_serving_config_validated_tasks.py
```diff
@@ -0,0 +1,196 @@
...
+@pytest.mark.tier1
+class TestServingConfigAndValidatedTasks:
+    """Tests for validatedTasks and servingConfig fields in custom catalog models (RHOAIENG-60668)."""
+
+    def test_model_with_serving_config_and_validated_tasks(
+        self: Self,
+        updated_catalog_config_map: ConfigMap,
+        model_catalog_rest_url: list[str],
+        model_registry_rest_headers: dict[str, str],
+    ):
```
SB159, line 99: Missing -> None return type annotations,

tests/model_registry/model_catalog/catalog_config/test_serving_config_validated_tasks.py
```diff
@@ -0,0 +1,196 @@
...
+        required_args = tool_calling.get("requiredArgs")
+        assert isinstance(required_args, list), f"requiredArgs should be a list, got: {type(required_args)}"
+        assert required_args == EXPECTED_REQUIRED_ARGS
+
+    def test_model_without_serving_config_and_validated_tasks(
+        self: Self,
+        updated_catalog_config_map: ConfigMap,
+        model_catalog_rest_url: list[str],
+        model_registry_rest_headers: dict[str, str],
+    ):
```
SB159, line 132: same as above.

tests/model_registry/model_catalog/catalog_config/test_serving_config_validated_tasks.py
```diff
@@ -0,0 +1,196 @@
...
+        assert not validated_tasks, (
+            f"validatedTasks should be empty for model '{MODEL_WITHOUT_SERVING_CONFIG}', got {validated_tasks}"
+        )
+
+    def test_serving_config_excluded_from_filter_options(
+        self: Self,
+        updated_catalog_config_map: ConfigMap,
+        model_catalog_rest_url: list[str],
+        model_registry_rest_headers: dict[str, str],
+    ):
```
SB159, line 156: same as above

tests/model_registry/model_catalog/catalog_config/test_serving_config_validated_tasks.py
```diff
@@ -0,0 +1,196 @@
...
+        )
+        filter_keys = set(response.get("filters", {}).keys())
+        assert "serving_config" not in filter_keys, "serving_config should be excluded from filter_options"
+
+    def test_validated_tasks_searchable_via_q_parameter(
+        self: Self,
+        updated_catalog_config_map: ConfigMap,
+        model_catalog_rest_url: list[str],
+        model_registry_rest_headers: dict[str, str],
+    ):
```
SB159, line 182: same as above

tests/model_registry/model_catalog/catalog_config/test_serving_config_validated_tasks.py
dbasunag: We don't require type annotations for tests. This was always the case.

tests/model_registry/model_catalog/catalog_config/test_serving_config_validated_tasks.py
dbasunag: Same response.

tests/model_registry/model_catalog/catalog_config/test_serving_config_validated_tasks.py
dbasunag: ditto

tests/model_registry/model_catalog/catalog_config/test_serving_config_validated_tasks.py
dbasunag: ditto.

SB159 (dismissed):
/lgtm

SB159 (dismissed):
/lgtm

tests/model_registry/model_catalog/catalog_config/test_serving_config_validated_tasks.py
```diff
@@ -0,0 +1,193 @@
...
+            },
+            id="test_serving_config_validated_tasks",
+        ),
+    ],
+    indirect=["updated_catalog_config_map"],
+)
+@pytest.mark.usefixtures("model_registry_namespace")
+@pytest.mark.jira("RHOAIENG-60668")
+@pytest.mark.tier1
+class TestServingConfigAndValidatedTasks:
```
fege, line 91: don't wee need `updated_dsc_component_state_scope_session` ?

dbasunag: ~you are right. tier1 requires custom namespace.~
Looks like we do that only for model registry tests. Model catalog or mcp catalog (tier1) does not use this fixture.

lugi0 (dismissed):

fege (approved):
/lgtm

lugi0 (approved):

PR #1624, opened 2026-05-29
RHAIENG-4973 - Fix Llama Stack vector store upgrade tests

jiripetrlik:
- Fix for this issue: When running the pre-upgrade tests, the llama_stack_client [_create_llama_stack_client](https://github.com/opendatahub-io/opendatahub-tests/blob/69f491453ff7fa60228af750c263cbd15583fa0a/tests/llama_stack/conftest.py#L665) function is currently deleting all vector store files in the yield section.  This is a bug, as then in post-upgrade the vector store exist but not the files
- Also, modify the vector store providers configuration in all the pre-upgrade and post-upgrade tests not to use inline providers, as they are being removed in 3.5 ea1
- Changes link to testing document


Artemon-line (approved):
LGTM!

dbasunag (commented):
@jiripetrlik can you please fix the dco by signing your commit? We can take the PR after that.

jgarciao (approved):

dbasunag (approved):

PR #1642, opened 2026-06-02
test(workbenches): add stopped notebook upgrade test

jstourac:

## Summary

- Adds pre- and post-upgrade tests verifying that a notebook stopped via `kubeflow-resource-stopped` annotation remains stopped after a platform upgrade (no unexpected pod creation, StatefulSet stays at replicas=0, annotation value preserved).
- Introduces a second notebook (`upgrade-wb-stopped`) with its own PVC and lifecycle in the shared upgrade namespace.

## Details

The notebook controller uses the `kubeflow-resource-stopped` annotation to scale the StatefulSet to 0 replicas. If the upgraded controller incorrectly handles this annotation (e.g., by removing it or resetting replicas), a stopped notebook could unexpectedly start -- wasting resources and potentially exposing workloads that were intentionally shut down.

This change adds `test_upgrade_stopped.py` with the following tests:

| Phase | Test | What it verifies |
|-------|------|------------------|
| Pre-upgrade | `test_stopped_notebook_has_zero_replicas` | StatefulSet replicas=0 after stop |
| Pre-upgrade | `test_stopped_notebook_pod_absent` | No pod exists for stopped notebook |
| Post-upgrade | `test_stopped_annotation_preserved_after_upgrade` | `kubeflow-resource-stopped` annotation still present |
| Post-upgrade | `test_stopped_annotation_value_unchanged_after_upgrade` |  `kubeflow-resource-stopped` annotation timestamp value not modified during upgrade   |
| Post-upgrade | `test_stopped_notebook_still_has_zero_replicas` | StatefulSet still at replicas=0 |
| Post-upgrade | `test_stopped_notebook_pod_absent_after_upgrade` | No pod unexpectedly (re)created |

New session-scoped fixtures in `conftest.py`:

- `stopped_notebook_pvc` / `stopped_notebook` / `stopped_notebook_statefulset` -- resource wrappers
- `stopped_notebook_pre_upgrade_shutdown` -- orchestrates the stop sequence (wait Ready, add annotation with real UTC timestamp, wait for pod termination, verify StatefulSet replicas=0)

Additional improvements:
- Simplified pod wait error handling by removing unnecessary inner try/except blocks (`.exists` and `collect_pod_information` do not raise exceptions by design)
- Stop annotation uses a real UTC timestamp matching production behavior
- Annotation value is persisted in the baseline ConfigMap and verified unchanged post-upgrade

## Test plan

- [x] `pytest --collect-only --pre-upgrade tests/workbenches/notebooks_server/controller/upgrade/` -- 7 tests collected
- [x] `pytest --collect-only --post-upgrade tests/workbenches/notebooks_server/controller/upgrade/` -- 14 tests collected
- [x] Deploy notebooks on a pre-upgrade cluster, run pre-upgrade tests, upgrade the platform, run post-upgrade tests
```
uv run pytest tests/workbenches/notebooks_server/controller/upgrade -v --tb=short --junitxml=/tmp/junit-workbenches.xml -o console_output_style=count --pre-upgrade
uv run pytest tests/workbenches/notebooks_server/controller/upgrade -v --tb=short --junitxml=/tmp/junit-workbenches.xml -o console_output_style=count --post-upgrade
```

- [x] Locally
- [ ] Jenkins

## Additional Requirements

## Summary by CodeRabbit

* **Tests**
  * Added test coverage and supporting fixtures for an “upgrade while notebook is stopped” scenario.
  * Pre-upgrade checks ensure stopped notebooks are scaled down and pods are absent.
  * Baseline capture now records the stopped annotation/timestamp.
  * Post-upgrade checks verify the stopped annotation value persists, replicas remain zero, and pods are not recreated.

### DIFF
--- a/tests/workbenches/notebooks_server/controller/upgrade/conftest.py
+++ b/tests/workbenches/notebooks_server/controller/upgrade/conftest.py
@@ -155,18 +155,8 @@ def upgrade_notebook_pod(
             timeout=Timeout.TIMEOUT_5MIN,
         )
     except (TimeoutError, TimeoutExpiredError, RuntimeError) as e:
-        try:
-            pod_exists = notebook_pod.exists
-        except Exception as exists_error:  # noqa: BLE001
-            LOGGER.warning(f"Failed to verify pod existence after timeout: {exists_error}")
-            pod_exists = False
-
-        if pod_exists:
-            try:
-                collect_pod_information(notebook_pod)
-            except Exception as collect_error:  # noqa: BLE001
-                LOGGER.warning(f"Failed to collect pod artifacts: {collect_error}")
-
+        if notebook_pod.exists:
+            collect_pod_information(notebook_pod)
             raise AssertionError(
                 f"Pod '{upgrade_notebook.name}-0' failed to reach Ready state "
                 f"within {Timeout.TIMEOUT_5MIN} seconds.\n"
--- a/tests/workbenches/notebooks_server/controller/upgrade/conftest.py
+++ b/tests/workbenches/notebooks_server/controller/upgrade/conftest.py
@@ -1,5 +1,6 @@
 import json
 from collections.abc import Generator
+from datetime import UTC, datetime
 from typing import Any
 
 import pytest
@@ -30,6 +31,7 @@ LOGGER = structlog.get_logger(name=__name__)
 
 UPGRADE_NAMESPACE = "upgrade-workbenches"
 UPGRADE_NOTEBOOK_NAME = "upgrade-workbenches"
+UPGRADE_STOPPED_NOTEBOOK_NAME = "upgrade-wb-stopped"
 UPGRADE_BASELINE_CM_NAME = "upgrade-workbenches-baseline"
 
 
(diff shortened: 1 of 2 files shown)


tests/workbenches/notebooks_server/controller/upgrade/conftest.py
```diff
@@ -232,6 +233,146 @@ def upgrade_notebook_reference_grant(
...
+) -> Generator[PersistentVolumeClaim, Any, Any]:
+    """PVC for the stopped notebook upgrade scenario."""
+    pvc_kwargs = {
+        "client": unprivileged_client,
+        "name": UPGRADE_STOPPED_NOTEBOOK_NAME,
+        "namespace": upgrade_notebook_namespace.name,
+    }
+
+    if pytestconfig.option.post_upgrade:
+        yield PersistentVolumeClaim(**pvc_kwargs)
```
dbasunag, line 242: Should not this be cleaned after post_upgrade?

tests/workbenches/notebooks_server/controller/upgrade/conftest.py
```diff
@@ -232,6 +233,146 @@ def upgrade_notebook_reference_grant(
...
+            status=Pod.Condition.Status.TRUE,
+            timeout=Timeout.TIMEOUT_5MIN,
+        )
+    except (TimeoutError, TimeoutExpiredError, RuntimeError) as e:
+        try:
+            pod_exists = notebook_pod.exists
+        except Exception as exists_error:  # noqa: BLE001
+            LOGGER.warning(f"Failed to verify pod existence after timeout: {exists_error}")
+            pod_exists = False
+
```
dbasunag: this code path won't execute.

tests/workbenches/notebooks_server/controller/upgrade/conftest.py
```diff
@@ -232,6 +233,146 @@ def upgrade_notebook_reference_grant(
...
+        if pod_exists:
+            try:
+                collect_pod_information(notebook_pod)
+            except Exception as collect_error:  # noqa: BLE001
+                LOGGER.warning(f"Failed to collect pod artifacts: {collect_error}")
+
```
dbasunag: Same comment.

harshad16 (commented):
lgtm overall
Thanks for the changes , it is the intended behaviour.

small nit-pick which i would say is up-to you to either change here or in upcoming PR or push back.

1. value of annotation is being set `upgrade-wb-stopped` , as in the controller original it is most cases either a timestamp string or odh-notebook-controller-lock. My suggestion would either to use the timestamp string as value or atleast add the comment at line 33, stating that this is only for testing case and wont not be seen in overall behavior in production execution of notebook-controller.
stating this as i feel any new contributor can get confused thing, can we use any value. even though they can it would just cause confusion for most.

2. In the new file `test_upgrade_stopped.py` , in function there is the parameter set :
`unprivileged_client,` , in most func where this showed up, we have type defined for this, so lets do the same `unprivileged_client: DynamicClient,`.

wdyt @jstourac ?

tests/workbenches/notebooks_server/controller/upgrade/conftest.py
jstourac: @dbasunag for PVCs we rely on the deletion of these via the overall namespace deletion. So we explicitly delete only the Notebook CR as it has some finalizers and other dependent resources. I presume that we can cleanup PVC too. I'm slightly keen to keep this as is (no explicit PVC deletion), although - if you feel it would be better and safer to change, I don't mind updating it - with this change, I would update it also for the other fixture where PVC is created. Please, let me know if you believe this should be changed.

tests/workbenches/notebooks_server/controller/upgrade/conftest.py
jstourac: @dbasunag ah, good catch 🤦 , thank you! I'll update the same for the other fixture too in a separate commit so it's visible fix.

tests/workbenches/notebooks_server/controller/upgrade/conftest.py
jstourac: @dbasunag, you are right, thank you - same answer as above!

jstourac:
@dbasunag thank you for your review! I addressed those exception handling issue and also provided a fix for the other already existing instance of this in the code. Regarding the PVC cleanup - I replied to you, please, let me know what you think.

@harshad16 thank you for your review! I actually addressed both of your comments and during that I realized we should also check that the value in the stopped annotation hasn't been changed during the upgrade, so I added it as a new check too.

All the changes can be seen via [this compare](https://github.com/opendatahub-io/opendatahub-tests/compare/4cf4f68cf86dde9dec57ee067af3e999fb867607..599e3f6e21763635ec9c82549e6b5667be12c78c). The other compare is just a rebase against the main branch.

dbasunag (approved):

harshad16 (approved):
thanks for addressing the changes.
Everything looks great

/lgtm

PR #1659, opened 2026-06-04
test(workbenches): split ImageStream health checks for disconnected cluster support

jstourac:
## Summary

On disconnected (air-gapped) clusters, only the latest 2 ImageStream version tags are mirrored per the customer guidance. This causes the existing `test_workbench_imagestreams_health` test to fail for older tags that are intentionally not present.

This PR splits the notebook ImageStream tag validation into two tests:

- **`test_workbench_imagestreams_health`** — validates ImageStream count and health of the latest 2 version tags. Runs on all clusters (connected and disconnected).
- **`test_workbench_imagestreams_older_tags_health`** — validates older tags (beyond the latest 2). Marked `skip_on_disconnected` so it is deselected by CI on disconnected clusters.

Trainer and runtime ImageStreams continue to validate all tags unconditionally.

## Details

- Added `_get_latest_n_tag_names` helper that sorts tags by version, handling the three historical naming schemes:
  - Current RHOAI versions (>= 3.4, e.g. 3.4, 3.5)
  - Year-based versions (e.g. 2024.1, 2025.2)
  - Legacy versions (e.g. 1.2)
- Refactored `_validate_imagestreams_with_label` to accept an optional `tags_to_validate` filter
- No new dependencies — `packaging` is already available transitively

## Test plan

- [x] Run on a connected cluster: all 4 test cases pass (3 health + 1 older-tags)
- [x] Run with `-m "not skip_on_disconnected"`: only 3 health tests are collected (older-tags tests deselected)
- [x] Run on a disconnected cluster: the 3 health tests pass (only latest 1 tags validated for notebook)

```
uv run pytest tests/workbenches/notebooks_server/operator/test_imagestream_health.py -v
uv run pytest tests/workbenches/notebooks_server/operator/test_imagestream_health.py -v -m "not skip_on_disconnected"
```

## Related Issues

N/A

- [x] Locally
- [ ] Jenkins

## Additional Requirements

## Summary by CodeRabbit

* **Tests**
  * Enhanced ImageStream health tests to support selective tag validation across mixed naming schemes.
  * Added version-aware ranking to identify and validate the latest N tags and an option to validate only those.
  * Added a companion test to explicitly validate older tags and fail if no older tags are present.

### DIFF
--- a/tests/workbenches/notebooks_server/operator/test_imagestream_health.py
+++ b/tests/workbenches/notebooks_server/operator/test_imagestream_health.py
@@ -6,11 +6,63 @@ import pytest
 import structlog
 from kubernetes.dynamic import DynamicClient
 from ocp_resources.image_stream import ImageStream
+from packaging.version import InvalidVersion, Version
 from pytest_testconfig import config as py_config
 
 pytestmark = [pytest.mark.smoke]
 LOGGER = structlog.get_logger(name=__name__)
 IMPORT_SUCCESS_CONDITION_TYPE = "ImportSuccess"
+LATEST_TAGS_COUNT = 2
+
+
+RHOAI_VERSIONING_START = Version(version="3.4")
+
+
+def _tag_sort_key(version: Version) -> tuple[int, Version]:
+    """Return a sort key reflecting the chronological ordering of ImageStream tag versions.
+
+    Three naming schemes have been used historically:
+      - Tier 2 (newest): RHOAI major.minor >= 3.4 (e.g. 3.4, 3.5)
+      - Tier 1 (middle): year-based versions with major >= 2000 (e.g. 2024.1, 2025.2)
+      - Tier 0 (oldest): legacy versions before the year scheme (e.g. 1.2)
+    """
+    if version >= RHOAI_VERSIONING_START and version.major < 2000:
+        return (2, version)
+    if version.major >= 2000:
+        return (1, version)
+    return (0, version)
+
+
+def _get_latest_n_tag_names(imagestream_data: dict[str, Any], n: int = LATEST_TAGS_COUNT) -> set[str]:
+    """Return the names of the N most recent tags by version.
+
+    Handles mixed version schemes: RHOAI versions (3.4, 3.5) are ranked above
+    legacy year-based versions (2025.2, 2025.1, 2024.2, ...).
+
+    Args:
+        imagestream_data: Raw ImageStream dict (from `instance.to_dict()`).
+        n: Number of latest tags to return.
+
+    Returns:
+        Set of tag name strings for the N highest-versioned tags.
+    """
+    spec_tags = imagestream_data.get("spec", {}).get("tags", [])
+    all_tag_names = {str(tag.get("name")) for tag in spec_tags if tag.get("name")}
+    versioned: list[tuple[tuple[int, Version], str]] = []
+    for tag in spec_tags:
+        name = tag.get("name")
+        if not name:
+            continue
+        try:
+            version = Version(version=name)
+        except InvalidVersion:
+            continue
+        versioned.append((_tag_sort_key(version=version), name))
+    versioned.sort(reverse=True)
+    if not versioned:
+        # Fail safe: never return an empty filter that disables validation.
+        return all_tag_names
+    return {name for _, name in versioned[:n]}
 
 
 def _validate_imagestream_tag_health(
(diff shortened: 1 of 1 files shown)


jiridanek (dismissed):

tests/workbenches/notebooks_server/operator/test_imagestream_health.py
```diff
@@ -211,8 +278,75 @@ def test_workbench_imagestreams_health(
...
+        latest_tags = _get_latest_n_tag_names(imagestream_data=imagestream_data)
+        all_tag_names = {
+            str(spec_tag.get("name"))
+            for spec_tag in imagestream_data.get("spec", {}).get("tags", [])
+            if spec_tag.get("name")
+        }
+        older_tags |= all_tag_names - latest_tags
+
+    if not older_tags:
+        pytest.skip("No older tags found beyond the latest 2 versions")
```
dbasunag: Based on the lasted management ask on no skip, we should consider failing here.

harshad16 (commented):
@jstourac in the description, you have stated: 
`On disconnected (air-gapped) clusters, only the latest 2 ImageStream version tags are mirrored per the customer guidance. ...`
based on the disconnected-helper, doesnt the user mirror image based on this: https://github.com/red-hat-data-services/rhoai-disconnected-install-helper/blob/main/rhoai-3.4.md

so if a user in past mirrored x images, now in latest they would have x + these new ones https://github.com/red-hat-data-services/rhoai-disconnected-install-helper/blob/main/rhoai-3.4.md#additional-images , right ?

jstourac:
> @jstourac in the description, you have stated: `On disconnected (air-gapped) clusters, only the latest 2 ImageStream version tags are mirrored per the customer guidance. ...` based on the disconnected-helper, doesnt the user mirror image based on this: https://github.com/red-hat-data-services/rhoai-disconnected-install-helper/blob/main/rhoai-3.4.md
> 
> so if a user in past mirrored x images, now in latest they would have x + these new ones https://github.com/red-hat-data-services/rhoai-disconnected-install-helper/blob/main/rhoai-3.4.md#additional-images , right ?

This is correct, Harshad. The thing is, that we no longer mirror the unsupported images section in our testing environment. As such, only N version image (image hashes baked directly in the product) + N-1 version image (image hashes are in the additional images section of the file you referenced above) are present in our testing environment. Older images aren't present there with our current setup. This change

Of course - if customer installed e.g. RHOAI 3.0 in their disconnected environment and mirrored all the images, then with each upgrade to a newer version, the latest images are mirrored too and as a result all tags are populated in imagestreams in his disconnected environment. Though, this doesn't apply for a clean installations.

Our [current documentation for RHOAI 3.4](https://docs.redhat.com/en/documentation/red_hat_openshift_ai_self-managed/3.4/html-single/installing_and_uninstalling_openshift_ai_self-managed_in_a_disconnected_environment/index#mirroring-images-to-a-private-registry-for-a-disconnected-installation_install) states with regards of this:

> Open the example image set configuration file (rhoai-<version>.md) from the [disconnected installer helper](https://github.com/red-hat-data-services/rhoai-disconnected-install-helper/tree/main) repository and examine its contents.

> The disconnected installer helper file includes a list of Additional images required to install OpenShift AI in a disconnected environment, as well as a list of older Unsupported images provided for reference only. These older images are no longer maintained by Red Hat but are included for convenience, such as when importing older resources or maintaining compatibility with previous environments.

Also, as you can see, the example [ImageSetConfiguration](https://github.com/red-hat-data-services/rhoai-disconnected-install-helper/blob/main/rhoai-3.4.md#imagesetconfiguration-example) doesn't contain those unsupported images.

So, based on this, devtestops updated our disconnected install so that unsupported images aren't mirrored and as such we see these errors. That is why I am trying to update our test so that it counts with this situation as expected.

tests/workbenches/notebooks_server/operator/test_imagestream_health.py
jstourac: Initially I didn't want to do this change. But your comment made me realize that runtime images have just one tag anyway, so they shouldn't be part of this new approach at all. Also, I didn't want to fail here as in case we introduce a new image with just a new tag, that would mean a standard situation. But then I realized again, that with the way this test is written, introducing a new image with just one tag is okay here. We would have to ditch all the older tags from all the workbench images imagestreams. So, yep, you are right, let's fail here. I'll push all my updates shared here soonish.

jstourac:
@harshad16 @dbasunag @jiridanek, thank you for your reviews. I have pushed a change addressing your review. Diff can be seen via [this comparation](https://github.com/opendatahub-io/opendatahub-tests/compare/3035275001a3595f6a0047edcb45341937b3e59c..bd4c5155c53ed1bf340fbe507149426262014430). The other force push is just a rebase.

jiridanek (approved):

dbasunag (approved):

PR #1663, opened 2026-06-04
test(workbenches): add kube-rbac-proxy auth resource upgrade tests

jstourac:

## Summary

- Adds pre- and post-upgrade tests verifying that the kube-rbac-proxy authentication resources (sidecar container, Service, ConfigMap, ClusterRoleBinding) survive a platform upgrade for both running and stopped notebooks.
- Introduces session-scoped fixtures for the auth-related resources of both the running and stopped notebook.

## Details

When `notebooks.opendatahub.io/inject-auth: "true"` is set on a Notebook CR, the ODH webhook injects a kube-rbac-proxy sidecar and the controller reconciles supporting resources (Service on port 8443, authorization ConfigMap, auth-delegator ClusterRoleBinding). If any of these are lost during upgrade, authenticated access to notebooks breaks.

This change also verifies that auth resources are preserved for a **stopped** notebook (replicas=0), catching the scenario where an upgraded controller might incorrectly garbage-collect resources for notebooks that aren't running.

This change adds `test_upgrade_auth.py` with the following tests:

| Phase | Test | What it verifies |
|-------|------|------------------|
| Pre-upgrade | `test_kube_rbac_proxy_sidecar_present` | Running pod has `kube-rbac-proxy` container |
| Pre-upgrade | `test_kube_rbac_proxy_service_exists` | Running notebook's proxy Service exists |
| Pre-upgrade | `test_kube_rbac_proxy_configmap_exists` | Running notebook's proxy ConfigMap exists |
| Pre-upgrade | `test_auth_delegator_cluster_role_binding_exists` | Running notebook's auth-delegator CRB exists |
| Pre-upgrade | `test_stopped_kube_rbac_proxy_service_exists` | Stopped notebook's proxy Service exists |
| Pre-upgrade | `test_stopped_kube_rbac_proxy_configmap_exists` | Stopped notebook's proxy ConfigMap exists |
| Pre-upgrade | `test_stopped_auth_delegator_crb_exists` | Stopped notebook's auth-delegator CRB exists |
| Post-upgrade | `test_kube_rbac_proxy_sidecar_present_after_upgrade` | Sidecar still in running pod |
| Post-upgrade | `test_kube_rbac_proxy_service_exists_after_upgrade` | Running notebook's Service survived |
| Post-upgrade | `test_kube_rbac_proxy_configmap_exists_after_upgrade` | Running notebook's ConfigMap survived |
| Post-upgrade | `test_auth_delegator_cluster_role_binding_exists_after_upgrade` | Running notebook's CRB survived |
| Post-upgrade | `test_stopped_kube_rbac_proxy_service_exists_after_upgrade` | Stopped notebook's Service survived |
| Post-upgrade | `test_stopped_kube_rbac_proxy_configmap_exists_after_upgrade` | Stopped notebook's ConfigMap survived |
| Post-upgrade | `test_stopped_auth_delegator_crb_exists_after_upgrade` | Stopped notebook's CRB survived |

New session-scoped fixtures in `conftest.py`:
- `auth_proxy_service` / `stopped_auth_proxy_service` -- kube-rbac-proxy Service wrappers
- `auth_proxy_configmap` / `stopped_auth_proxy_configmap` -- kube-rbac-proxy ConfigMap wrappers
- `auth_delegator_crb` / `stopped_auth_delegator_crb` -- auth-delegator ClusterRoleBinding wrappers

## Test plan

- [x] Deploy notebooks with `inject-auth: "true"` on a pre-upgrade cluster, stop one, run pre-upgrade tests, run post-upgrade tests

- [x] Locally
- [ ] Jenkins

## Additional Requirements

## Summary by CodeRabbit

* **Tests**
  * Added session fixtures to supply authentication-proxy Service, ConfigMap, and auth-delegator role binding resources for upgrade tests.
  * Added pre- and post-upgrade tests confirming these resources persist for both running and stopped notebook servers.
  * Added assertions that the proxy exposes the expected port, the role binding references the expected role, and the proxy sidecar remains for running notebooks.
  * Updated a test failure message formatting for clearer route name display.

### DIFF
--- a/tests/workbenches/notebooks_server/controller/upgrade/conftest.py
+++ b/tests/workbenches/notebooks_server/controller/upgrade/conftest.py
@@ -6,6 +6,7 @@ from typing import Any
 import pytest
 import structlog
 from kubernetes.dynamic import DynamicClient
+from ocp_resources.cluster_role_binding import ClusterRoleBinding
 from ocp_resources.config_map import ConfigMap
 from ocp_resources.namespace import Namespace
 from ocp_resources.notebook import Notebook
@@ -221,6 +222,82 @@ def upgrade_notebook_reference_grant(
         client=admin_client,
         name="notebook-httproute-access",
         namespace=upgrade_notebook_namespace.name,
+    )
+
+
+@pytest.fixture(scope="session")
+def auth_proxy_service(
+    unprivileged_client: DynamicClient,
+    upgrade_notebook: Notebook,
+) -> Service:
+    """kube-rbac-proxy Service for the notebook."""
+    return Service(
+        client=unprivileged_client,
+        name=f"{upgrade_notebook.name}-kube-rbac-proxy",
+        namespace=upgrade_notebook.namespace,
+    )
+
+
+@pytest.fixture(scope="session")
+def auth_proxy_configmap(
+    unprivileged_client: DynamicClient,
+    upgrade_notebook: Notebook,
+) -> ConfigMap:
+    """kube-rbac-proxy ConfigMap for the notebook."""
+    return ConfigMap(
+        client=unprivileged_client,
+        name=f"{upgrade_notebook.name}-kube-rbac-proxy-config",
+        namespace=upgrade_notebook.namespace,
+    )
+
+
+@pytest.fixture(scope="session")
+def auth_delegator_crb(
+    admin_client: DynamicClient,
+    upgrade_notebook: Notebook,
+) -> ClusterRoleBinding:
+    """auth-delegator ClusterRoleBinding for the notebook's kube-rbac-proxy."""
+    return ClusterRoleBinding(
+        client=admin_client,
+        name=f"{upgrade_notebook.name}-rbac-{upgrade_notebook.namespace}-auth-delegator",
+    )
+
+
+@pytest.fixture(scope="session")
+def stopped_auth_proxy_service(
+    unprivileged_client: DynamicClient,
+    stopped_notebook: Notebook,
+) -> Service:
+    """kube-rbac-proxy Service for the stopped notebook."""
+    return Service(
+        client=unprivileged_client,
+        name=f"{stopped_notebook.name}-kube-rbac-proxy",
+        namespace=stopped_notebook.namespace,
+    )
+
+
+@pytest.fixture(scope="session")
+def stopped_auth_proxy_configmap(
+    unprivileged_client: DynamicClient,
+    stopped_notebook: Notebook,
+) -> ConfigMap:
+    """kube-rbac-proxy ConfigMap for the stopped notebook."""
+    return ConfigMap(
+        client=unprivileged_client,
+        name=f"{stopped_notebook.name}-kube-rbac-proxy-config",
+        namespace=stopped_notebook.namespace,
+    )
+
+
+@pytest.fixture(scope="session")
+def stopped_auth_delegator_crb(
+    admin_client: DynamicClient,
+    stopped_notebook: Notebook,
+) -> ClusterRoleBinding:
+    """auth-delegator ClusterRoleBinding for the stopped notebook's kube-rbac-proxy."""
+    return ClusterRoleBinding(
+        client=admin_client,
+        name=f"{stopped_notebook.name}-rbac-{stopped_notebook.namespace}-auth-delegator",
     )
 
 
--- a/tests/workbenches/notebooks_server/controller/upgrade/test_upgrade_auth.py
+++ b/tests/workbenches/notebooks_server/controller/upgrade/test_upgrade_auth.py
@@ -1,0 +1,277 @@
+import pytest
+from ocp_resources.cluster_role_binding import ClusterRoleBinding
+from ocp_resources.config_map import ConfigMap
+from ocp_resources.pod import Pod
+from ocp_resources.service import Service
+
+KUBE_RBAC_PROXY_CONTAINER = "kube-rbac-proxy"
+KUBE_RBAC_PROXY_PORT = 8443
+AUTH_DELEGATOR_ROLE = "system:auth-delegator"
+
+
+@pytest.mark.usefixtures("capture_notebook_baseline")
+class TestPreUpgradeNotebookAuth:
+    """Verify kube-rbac-proxy auth resources exist before the platform upgrade.
+
+    Steps:
+        1. Verify the notebook pod has a kube-rbac-proxy sidecar container.
+        2. Verify the kube-rbac-proxy Service exists.
+        3. Verify the kube-rbac-proxy ConfigMap exists.
+        4. Verify the auth-delegator ClusterRoleBinding exists.
+    """
+
+    @pytest.mark.pre_upgrade
+    def test_kube_rbac_proxy_sidecar_present(
+        self,
+        upgrade_notebook_pod: Pod,
+    ) -> None:
+        """Given a notebook with inject-auth annotation,
+        When the ODH webhook injects the sidecar,
+        Then the pod should have a kube-rbac-proxy container.
+        """
+        containers = upgrade_notebook_pod.instance.spec.containers
+        container_names = [container.name for container in containers]
+
+        assert KUBE_RBAC_PROXY_CONTAINER in container_names, (
+            f"Pod '{upgrade_notebook_pod.name}' missing '{KUBE_RBAC_PROXY_CONTAINER}' sidecar. "
+            f"Containers: {container_names}"
+        )
+
+    @pytest.mark.pre_upgrade
+    def test_kube_rbac_proxy_service_exists(
+        self,
+        auth_proxy_service: Service,
+    ) -> None:
+        """Given a notebook with inject-auth annotation,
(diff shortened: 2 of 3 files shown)


tests/workbenches/notebooks_server/controller/upgrade/test_upgrade_auth.py
```diff
@@ -0,0 +1,228 @@
...
+    def test_kube_rbac_proxy_sidecar_present_after_upgrade(
+        self,
+        upgrade_notebook_pod: Pod,
+    ) -> None:
+        """Given a notebook with auth sidecar existed before upgrade,
+        When the upgrade completes,
+        Then the pod should still have the kube-rbac-proxy container.
+        """
+        containers = upgrade_notebook_pod.instance.spec.containers
+        container_names = [c.name for c in containers]
```
dbasunag: If we can avoid using one letter variables, that would be great.

tests/workbenches/notebooks_server/controller/upgrade/test_upgrade_auth.py
```diff
@@ -0,0 +1,228 @@
...
+    """Verify kube-rbac-proxy auth resources exist for a stopped notebook before upgrade.
+
+    Steps:
+        1. Verify the kube-rbac-proxy Service exists for the stopped notebook.
+        2. Verify the kube-rbac-proxy ConfigMap exists for the stopped notebook.
+        3. Verify the auth-delegator ClusterRoleBinding exists for the stopped notebook.
+    """
+
+    @pytest.mark.pre_upgrade
+    def test_stopped_kube_rbac_proxy_service_exists(
```
dbasunag, line 96: if you are validating existence of the resource, you can parameterize the test, by passing the name and namespace(if any) and the fixture can read the name and namespace name and set kwargs to the resource call. e.g.
```
@pytest.fixture(scope="function")
  def resource(request):
      """Kubernetes resource from request params."""
      kwargs = {"name": request.param["name"]}
      if request.param.get("namespace"):
          kwargs["namespace"] = request.param["namespace"]
      resource_class = request.param["class"]
      return resource_class(**kwargs)
```
Then tests can be like 
```
@pytest.mark.parametrize(
      "resource",
      [
          pytest.param({"class": ClusterRoleBinding, "name": "name1"}, id="test_crb"),
          pytest.param({"class": ConfigMap, "name": "my-cm", "namespace": "my-namespace"}, id="test_cm"),
      ],
      indirect=True,
  )
  def test_something(resource):
      ...
```
Same can be done for post upgrade. 
Your majority tests would be reusing code and code would be cleaner this way. Of course the one checking container needs to be different.

tests/workbenches/notebooks_server/controller/upgrade/test_upgrade_auth.py
jstourac: Good catch, thank you - updated together with some more such ocurrences.

tests/workbenches/notebooks_server/controller/upgrade/test_upgrade_auth.py
dbasunag: @jstourac bringing your attention to this one. Please let me know if it makes sense.

jstourac: @dbasunag thank you :slightly_smiling_face: I'm aware of it, just haven't energy to think about it yet. I'll try to assess on Monday and post update here. Thanks for your great insights!

jstourac: @dbasunag okay, I looked into this one and thought about it. Your idea is good, but to be honest, I tend to keep the code as is for now. Why I think it's slightly better not to parametrize:
1. With recent changes pushed to this branch, some of the fixtures that could be easily parametrized before are note anymore as they now check not only the resource presence, but other metadata too (e.g. service port).
2. With the parametrization, the tests may not be as clear as they are now during the execution logging and eventual failure check.

First reason is kind of strong opinion, the second reason much less. Still, unless you truly prefer to introduce the parametrization here, I would keep it as is for now.

tests/workbenches/notebooks_server/controller/upgrade/test_upgrade_auth.py
```diff
@@ -0,0 +1,266 @@
...
+    @pytest.mark.pre_upgrade
+    def test_kube_rbac_proxy_service_exists(
+        self,
+        auth_proxy_service: Service,
+    ) -> None:
+        """Given a notebook with inject-auth annotation,
+        When the ODH controller reconciles,
+        Then a kube-rbac-proxy Service should exist.
+        """
+        assert auth_proxy_service.exists, f"kube-rbac-proxy Service '{auth_proxy_service.name}' does not exist"
```
harshad16, line 49:
```suggestion
        """
        assert auth_proxy_service.exists, f"kube-rbac-proxy Service '{auth_proxy_service.name}' does not exist"
        
        port_numbers = [port.port for port in auth_proxy_service.instance.spec.ports]
    assert KUBE_RBAC_PROXY_PORT in port_numbers, (
        f"Service '{auth_proxy_service.name}' missing port {KUBE_RBAC_PROXY_PORT}. "
        f"Found ports: {port_numbers}"
    )   
```

Pre-upgrade Service tests only assert .exists, but post-upgrade also checks port 8443. Controller always uses 8443 and routing tests depend on it — please add the same port assertion pre-upgrade (running + stopped) for symmetry.

tests/workbenches/notebooks_server/controller/upgrade/test_upgrade_auth.py
```diff
@@ -0,0 +1,266 @@
...
+        self,
+        stopped_auth_proxy_service: Service,
+    ) -> None:
+        """Given a stopped notebook with inject-auth annotation,
+        When the notebook is stopped,
+        Then the kube-rbac-proxy Service should still exist.
+        """
+        assert stopped_auth_proxy_service.exists, (
+            f"kube-rbac-proxy Service '{stopped_auth_proxy_service.name}' does not exist for stopped notebook"
+        )
```
harshad16, line 106: Lets also add the same port checks in here for consistency

```suggestion
        )
        
        port_numbers = [port.port for port in stopped_auth_proxy_service.instance.spec.ports]
        assert KUBE_RBAC_PROXY_PORT in port_numbers, (
            f"Service '{stopped_auth_proxy_service.name}' lost port {KUBE_RBAC_PROXY_PORT} after upgrade. "
            f"Found ports: {port_numbers}"
        )
```

tests/workbenches/notebooks_server/controller/upgrade/conftest.py
```diff
@@ -224,6 +225,82 @@ def upgrade_notebook_reference_grant(
     )
 
 
+@pytest.fixture(scope="session")
+def auth_proxy_service(
+    admin_client: DynamicClient,
```
harshad16: were we not using `unprivileged_client` ? 
auth_proxy_service / auth_proxy_configmap use admin_client, but upgrade_notebook_service uses unprivileged_client for the same namespace.

harshad16 (changes requested):
Thanks for the auth upgrade coverage — naming matches the controller and complements routing tests well.

I have few small request.

harshad16 (approved):
Thank you for the changes.
everything looks good, approving it, only thing remaining is addressing the comment:
https://github.com/opendatahub-io/opendatahub-tests/pull/1663#discussion_r3358952129

/lgtm

tests/workbenches/notebooks_server/controller/upgrade/conftest.py
jstourac: good catch, hopefully updated everywhere where possible, thanks!

tests/workbenches/notebooks_server/controller/upgrade/test_upgrade_auth.py
jstourac: good point, added

tests/workbenches/notebooks_server/controller/upgrade/test_upgrade_auth.py
jstourac: you are right, added

jstourac:
Thank you for your review, @harshad16 ! I addressed your comments and you already re-reviewed, cool! :slightly_smiling_face: 

Regarding the parametrization thread you mentioned, as I stated in my comment, I would keep it the way it is for now. @dbasunag any objections?

dbasunag (approved):

tests/workbenches/notebooks_server/controller/upgrade/test_upgrade_auth.py
dbasunag: With latest pr, parameterization doesn't make sense. Resolving.

jstourac:
/cherry-pick 3.4

The cherry-pick workflow doesn't use `-x` when backporting the commit, so the original link is missed in the standard git log and one has to go to github again... but it has other advantage - I can simply merge the changes without waiting for somebody else to perform a review.

PR #1667, opened 2026-06-05
Replace exact snapshot comparison with fuzzy validation for Predictive Runtimes

amehtaja:
## Summary
  Fixes assertion errors in model inference tests (Triton, OVMS) that fail on different GPU hardware types due to numerical precision differences.

  ## Problem
model inference tests use exact snapshot comparison (`assert response == response_snapshot`) which fails when running on different GPU hardware:
  - **NVIDIA GPUs** produce: `[0.9234567, 0.0123456, ...]`
  - **AMD GPUs** produce: `[0.9234568, 0.0123455, ...]`
  - **Gaudi accelerators** produce slightly different values
  - **CPU inference** produces different values

  Even though the model predictions are correct (same top-5 classes), tests fail due to tiny float differences in the output array.

  ## Root Cause
  Tests performed exact equality comparison on inference responses:
  ```python
  assert response == response_snapshot  # ❌ Fails on different hardware
```
  This causes false-negative test failures when the same model runs on different accelerator types, even though the predictions are structurally correct and semantically identical.

  ## Solution

  Implemented fuzzy validation (similar to existing VLLM inference tests) that validates response structure and correctness without comparing exact float values:

  JIRA

  Link : https://redhat.atlassian.net/browse/RHOAIENG-66478

## Summary by CodeRabbit

* **Tests**
  * Switched model-serving inference checks (OpenVINO, Triton, MLServer) from strict snapshot matching to flexible validations that ensure responses include outputs, non-empty output data, and sufficient elements/top-k indices.
  * Removed several OpenVINO inference snapshot artifacts for multiple deployment configurations.

### CODE REVIEW

**Review by dbasunag — DISMISSED** [review]

File: tests/model_serving/model_runtime/openvino/utils.py [hunk 1]
```diff
@@ -118,7 +118,16 @@ def validate_inference_request(
         model_version=model_version,
     )
 
-    assert response == response_snapshot, f"Output mismatch: {response} != {response_snapshot}"
```
Snomaan6846, line 121: It seems Handling is the same for both the runtimes openvino and triton. I believe it would be a good idea to move it to a helper function and reuse it.

tests/model_serving/model_runtime/triton/basic_model_deployment/utils.py
```diff
@@ -144,7 +144,17 @@ def validate_inference_request(
         protocol=protocol,
         root_dir=root_dir,
     )
-    assert response == response_snapshot, f"Output mismatch: {response} != {response_snapshot}"
+
+    assert response, "Response is empty"
```
Snomaan6846, line 148: Same comment as above...

tests/model_serving/model_runtime/openvino/utils.py
```diff
@@ -118,7 +118,16 @@ def validate_inference_request(
...
+    assert response, "Response is empty"
```
Snomaan6846, line 121: I remember that we planned to keep the snapshots for the references. Is there any change of plan?

Snomaan6846:
It seems the MLServer runtime handling is missing.

tests/model_serving/model_runtime/openvino/utils.py
Raghul-M: @Snomaan6846 what you been references ?

Raghul-M (approved):

mwaykole (approved):

PR #1698, opened 2026-06-08
fix: Replace MariaDB Operator with standalone Deployment + TLS

sheltoncyril:
## Summary
- Replace MariaDB Operator-based deployment with standalone Kubernetes Deployment using Red Hat registry image (`registry.redhat.io/rhel9/mariadb-1011`) to avoid Docker Hub rate limits
- Generate self-signed TLS certificates in Python using `cryptography` library and configure MariaDB with SSL/TLS support
- Fix DB migration test race condition by waiting for route readiness after pod restart

## Test plan
- [x] All 66 TrustyAI tests passed on cluster (2.25 branch)
- [ ] Run TrustyAI test suite on main

## Summary by CodeRabbit

* **Tests**
  * Switched tests to use standalone TLS-enabled MariaDB deployments instead of operator-backed DBs for simpler, more reliable isolation.
  * Updated multi-namespace provisioning and CA handling to use per-namespace standalone DBs and fixed CA secret names.
  * Added provisioning helpers (certificates, service, storage), pod readiness waits, and route readiness checks before health/metric assertions.
  * Broadened DB connection-error matching to cover more TLS/connection variants.

* **Chores**
  * Added cryptography dependency for TLS certificate handling in tests.

### DIFF
--- a/tests/ai_safety/trustyai_service/utils.py
+++ b/tests/ai_safety/trustyai_service/utils.py
@@ -30,8 +30,7 @@ from utilities.general import validate_container_images, wait_for_pods_by_labels
 LOGGER = structlog.get_logger(name=__name__)
 
 MARIADB_IMAGE = (
-    "registry.redhat.io/rhel9/mariadb-1011"
-    "@sha256:092407d87f8017bb444a462fb3d38ad5070429e94df7cf6b91d82697f36d0fa9"
+    "registry.redhat.io/rhel9/mariadb-1011@sha256:092407d87f8017bb444a462fb3d38ad5070429e94df7cf6b91d82697f36d0fa9"
 )
 
 
@@ -69,9 +68,12 @@ def _generate_mariadb_tls_certs(namespace_name: str) -> tuple[str, str, str]:
     """
     ca_key = rsa.generate_private_key(public_exponent=65537, key_size=2048)
 
-    ca_subject = ca_issuer = x509.Name([x509.NameAttribute(oid=NameOID.COMMON_NAME, value=f"mariadb-ca-{namespace_name}")])
+    ca_subject = ca_issuer = x509.Name(
+        attributes=[x509.NameAttribute(oid=NameOID.COMMON_NAME, value=f"mariadb-ca-{namespace_name}")]
+    )
     ca_cert = (
-        x509.CertificateBuilder()
+        x509
+        .CertificateBuilder()
         .subject_name(ca_subject)
         .issuer_name(ca_issuer)
         .public_key(ca_key.public_key())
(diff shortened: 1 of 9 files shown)


tests/ai_safety/trustyai_service/utils.py
```diff
@@ -63,6 +56,249 @@ def _get_mariadb_pods() -> list[Pod]:
...
+def _generate_mariadb_tls_certs(namespace_name: str) -> tuple[str, str, str]:
+    """Generate self-signed TLS certificates for MariaDB.
+
+    Returns:
+        tuple: (ca_cert_pem, server_cert_pem, server_key_pem)
+    """
+    from cryptography import x509
+    from cryptography.hazmat.primitives import hashes, serialization
+    from cryptography.hazmat.primitives.asymmetric import rsa
+    from cryptography.x509.oid import NameOID
```
dbasunag: Should go to the top of the file.

tests/ai_safety/trustyai_service/utils.py
```diff
@@ -24,30 +25,22 @@
...
-        "mariadb-operator-helm-controller-manager",
-        "mariadb-operator-webhook",
-    ]
-
-    for name in expected_deployment_names:
-        deployment = Deployment(client=client, name=name, namespace=mariadb_operator.namespace)
-        deployment.wait_for_replicas()
+MARIADB_IMAGE = (
+    "registry.redhat.io/rhel9/mariadb-1011@sha256:092407d87f8017bb444a462fb3d38ad5070429e94df7cf6b91d82697f36d0fa9"
+)
```
dbasunag: if you don't need this specific image, I would recommend using https://github.com/opendatahub-io/opendatahub-tests/blob/52d32422f5cd58696436e8f6320e3017dd4eccda/tests/ai_hub/utils.py#L39. It is multi arch, and already is part of disconnected mirroring.

tests/ai_safety/trustyai_service/utils.py
```diff
@@ -63,6 +56,249 @@ def _get_mariadb_pods() -> list[Pod]:
...
+                            {
+                                "name": "MYSQL_DATABASE",
+                                "valueFrom": {
+                                    "secretKeyRef": {"name": db_credentials_secret_name, "key": "databaseName"}
+                                },
+                            },
+                            {
+                                "name": "MARIADB_ROOT_PASSWORD",
+                                "valueFrom": {
+                                    "secretKeyRef": {"name": db_credentials_secret_name, "key": "databasePassword"}
```
dbasunag, line 215: We have very similar code. May be consider reusing these https://github.com/opendatahub-io/opendatahub-tests/blob/main/tests/ai_hub/utils.py

kpunwatk (approved):

tests/ai_safety/trustyai_service/utils.py
sheltoncyril: Claude
 ```
Easy to move: just the image constant. MARIA_DB_IMAGE has zero dependencies — pure string. Drop it in utilities/constants.py (or a new utilities/database.py), import from
  both places. Trivial.

  Hard to move: the deployment helpers. The functions (get_database_env_vars, get_model_registry_deployment_template_dict, etc.) are deeply coupled to model-registry
  conventions:
  - Secret keys differ: ai_hub uses database-user/database-password, TrustyAI uses databaseUsername/databasePassword
  - Template includes model-registry labels (sidecar.istio.io/inject, app.kubernetes.io/*)
  - Constants like DB_BASE_RESOURCES_NAME = "db-model-registry", MODEL_REGISTRY_DB_SECRET_STR_DATA are model-registry-specific
  - Handles 3 backends (mysql/mariadb/postgres) — TrustyAI only needs mariadb with TLS
  
  To truly share the deployment code, you'd need to parameterize secret key names, label schemes, and TLS setup — basically rewriting the functions into a generic DB
  deployment builder. That's a bigger refactor than this PR warrants.

  Recommendation: Move just MARIA_DB_IMAGE to global (addresses reviewer comment #2 directly), keep TrustyAI's deployment template as-is (explain in PR reply that secret
  schemas differ so full reuse needs a follow-up).
```

dbasunag (approved):

PR #1706, opened 2026-06-09
fix: Make Llama Stack agents test assertions resilient to LLM model changes

Artemon-line:
## Summary
- Replace brittle single-keyword assertions in `test_agents_simple_agent` with multi-keyword matching (same pattern used in `test_responses.py`)
- Remove the `@pytest.mark.skip` — tests are no longer disabled
- Improved error messages include the question and actual response for easier debugging

## Root cause
The agents tests checked for a single specific word in LLM responses (`"model"` for "Who are you?", `"answer"` for "What can you do?"). When the serving model changed, the LLM phrased responses differently, causing assertion failures.

## Fix
Use `any(keyword in text.lower() for keyword in expected_keywords)` with a broader set of acceptable keywords, matching the approach already used in `test_responses.py`.

## Test plan
- [ ] Verify agents tests pass in Jenkins with current LLM model
- [ ] Confirm no regressions in other Llama Stack tests


jgarciao:
I think it would be better to fix this test and, once it is confirmed that the Llama Stack tests don't need to be run on RHOAI 2.25 EUS, disable them in Jenkins setting minRoai = 3.0 in the jenkins repo at configs/components-testing/components/llama_stack/main.yaml?ref_type=heads#L7-8

tests/llama_stack/agents/test_agents.py
```diff
@@ -1,13 +1,16 @@
...
 from llama_stack_client import Agent, LlamaStackClient, RAGDocument
 from simple_logger.logger import get_logger
+
 from tests.llama_stack.constants import ModelInfo
 from tests.llama_stack.utils import get_torchtune_test_expectations, validate_rag_agent_responses
 
 LOGGER = get_logger(name=__name__)
 
 
+@pytest.mark.skip(reason="Llama Stack is deprecated in RHOAI 2.25 EUS (RHAIENG-5523)")
```
dbasunag: Why not remove the tests from the branch?

dbasunag (dismissed):

Artemon-line:
@jgarciao Thanks for the suggestion! I've updated the approach — instead of skipping the tests, I fixed the brittle assertions that were failing due to the LLM model change. The tests now use multi-keyword matching (same pattern as `test_responses.py`) so they're resilient to different model phrasings.

Once we confirm Llama Stack doesn't need to run on RHOAI 2.25 EUS, we can set `minRhoai = 3.0` in Jenkins as you suggested.

tests/llama_stack/agents/test_agents.py
Artemon-line: Changed the approach — instead of removing/skipping, I fixed the root cause. The assertions were brittle (checking for a single keyword like `"model"` or `"answer"`). Now using multi-keyword matching with a broader set of acceptable keywords, same pattern as `test_responses.py`.

dbasunag (approved):

jgarciao (approved):

jgarciao:
LGTM. Thanks @Artemon-line !

PR #1708, opened 2026-06-09
test(workbenches): Add post-upgrade notebook creation test for notebook controller

jstourac:

## Summary

- Adds `test_upgrade_creation.py` with 6 post-upgrade-only tests that create a fresh Notebook CR on the upgraded platform and verify full controller + webhook functionality (pod Ready, StatefulSet, Service, HTTPRoute, kube-rbac-proxy sidecar injection, reconciliation lock clearance).
- Adds supporting session-scoped fixtures (`new_notebook`, `new_notebook_pvc`, `new_notebook_pod`, `new_notebook_statefulset`, `new_notebook_service`, `new_notebook_httproute`) to `conftest.py` with proper cleanup (`teardown=True`).

## Test plan

- [x] Execute on an upgraded cluster: confirm the new notebook pod reaches Ready, all owned resources are created, auth sidecar is injected, and the reconciliation lock annotation is cleared

- [x] Locally
- [ ] Jenkins

## Additional Requirements

## Summary by CodeRabbit

* **Tests**
  * Enhanced post-upgrade notebook creation test coverage, validating pod readiness, resource initialization, and authentication features.
  * Improved test fixture exception handling for increased reliability.
  * Refined test structure to consolidate pod validation in fixtures.


tests/workbenches/notebooks_server/controller/upgrade/conftest.py
```diff
@@ -519,3 +520,116 @@ def upgrade_notebook_baseline(
...
+) -> Pod:
+    """Pod for the post-upgrade new notebook; waits for Ready state."""
+    notebook_pod = Pod(
+        client=unprivileged_client,
+        namespace=new_notebook.namespace,
+        name=f"{new_notebook.name}-0",
+    )
+
+    try:
+        notebook_pod.wait()
```
dbasunag, line 575: Do you want to wait for the pod to be created or for the pod to be in running state? using wait_for_status() would check that the pod exists first.

tests/workbenches/notebooks_server/controller/upgrade/conftest.py
```diff
@@ -519,3 +520,116 @@ def upgrade_notebook_baseline(
...
+        notebook_pod.wait_for_condition(
+            condition=Pod.Condition.READY,
+            status=Pod.Condition.Status.TRUE,
+            timeout=Timeout.TIMEOUT_5MIN,
+        )
+    except (TimeoutError, TimeoutExpiredError, RuntimeError) as e:
```
dbasunag: I don't see a possibility of RuntimeError here, however as per https://github.com/RedHatQE/openshift-python-wrapper/blob/main/ocp_resources/resource.py#L1310C13-L1310C27 ConditionError is a possibility

tests/workbenches/notebooks_server/controller/upgrade/test_upgrade_creation.py
```diff
@@ -0,0 +1,107 @@
...
+    @pytest.mark.post_upgrade
+    def test_new_notebook_pod_ready(
+        self,
+        new_notebook_pod: Pod,
+    ) -> None:
+        """Given the platform was upgraded,
+        When a new Notebook CR is created,
+        Then the notebook pod should reach Ready state.
+        """
+        assert new_notebook_pod.exists, f"Pod '{new_notebook_pod.name}' was not created on upgraded platform"
```
dbasunag: if you reach here then the pod must exist, this test should probably just log something and pass.

tests/workbenches/notebooks_server/controller/upgrade/test_upgrade_creation.py
```diff
@@ -0,0 +1,107 @@
...
+    def test_new_notebook_has_auth_sidecar(
+        self,
+        new_notebook_pod: Pod,
+    ) -> None:
+        """Given a new notebook is created post-upgrade with inject-auth,
+        When the ODH webhook mutates the CR,
+        Then the pod should have a kube-rbac-proxy sidecar container.
+        """
+        containers = new_notebook_pod.instance.spec.containers
+        container_names = [c.name for c in containers]
```
dbasunag: Please use words for variable names :)

tests/workbenches/notebooks_server/controller/upgrade/test_upgrade_creation.py
```diff
@@ -0,0 +1,107 @@
...
+    @pytest.mark.post_upgrade
+    def test_new_notebook_httproute_exists(
+        self,
+        new_notebook_httproute: HTTPRoute,
+    ) -> None:
+        """Given a new notebook is created post-upgrade,
+        When the ODH controller reconciles routing,
+        Then an HTTPRoute should be created.
+        """
+        assert new_notebook_httproute.exists, (
```
harshad16, line 86: Checking on for existence of httproute, feel like it is inconsistent with the other route upgrade, right ?
For post creation, we should also check for parentref and backref as we did here:
https://github.com/opendatahub-io/opendatahub-tests/blob/991f487efbf1c095c0f3c02d8dde1fcd93a6cc4a/tests/workbenches/notebooks_server/controller/upgrade/test_upgrade_routing.py#L16
or we maybe we can reuse this function ?

tests/workbenches/notebooks_server/controller/upgrade/test_upgrade_creation.py
```diff
@@ -0,0 +1,107 @@
...
+            f"HTTPRoute '{new_notebook_httproute.name}' was not created on upgraded platform"
+        )
+
+    @pytest.mark.post_upgrade
+    def test_new_notebook_has_auth_sidecar(
```
harshad16, line 102: This good, though Sidecar check covers webhook injection, but `test_upgrade_auth.py` also verifies controller auth reconciliation for inject-auth notebooks:
- Service `{name}-kube-rbac-proxy` with port 8443
- ConfigMap `{name}-kube-rbac-proxy-config`
- ClusterRoleBinding `{name}-rbac-{ns}-auth-delegator` → `system:auth-delegator`

should we include them as well ?

harshad16 (changes requested):
Thanks for including this check, this cover the new nb creation very well.
Just have few suggestion/question, take a look at the comment.

tests/workbenches/notebooks_server/controller/upgrade/conftest.py
jstourac: I would like to be sure that the pod is not only running, but it's fully up and as such the condition of itself is set to ready. That is the reason why wait and wait_for_condition calls are used here. I presume that we can replace the wait with the wait_for_status, but since we would keep the separate line, I am not sure what would be the benefit here? :thinking:

tests/workbenches/notebooks_server/controller/upgrade/conftest.py
jstourac: You are correct! I removed the `RuntimeError`. The `ConditionError` is a possibility, but only if we would use the `stop_condition`, which we don't, so I didn't add it there. I also updated similar cases in other places of our code for our component. Thanks for this!

tests/workbenches/notebooks_server/controller/upgrade/test_upgrade_creation.py
jstourac: Correct, very good catch. I removed the assertion, though - I kept the test as a simple gating test for a clearer log in case the fixture itself fails - it would fail in a test named accordingly instead of in a test that checks a statefulset. I don't feel strongly about my change, so if you would prefer otherwise, let me know.

I updated also two other places where this was relevant too.

tests/workbenches/notebooks_server/controller/upgrade/test_upgrade_creation.py
jstourac: Ah, correct. Apologies - I have more commits in queue and pulling them step by step for smaller review and I haven't applied your comment for these new commits. I'll try not to forget this in the future. Should be fixed now. Thanks!

tests/workbenches/notebooks_server/controller/upgrade/test_upgrade_creation.py
jstourac: Very good observation. I was thinking that maybe checking an existence may be enough here as this is a brand new workbench creation and if there is a bug in this, we would already catch it with our regular non-upgrade tests. But you are right - there could be anything that went wrong during the upgrade itself. Let's perform a full check here as we do in the routing test.

tests/workbenches/notebooks_server/controller/upgrade/test_upgrade_creation.py
jstourac: Good call, this is a bit similar to the previous one as I thought to be just brief with our checks, but you are correct - let's have the decent checks here too as anything could break during the migration/upgrade!

jstourac:
@dbasunag , @harshad16 thank you both for your reviews. I addressed your comments via [this compare](https://github.com/opendatahub-io/opendatahub-tests/compare/c92761e38e5209e1c1ddaa1e3d942331164edccb..9739e4dc33c2eb546fa6e2ab17f5da92900ebaa3). The second compare is just a rebase against the latest `main` branch. Please, feel free to re-reivew. Thank you!

jstourac:
@coderabbitai full review

jiridanek (approved):

tests/workbenches/notebooks_server/controller/upgrade/test_upgrade_creation.py
harshad16: thanks for addressing this issue.

tests/workbenches/notebooks_server/controller/upgrade/test_upgrade_creation.py
harshad16: thanks for addressing this 💯

harshad16 (approved):
Thanks for the changes 💯

PR #1731, opened 2026-06-11
test(workbenches): Add CA bundle ConfigMap preservation checks for notebook controller upgrade

jstourac:

## Summary

- Adds pre- and post-upgrade tests verifying the `workbench-trusted-ca-bundle` ConfigMap persists across an upgrade with its `ca-bundle.crt` key intact and its `resourceVersion` unchanged.
- Introduces a `workbench_trusted_ca_bundle` fixture in `conftest.py` and stores the ConfigMap's `resourceVersion` in the upgrade baseline.
- Moves shared constants (`WORKBENCH_TRUSTED_CA_BUNDLE_NAME`, `CA_BUNDLE_CERT_KEY`) to `utils.py` for reuse across test and fixture modules.

## Test plan

- [x] Run `--pre-upgrade`: confirm `test_ca_bundle_configmap_exists_before_upgrade` passes (ConfigMap exists with `ca-bundle.crt`)
- [x] Run `--post-upgrade`: confirm `test_ca_bundle_configmap_exists_after_upgrade` and `test_ca_bundle_configmap_not_modified_after_upgrade` pass
- [x] Verify the immutability test gracefully skips if the CA bundle was absent pre-upgrade

- [x] Locally
- [ ] Jenkins

## Additional Requirements

## Summary by CodeRabbit

* **Tests**
  * Enhanced notebook upgrade tests to capture and persist trusted CA bundle state as part of pre-upgrade baselines.
  * Added pre-upgrade and post-upgrade assertions that trusted CA bundle data exists.
  * Added consistency checks to ensure workbench bundle changes only when the source bundle changes.

### DIFF
--- a/tests/workbenches/notebooks_server/controller/upgrade/conftest.py
+++ b/tests/workbenches/notebooks_server/controller/upgrade/conftest.py
@@ -17,6 +17,7 @@ from pytest_testconfig import config as py_config
 from timeout_sampler import TimeoutExpiredError
 
 from tests.workbenches.notebooks_server.controller.utils import (
+    WORKBENCH_TRUSTED_CA_BUNDLE_NAME,
     StatefulSet,
     build_notebook_dict,
     resolve_notebook_image,
@@ -35,6 +36,7 @@ UPGRADE_NOTEBOOK_NAME = "upgrade-workbenches"
 UPGRADE_STOPPED_NOTEBOOK_NAME = "upgrade-wb-stopped"
 NEW_NOTEBOOK_NAME = "upgrade-wb-new"
 UPGRADE_BASELINE_CM_NAME = "upgrade-workbenches-baseline"
+ODH_TRUSTED_CA_BUNDLE_NAME = "odh-trusted-ca-bundle"
 
 
 @pytest.fixture(scope="session")
@@ -437,6 +439,31 @@ def stopped_notebook_pre_upgrade_shutdown(
 
 
 @pytest.fixture(scope="session")
+def workbench_trusted_ca_bundle(
+    unprivileged_client: DynamicClient,
+    upgrade_notebook_namespace: Namespace,
+) -> ConfigMap:
+    """The workbench-trusted-ca-bundle ConfigMap created by the ODH controller."""
+    return ConfigMap(
+        client=unprivileged_client,
+        name=WORKBENCH_TRUSTED_CA_BUNDLE_NAME,
+        namespace=upgrade_notebook_namespace.name,
+    )
+
+
+@pytest.fixture(scope="session")
+def odh_trusted_ca_bundle(
+    admin_client: DynamicClient,
+) -> ConfigMap:
+    """The odh-trusted-ca-bundle ConfigMap in the applications namespace (source of trust)."""
+    return ConfigMap(
+        client=admin_client,
+        name=ODH_TRUSTED_CA_BUNDLE_NAME,
+        namespace=py_config["applications_namespace"],
+    )
+
+
+@pytest.fixture(scope="session")
 def capture_notebook_baseline(
     pytestconfig: pytest.Config,
     admin_client: DynamicClient,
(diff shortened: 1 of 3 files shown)


jiridanek (dismissed):

tests/workbenches/notebooks_server/controller/upgrade/test_upgrade.py
```diff
@@ -158,3 +178,43 @@ def test_service_not_modified_after_upgrade(
...
+    ) -> None:
+        """Given the workbench-trusted-ca-bundle existed before upgrade,
+        When the upgrade completes,
+        Then its resourceVersion should be unchanged.
+        """
+        saved_resource_version = upgrade_notebook_baseline.get("ca_bundle_resource_version")
+        if not saved_resource_version:
+            pytest.skip("CA bundle baseline not captured (ConfigMap may not have existed pre-upgrade)")
+
+        current_resource_version = workbench_trusted_ca_bundle.instance.metadata.resourceVersion
```
harshad16: I like the approach, however i would recommend , that we should check the resourceVersion of the top level certs in `odh-trusted-ca-bundle` , if this changes, we would expect change in workbench_trusted_ca_bundle, so it wont be a wrong change.

harshad16 (dismissed):
Thank for the check, 
i believe from a upgrade test point of view, what we wanna check is, if the workbench trusted ca bundle changes after upgrade, it should still not trigger a workbench to reconcile , can we cover this scenario if we havent dont this in other past PR.

validating if bundle has changed, has no direct impact on the customer use-case, unless we can storngly say certs doesnt change, which is not true.
as customer can add new certs, maybe they do it during upgrades.
wdyt?

tests/workbenches/notebooks_server/controller/upgrade/conftest.py
```diff
@@ -472,6 +487,10 @@ def capture_notebook_baseline(
 
     stopped_annotation = stopped_notebook.instance.metadata.annotations.get("kubeflow-resource-stopped")
 
+    ca_bundle_resource_version = ""
+    if workbench_trusted_ca_bundle.exists:
+        ca_bundle_resource_version = workbench_trusted_ca_bundle.instance.metadata.resourceVersion
+
```
dbasunag: Do you expect it to be present always? Right now it would be silently proceeding to next step if the configmap does not exist, is that the intention? Why not fail early?

tests/workbenches/notebooks_server/controller/upgrade/test_upgrade.py
jstourac: This is actually very good proposal - i implemented it - so basically we now just check that either both resources were updated or none of them.

tests/workbenches/notebooks_server/controller/upgrade/conftest.py
jstourac: Good catch, I updated the code so that now we basically expect the bundle is always present.

jstourac:
@harshad16 

> Thank for the check, i believe from a upgrade test point of view, what we wanna check is, if the workbench trusted ca bundle changes after upgrade, it should still not trigger a workbench to reconcile , can we cover this scenario if we havent dont this in other past PR.

Good point, I will try to raise it as a separate PR as I think this can be a standard test, not an upgrade test.
 
> validating if bundle has changed, has no direct impact on the customer use-case, unless we can storngly say certs doesnt change, which is not true. as customer can add new certs, maybe they do it during upgrades. wdyt?

Actually, I wasn't fully sure myself whether this is a good check, but changing the test per your suggestion (odh bundle changed implies workbnech bundle changed and vice versa) is a proper approach here.

jstourac:
@harshad16 @dbasunag thank you both for your reviews, I addressed your comments and replied here.

Changes can be seen via [this](https://github.com/opendatahub-io/opendatahub-tests/compare/a01c0471cb3eb6bdadd2b7d93eade97ffa7056ca..119bdd60ff2c90ed98a0883429db4e8df4344079) and [this followup](https://github.com/opendatahub-io/opendatahub-tests/compare/119bdd60ff2c90ed98a0883429db4e8df4344079..07a2b348279685a33848f11b519da7f1698b397d).

harshad16 (commented):
thanks for the change, this looks good to me.

jiridanek (approved):

PR #1767, opened 2026-06-16
fix(trustyai): remove double slash and trailing slash in metric endpoint URLs

SudipSinha:
## Summary
- Fix double leading slash (`//metrics/drift/meanshift/`) in one-shot metric request URLs caused by prepending `/` to a base URL that already starts with `/`
- Fix trailing slash from empty ternary branch that triggered 307 redirects on some servers, dropping POST bodies
- Add `rstrip('/')` guard in `_get_metric_base_url` to prevent future double-slash issues

## Test plan
- [ ] Run `pytest tests/ai_safety/trustyai_service/drift/test_drift.py::TestDriftMetrics` against both Java and Python TrustyAI service images
- [ ] Verify no 307 redirects or 404s from malformed URLs in test logs

## Release Notes

* **Bug Fixes**
  * Fixed malformed metric URLs in TrustyAI service that could occur during metric requests. URL path construction now properly handles base URLs and metric names to ensure correct endpoint resolution.

### DIFF
--- a/tests/ai_safety/trustyai_service/trustyai_service_utils.py
+++ b/tests/ai_safety/trustyai_service/trustyai_service_utils.py
@@ -82,7 +82,7 @@ class TrustyAIServiceClient:
             base_url = "/metrics/drift"
         else:
             raise MetricValidationError(f"Unknown metric: {metric_name}")
-        return f"{base_url}/{metric_name}"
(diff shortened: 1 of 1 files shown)


sheltoncyril (approved):
/lgtm

kpunwatk (approved):

PR #1773, opened 2026-06-17
Fix Python 2 except syntax in evalhub CRD check

tareqalayan:
Replace Python 2 syntax `except AttributeError, KeyError:` with Python 3 syntax `except (AttributeError, KeyError):` in the _is_evalhub_crd_available() helper function.

The old syntax only catches AttributeError and assigns the exception to a variable named KeyError, which masks real KeyError exceptions. This could cause tests to run against missing CRDs or silently skip when they shouldn't.

Fixes: RHOAIENG-68255

# Pull Request

## Related Issues

- Fixes: 
- JIRA: 

- [ ] Locally
- [ ] Jenkins

## Additional Requirements

### DIFF
--- a/tests/ai_safety/evalhub/kueue/conftest.py
+++ b/tests/ai_safety/evalhub/kueue/conftest.py
@@ -71,7 +71,10 @@ def _is_evalhub_crd_available(admin_client: DynamicClient) -> bool:
             name=crd_name,
         )
         return crd.exists
-    except AttributeError, KeyError:
+    except (
+        AttributeError,
+        KeyError,
+    ):
         return False
 
 
--- a/tests/ai_safety/evalhub/conftest.py
+++ b/tests/ai_safety/evalhub/conftest.py
@@ -61,7 +61,10 @@ def _is_evalhub_crd_available(admin_client: DynamicClient) -> bool:
             name=crd_name,
         )
         return crd.exists
-    except AttributeError, KeyError:
+    except (
+        AttributeError,
+        KeyError,
+    ):
         return False
 
 


sheltoncyril (dismissed):
/lgtm

kpunwatk (approved):

tests/ai_safety/evalhub/kueue/conftest.py
```diff
@@ -71,7 +71,10 @@ def _is_evalhub_crd_available(admin_client: DynamicClient) -> bool:
             name=crd_name,
         )
         return crd.exists
-    except AttributeError, KeyError:
+    except (
```
dbasunag, line 74: We are on python 3.14. Current syntax is legit and supported

dbasunag (changes requested):

tests/ai_safety/evalhub/kueue/conftest.py
dbasunag: https://peps.python.org/pep-0758/.

tareqalayan: Thanks @dbasunag  ,

then, in case we run only in 3.14+ i think this change is redunduant !

cc @sheltoncyril   ?

sheltoncyril: Yep, if it's on 3.14+ we don't need it, I wonder if older branches use <3.14 tho

dbasunag: For branches with 3.13 (2.25 and 3.3) we should not have this syntax. The changes were introduced as a part of the update if I remember correctly.

PR #1781, opened 2026-06-17
fix(trustyai): 3.5ea2 add post-upgrade metric re-registration check and fix DB upgrade test reliability

ssaleem-rh:
# Pull Request

## Summary

 - Added verify_trustyai_service_metric_registered_post_upgrade() helper and two new post-upgrade tests (for both PVC and DB storage paths) to verify that metrics scheduled before upgrade are
  automatically re-registered by TrustyAI after upgrade, with clear failure messages pointing to known 3.5 breaking changes
  - Fixed double slash in TrustyAIServiceClient request URLs by stripping leading slashes from endpoint paths (.lstrip('/')), and fixed the ISVC logger sink URL scheme check to derive http/https from the
  deployment mode annotation rather than hardcoding https
  - Added db_resources_pre_existing marker and fixture to scope DB resource reuse to test classes that explicitly declare it, and surfaced the full response body in metric delete assertion errors for
  easier debugging
 
 
## Related Issues

- Fixes: 
- JIRA: (https://redhat.atlassian.net/browse/RHOAIENG-68448)

Tested Independently without upgrade 3.5ea2 pre -> 3.5ea2 post
AND
Tested with upgrade from 2.25 -> 3.5ea2

Pre (2.25):
image

Post (3.5):
image

- [x] Locally
- [ ] Jenkins

## Additional Requirements

## Summary by CodeRabbit

* **Tests**
  * Added post-upgrade assertions ensuring pre-existing MEANSHIFT drift metrics are re-registered after TrustyAI service upgrades (both PVC-to-DB and DB upgrade paths).
  * Improved post-upgrade coverage by adding upgrade-scoped verification helpers and gating with dedicated readiness dependencies.
  * Enhanced failure diagnostics by including delete response bodies when delete requests fail and by providing clearer, upgrade-specific assertion messages for known verification expectations.
  * Updated test request URL construction to handle endpoint paths more consistently.


pytest.ini
```diff
@@ -16,6 +16,7 @@ markers =
     slow: Mark tests which take more than 10 minutes as slow tests.
     pre_upgrade: Mark tests which should be run before upgrading the product.
     post_upgrade: Mark tests which should be run after upgrading the product.
+    db_resources_pre_existing: Mark test classes whose DB resources (MariaDB, secrets) pre-exist from a prior run.
```
dbasunag: Why would you leave resources behind, outside of pre-upgrade?

ssaleem-rh: The PVC upgrade test starts with no DB resources, so they are created fresh during post-upgrade

The DB upgrade test already has MariaDB/secrets pre-upgrade, so we use the existing resources post-upgrade. I thought a marker would be cleaner than an if/else on namespace name comparison.

dbasunag: I would use if/else here not a marker. Use marker if you need to utilize the hooks.

dbasunag (changes requested):

dbasunag (dismissed):

tests/ai_safety/trustyai_service/conftest.py
```diff
@@ -182,7 +187,7 @@ def db_credentials_secret(
     In pre-upgrade mode (or when no upgrade flag is set), creates a new secret with MariaDB
     connection details and manages cleanup via teardown_resources.
     """
-    if pytestconfig.option.post_upgrade:
+    if pytestconfig.option.post_upgrade and _is_db_upgrade_namespace(model_namespace.name):
```
dbasunag: @ssaleem-rh please use key=value in function calls.

dbasunag (approved):

PR #1824, opened 2026-06-20
chore: Add pre-commit hooks for test anti-patterns and pattern documentation

IKRedHat:
## Summary

- Add a pre-commit hook script that catches common test anti-patterns before they
  enter the codebase
- Add a `PATTERNS.md` reference document for model runtime test conventions

## Motivation

**CI failures observed:** The `"deployment_type"` dict key mismatch and dead-code
`CHECK_SNAPSHOT` toggle persisted across dozens of files undetected through code review,
causing `ValueError: Invalid deployment_mode Standard` in production CI
([launch 13453, items 1066450-1066453](https://reportportal-opendatascience.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#opendatascience/launches/all/13453)).
The pre-commit hook catches both anti-patterns at commit time before they can re-enter
the codebase.

During analysis of the model runtime test suite, several recurring anti-patterns
were identified that led to test failures, maintenance burden, or misleading results:

1. **`"deployment_type"` vs `"deployment_mode"` dict key mismatch** — conftest fixtures
   read `deployment_mode` but test params used `deployment_type`, causing silent KeyError
   or wrong defaults
2. **`CHECK_SNAPSHOT` env-var toggles** — snapshot validation was gated behind an env var
   that was never set in CI, making snapshot files dead code
3. **Hardcoded `RawDeployment` strings** — using string literals instead of the
   `KServeDeploymentType` enum leads to drift when upstream renames values
4. **Missing `__init__.py`** in test packages — causes pytest collection failures

The pre-commit hook flags these patterns at commit time with actionable error messages,
preventing regressions.

### New files (2)
- `scripts/check_model_runtime_patterns.sh` — shell script checking for anti-patterns
  in `tests/model_serving/model_runtime/`. Runs only on files staged for commit.
  Checks: `deployment_type` dict keys, `CHECK_SNAPSHOT` usage, hardcoded deployment
  mode strings, missing `__init__.py` files
- `tests/model_serving/model_runtime/PATTERNS.md` — documents the team's conventions
  for conftest structure, parametrize patterns, deployment mode usage, and validation
  approach

### Modified files (1)
- `.pre-commit-config.yaml` — adds `check-model-runtime-patterns` hook (local repo,
  runs on Python files under `tests/model_serving/model_runtime/`)

## Test plan

- [ ] Stage a file with `"deployment_type"` key → hook blocks commit with descriptive message
- [ ] Stage a file with `CHECK_SNAPSHOT` → hook blocks commit
- [ ] Stage a clean file → hook passes
- [ ] `pre-commit run check-model-runtime-patterns --all-files` — passes on current codebase

## Summary by CodeRabbit

* **Tests**
  * Added automated pre-commit validation to enforce model runtime test pattern compliance, catching common anti-patterns during development.

* **Documentation**
  * Introduced comprehensive pattern guidelines for model runtime tests, covering validation approaches, fixture configuration, deployment conventions, and test structure best practices.

* **Chores**
  * Updated code review configuration to include targeted pattern compliance checks for model serving tests.


scripts/check_model_runtime_patterns.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+# TC-GUARD-001: Pre-commit anti-pattern checker for model runtime tests.
```
dbasunag, line 2: pre-commit runs on every single pr including model run time. These seems like good instructions for code review, but adding these would increase pre-commit execution and it does not seem necessary for any other components.

dbasunag (changes requested):
Please add guideline for code rabbit instead. May be something like this would work?
```
 # .coderabbit.yaml
  reviews:
    path_instructions:
      - path: "tests/model_serving/model_runtime/**"
        instructions: |
          Review against the patterns documented in tests/model_serving/model_runtime/PATTERNS.md.
          Flag violations of anti-patterns listed there as errors, specifically:
          - CHECK_SNAPSHOT usage (use fuzzy validation instead)
          - gRPC protocol references in Triton tests (not supported on RHOAI)
          - Hardcoded deployment mode strings (use KServeDeploymentType enum)
          - Exact snapshot comparisons in vLLM tests (use fuzzy validation for non-deterministic output)
```
along with 
```
code_guidelines:
    file_patterns:
      - "**/PATTERNS.md"
```
may work better.

scripts/check_model_runtime_patterns.sh
IKRedHat: Good point — a pre-commit hook running on every PR across the whole repo is too broad for model_runtime-specific patterns. We could scope it with a path filter (only trigger when `tests/model_serving/model_runtime/` files are changed), or move these to a targeted CI job instead. If neither approach works for the repo's workflow, happy to close this PR and keep them as code review guidelines. What's your preference?

dbasunag: I think telling coderabbit to catch the issues https://github.com/opendatahub-io/opendatahub-tests/pull/1824#pullrequestreview-4548869299 might work well.

IKRedHat: Good idea — we can actually do both. The pre-commit hook is already scoped with `files: ^tests/model_serving/model_runtime/` so it only runs when model_runtime files are changed, no impact on other PRs. This gives us a reliable safety net that doesn't depend on CodeRabbit availability (e.g. when it hits rate limits). I'll add the `.coderabbit.yaml` path_instructions on top so reviewers get the pattern guidance during code review as well.

IKRedHat:
@dbasunag Added CodeRabbit instructions per your suggestion in 60d1114:

- **`path_instructions`** for `tests/model_serving/model_runtime/**` — flags anti-patterns (snapshot assertions, gRPC in Triton, hardcoded deployment modes) during review
- **`code_guidelines`** — registered `PATTERNS.md` as a guidelines source under `knowledge_base` so CodeRabbit references it automatically

So now we have both layers: the pre-commit hook (scoped to model_runtime files only) as a local safety net, and CodeRabbit for review-time guidance.

scripts/check_model_runtime_patterns.sh
dbasunag: Pre-commit I feel is an overkill. But this is your team's call and we can always revisit at a later point. So not holding the PR.

dbasunag (approved):

scripts/check_model_runtime_patterns.sh
```diff
@@ -0,0 +1,42 @@
...
+set -euo pipefail
+
+ERRORS=0
+TARGET_DIR="tests/model_serving/model_runtime"
+
+if [ ! -d "$TARGET_DIR" ]; then
+    exit 0
+fi
+
+# Anti-pattern 1: CHECK_SNAPSHOT environment variable (removed, should not return)
```
sheltoncyril, line 14: An AI-assisted review flagged this and I wanted to check before treating it as actionable.

The `CHECK_SNAPSHOT` (AP1) and `TRITON_GRPC` (AP2) patterns seem to match pre-existing legitimate definitions already in the tree (`utils.py`, `triton/constant.py`), and since the hook runs `pass_filenames: false` over the whole dir it looks like it would fail on the current codebase. Also, the `deployment_type` vs `deployment_mode` key mismatch the PR targets doesn't appear to be detected. Did `pre-commit run --all-files` pass for you locally? Thanks!

PR #441, opened 2025-07-15
Clean up mariadb pvcs

dbasunag:
This pvcs are created with mariadb kind, however, they are being left behind.
## Merge criteria:

- [ ] The commits are squashed in a cohesive manner and have meaningful messages.
- [ ] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [ ] The developer has manually tested the changes and verified that the changes work

PR #1832, opened 2026-06-22
fix: default AWS credentials to dummy value to prevent OGX core crash

Artemon-line:
## Summary
- Default `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` to `"dummy"` instead of `""` in `tests/ogx/constants.py`

## Context
Empty string defaults cause `botocore.exceptions.NoCredentialsError` during S3 provider initialization, crashing the OGX core pod even when S3 is not used. After [ogx-distribution PR #359](https://github.com/opendatahub-io/ogx-distribution/pull/359) added Bedrock STS auth config, the distro `config.yaml` S3 provider picks up `AWS_ACCESS_KEY_ID` via `${env.AWS_ACCESS_KEY_ID:=}`. When the test secret contains empty strings, botocore treats them as "credentials provided but invalid" rather than "no credentials", and crashes at startup — blocking **all** OGX tests, not just S3-dependent ones.

This is a safety net for local runs — CI uses vault-injected values.

## Test plan
- [x] Verified locally: changing default from `""` to `"dummy"` resolves the `NoCredentialsError` crash
- [x] OGX core pod starts successfully with dummy creds when S3 is not enabled
- [ ] CI: S3 tests use real AWS creds from vault (unaffected by this change)

## Summary by CodeRabbit

* **Tests**
  * Updated default values for test environment variable configuration to improve test reliability and consistency.


dbasunag (dismissed):

dbasunag (dismissed):

tests/fixtures/vector_io.py
```diff
@@ -473,6 +473,18 @@ def get_pgvector_deployment_template() -> dict[str, Any]:
...
+                            "exec": {
+                                "command": [
+                                    "/bin/sh",
+                                    "-c",
+                                    "sleep 5 && psql -h localhost -U $POSTGRES_USER"
+                                    ' -d $POSTGRES_DB -c "CREATE EXTENSION IF NOT EXISTS vector;" || true',
+                                ]
+                            }
+                        }
+                    },
```
jgarciao: If the test is failing due to  product bug https://github.com/ogx-ai/ogx/issues/6164 I think the right approach will be to create a jira about it and add it as known issue at the release notes (with the workaround of manually enabling the extension) and mark the test with @pytest.mark.xfail

jgarciao (changes requested):

tests/fixtures/vector_io.py
Artemon-line: Thanks for the review Jorge!

I'd argue this isn't a workaround — it's restoring proper test infrastructure setup that was removed prematurely in PR #1308 (Mar 26). The commit message there said *"the pgvector image now auto-enables the vector extension"*, but we verified today that neither the pgvector container image (`quay.io/jgarciao/pgvector`) nor `docker.io/pgvector/pgvector` auto-create the extension — both have an empty `/docker-entrypoint-initdb.d/`.

The upstream bug ([ogx-ai/ogx#6164](https://github.com/ogx-ai/ogx/issues/6164)) is real and we already filed a fix PR ([ogx-ai/ogx#6168](https://github.com/ogx-ai/ogx/pull/6168)), but even after that lands, the test infra should still pre-create the extension — same pattern as [ogx-distribution PR #441](https://github.com/opendatahub-io/ogx-distribution/pull/441) which does `CREATE EXTENSION IF NOT EXISTS vector` in the CI postgres setup. We don't want tests depending on the product's auto-setup working correctly.

Also note: this version uses `$POSTGRES_USER` / `$POSTGRES_DB` env vars (already injected via secretKeyRef) instead of the hardcoded password that motivated the original removal.

Jira blocker is already filed: [RHAIENG-5840](https://redhat.atlassian.net/browse/RHAIENG-5840).

jgarciao: >I'd argue this isn't a workaround — it's restoring proper test infrastructure setup that was removed prematurely in PR https://github.com/opendatahub-io/opendatahub-tests/pull/1308 (Mar 26). The commit message there said "the pgvector image now auto-enables the vector extension", but we verified today that neither the pgvector container image (quay.io/jgarciao/pgvector) nor docker.io/pgvector/pgvector auto-create the extension — both have an empty /docker-entrypoint-initdb.d/.

The pgvector extension is enabled automatically by the ogx server since this [PR](https://github.com/ogx-ai/ogx/pull/4660) was merged . That's why we removed it from the test infrastructure setup in https://github.com/opendatahub-io/opendatahub-tests/pull/1308, because it was no longer necessary and it was printing the passwords in the logs

>We don't want tests depending on the product's auto-setup working correctly.

In my opinion is it good that the tests verify that the product auto-setup work as expected

Artemon-line: Reverted the postStart hook — you're right, the pgvector extension issue is a product bug and should be tracked separately.

The AWS creds fix (constants.py) is still needed and is the only change remaining in this PR. The empty string defaults have been resolved by adjusting the CI env vars, but the code default should still be `"dummy"` instead of `""` as a safety net for local runs and environments without real AWS creds.

Upstream bug tracked in [ogx-ai/ogx#6164](https://github.com/ogx-ai/ogx/issues/6164), fix PR: [ogx-ai/ogx#6168](https://github.com/ogx-ai/ogx/pull/6168), Jira: [RHAIENG-5840](https://redhat.atlassian.net/browse/RHAIENG-5840).

jgarciao (dismissed):

dbasunag (approved):

jgarciao (approved):

PR #1845, opened 2026-06-23
fix: remove wait_for_condition from minimal_subscription_for_free_user

liangwen12year:
## Summary
Remove `wait_for_condition(condition="Ready", status="True", timeout=300)` from `minimal_subscription_for_free_user` fixture.

## Problem
PR #1757 added a wait for Ready to this fixture, but the fixture intentionally creates a MaaSModelRef pointing to a non-existent LLMInferenceService. The subscription correctly stays in Degraded/Failed (model ref is invalid) — this is expected MaaS behavior. The wait times out after 300s, causing `test_api_key_can_list_models` to fail in setup.

## Fix
Remove the wait. This fixture only needs the subscription to exist for auth testing, not to be Active. The original issue (#1757 was fixing) was about subscriptions not being waited for — but this specific fixture is a minimal dummy that can't become Ready by design.

## Test plan
- [ ] `test_api_key_can_list_models` no longer times out in fixture setup
- [ ] Other MaaS smoke tests unaffected

## Release Notes

This pull request contains internal test infrastructure improvements with **no user-visible changes**. 

* **Tests**
  * Simplified test fixture setup and resource management behavior.

### DIFF
--- a/tests/model_serving/maas_billing/conftest.py
+++ b/tests/model_serving/maas_billing/conftest.py
@@ -1252,5 +1252,4 @@ def minimal_subscription_for_free_user(
             wait_for_resource=True,
         ) as subscription,
     ):
-        subscription.wait_for_condition(condition="Ready", status="True", timeout=300)
         yield subscription


SB159:
@liangwen12year - Thanks for opening #1845. That fix made sense for the old minimal_subscription_for_free_user fixture, which used a dummy model ref to a non-existent LLMInferenceService, so the subscription would never become Ready and the wait would always time out. I’ve already fixed this on fix/minimal-subscription-for-free-user: 
The fixture now reuses the real TinyLlama model ref (maas_model_tinyllama_free)
Wekeepthe Ready wait, since the subscription can actually become Ready now test_api_key_can_list_modelsno longer uses this fixture (it usesmaas_subscription_tinyllama_freeinstead), So I think we can close #1845 — removing the wait would bring back flakiness for the other API key tests that still depend on this fixture.

PR #1863, opened 2026-06-25
test: add BBR path-wins auth and streaming inference tests

SB159:
# Pull Request

## Summary

add BBR path-wins auth and streaming inference tests

## Related Issues

- Fixes: 
- JIRA: 

- [x] Locally
- [ ] Jenkins

## Additional Requirements

## Summary by CodeRabbit

* **Tests**
  * Added coverage for inference requests that fail authentication, including invalid or missing API keys.
  * Added a regression check to ensure route-based behavior is preserved even when the request body references a different model.
  * Added streaming coverage to verify inference responses return server-sent events with streamed data chunks.
  * Updated test setup to use module-level fixture application for cleaner coverage setup.


tests/model_serving/maas_billing/body_base_routing/pre_auth_model_header/test_bbr_pre_auth_inference.py
```diff
@@ -63,6 +63,66 @@ def test_inference_model_in_body_only_invalid_key_returns_401(
             expected_status=401,
         )
 
+    @pytest.mark.tier1
+    @pytest.mark.parametrize("ocp_token_for_actor", [{"type": "free"}], indirect=True)
```
dbasunag: This is a class fixture, being used in function context. This does not seem right.

tests/model_serving/maas_billing/body_base_routing/pre_auth_model_header/test_bbr_pre_auth_inference.py
```diff
@@ -63,6 +63,66 @@ def test_inference_model_in_body_only_invalid_key_returns_401(
...
+            timeout=60,
+        )
+        assert response.status_code not in (401, 403), (
+            f"Auth rejected based on body model; expected path identity to win on /llm/ paths, "
+            f"got {response.status_code}"
+        )
+        LOGGER.info(f"Path-wins auth verified: got {response.status_code} (not 401/403)")
+
+    @pytest.mark.tier1
+    @pytest.mark.parametrize("ocp_token_for_actor", [{"type": "free"}], indirect=True)
```
dbasunag: Same.

tests/model_serving/maas_billing/body_base_routing/pre_auth_model_header/test_bbr_pre_auth_inference.py
SB159: Thanks, missed, Updated.

tests/model_serving/maas_billing/body_base_routing/pre_auth_model_header/test_bbr_pre_auth_inference.py
SB159: Thanks, missed, Updated.

tests/model_serving/maas_billing/body_base_routing/pre_auth_model_header/test_bbr_pre_auth_inference.py
```diff
@@ -63,6 +63,66 @@ def test_inference_model_in_body_only_invalid_key_returns_401(
...
+            json=streaming_payload,
+            timeout=60,
+            stream=True,
+        )
+        assert response.status_code == 200, f"Expected 200 for streaming BBR inference, got {response.status_code}"
+        content_type = response.headers.get("content-type", "")
+        assert "text/event-stream" in content_type, (
+            f"Expected text/event-stream content type for streaming response, got '{content_type}'"
+        )
+        chunks = [line for line, _ in zip(response.iter_lines(), range(10)) if line]
```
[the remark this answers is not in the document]
SB159: Addressed

tests/model_serving/maas_billing/body_base_routing/pre_auth_model_header/test_bbr_pre_auth_inference.py
```diff
@@ -63,6 +63,66 @@ def test_inference_model_in_body_only_invalid_key_returns_401(
...
+        response = request_session_http.post(
+            url=bbr_inference_url,
+            headers=bbr_api_key_headers,
+            json=wrong_model_payload,
+            timeout=60,
+        )
+        assert response.status_code not in (401, 403), (
+            f"Auth rejected based on body model; expected path identity to win on /llm/ paths, "
+            f"got {response.status_code}"
+        )
```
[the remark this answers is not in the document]
SB159: Addressed

tests/model_serving/maas_billing/body_base_routing/pre_auth_model_header/test_bbr_pre_auth_inference.py
dbasunag: This is not addressed.

tests/model_serving/maas_billing/body_base_routing/pre_auth_model_header/test_bbr_pre_auth_inference.py
dbasunag: Same.

tests/model_serving/maas_billing/body_base_routing/pre_auth_model_header/conftest.py
```diff
@@ -55,7 +55,7 @@ def bbr_inference_url(
     return f"{maas_scheme}://{maas_host}/llm/{maas_inference_service_tinyllama_free.name}/v1/chat/completions"
 
 
-@pytest.fixture(scope="class")
```
dbasunag, line 58: Changing class scoped fixture would have unwanted side effect.

SB159: Addressed in latest commit: class-level parametrize + class-scoped fixtures restored.

kpunwatk (approved):

tests/model_serving/maas_billing/body_base_routing/pre_auth_model_header/test_bbr_pre_auth_inference.py
```diff
@@ -14,38 +14,23 @@
 
 LOGGER = structlog.get_logger(name=__name__)
 
+pytestmark = [
+    pytest.mark.usefixtures(
+        "maas_unprivileged_model_namespace",
+        "maas_subscription_controller_enabled_latest",
+        "maas_gateway_api",
+        "maas_api_gateway_reachable",
+        "maas_free_group",
```
sheltoncyril, line 23: **High confidence** — Moving `usefixtures` from class-level to module-level `pytestmark` applies these 10 fixtures to every test and class in the file. If someone adds a new test class or standalone function later, they silently inherit potentially expensive cluster setup fixtures. Consider keeping the decorator on each class explicitly for a narrower blast radius.

tests/model_serving/maas_billing/body_base_routing/pre_auth_model_header/test_bbr_pre_auth_inference.py
```diff
@@ -14,38 +14,23 @@
...
+        "maas_model_tinyllama_free",
+        "maas_auth_policy_tinyllama_free",
+        "maas_subscription_tinyllama_free",
+        "maas_inference_service_tinyllama_free",
+    ),
```
sheltoncyril, line 28: **High confidence** — The old class had `@pytest.mark.parametrize("ocp_token_for_actor", [{"type": "free"}], indirect=True)` on each test method. After the split, `TestBBRPreAuthInferenceAuth` no longer has it — only `TestBBRPreAuthInference` does. If the `ocp_token_for_actor` fixture requires the indirect parametrize to be present, the auth tests (invalid key, no key) may error or pass for the wrong reason. Could you verify this is intentional?

tests/model_serving/maas_billing/body_base_routing/pre_auth_model_header/test_bbr_pre_auth_inference.py
```diff
@@ -89,8 +74,84 @@ def test_inference_no_api_key_returns_401_or_403(
...
+        bbr_api_key_headers: dict[str, str],
+    ) -> None:
+        """Verify that inference with stream=True returns 200 with SSE text/event-stream chunks."""
+        streaming_payload = {**bbr_chat_payload, "stream": True}
+        with request_session_http.post(
+            url=bbr_inference_url,
+            headers=bbr_api_key_headers,
+            json=streaming_payload,
+            timeout=60,
+            stream=True,
```
sheltoncyril, line 141: **High confidence** — The `chunks` list can be empty if the server returns blank keep-alive lines before actual data (all filtered out by `if line`). In that case `any([])` returns `False` and the assertion fails with no diagnostic context. Consider adding `assert chunks, 'No SSE lines received from streaming response'` before the `data:` prefix check.

sheltoncyril (commented):
## Review Summary (AI-assisted, multi-persona)

_This review was generated by an AI agent using three reviewer personas (chill, grumpy, unhinged) with confidence scoring. Findings flagged by multiple personas are higher confidence._

**High confidence (3):** streaming test empty-chunks edge case, module-level pytestmark blast radius, auth class missing ocp_token_for_actor parametrize.
**Medium confidence (4):** path-wins 404 assertion fragility, SSE data: prefix check too loose, status code assert doesn't gate chunk parsing, chunks variable scope after with-block.

tests/model_serving/maas_billing/body_base_routing/pre_auth_model_header/test_bbr_pre_auth_inference.py
```diff
@@ -89,8 +74,84 @@ def test_inference_no_api_key_returns_401_or_403(
...
+
+    @pytest.mark.tier1
+    def test_inference_path_wins_over_body_model_for_auth(
+        self: Self,
+        request_session_http: requests.Session,
+        bbr_inference_url: str,
+        bbr_api_key_headers: dict[str, str],
+    ) -> None:
+        """Verify that on /llm/ paths, auth uses path identity — wrong body model returns 404, not 401/403.
+
```
sheltoncyril, line 107: **Medium confidence** — The test asserts `expected_status=404` for a wrong body model, but depending on server implementation this could also be 400 or 422. The test name says "path wins for auth" but a 404 proves routing behavior, not auth behavior. Consider clarifying the test name and adding a comment explaining why 404 specifically is the expected outcome in this setup.

tests/model_serving/maas_billing/body_base_routing/pre_auth_model_header/test_bbr_pre_auth_inference.py
```diff
@@ -89,8 +74,84 @@ def test_inference_no_api_key_returns_401_or_403(
...
+        """Verify that inference with stream=True returns 200 with SSE text/event-stream chunks."""
+        streaming_payload = {**bbr_chat_payload, "stream": True}
+        with request_session_http.post(
+            url=bbr_inference_url,
+            headers=bbr_api_key_headers,
+            json=streaming_payload,
+            timeout=60,
+            stream=True,
+        ) as response:
+            assert response.status_code == 200, f"Expected 200 for streaming BBR inference, got {response.status_code}"
```
sheltoncyril, line 143: **Medium confidence** — `chunk.startswith(b"data:")` matches the SSE prefix but doesn't verify the payload structure. A malformed response like `data:garbage` (no space after colon) would also pass. Consider checking for `b"data: "` (with space per SSE spec), and optionally parsing at least one chunk as JSON to confirm the response contains valid inference data.

tests/model_serving/maas_billing/body_base_routing/pre_auth_model_header/test_bbr_pre_auth_inference.py
```diff
@@ -89,8 +74,84 @@ def test_inference_no_api_key_returns_401_or_403(
...
+        LOGGER.info("Path-wins auth verified: auth passed and routing returned 404 for wrong body model")
+
+    @pytest.mark.tier1
+    def test_inference_streaming_returns_sse(
+        self: Self,
+        request_session_http: requests.Session,
+        bbr_inference_url: str,
+        bbr_chat_payload: dict[str, Any],
+        bbr_api_key_headers: dict[str, str],
+    ) -> None:
```
sheltoncyril, line 133: **Medium confidence** — The status code assertion is inside the `with` block, but if it fails, chunk parsing still proceeds on the error response body. A 500 error containing `data: internal error` could pass the SSE assertion. Consider returning early or skipping chunk parsing when the status code check fails.

tests/model_serving/maas_billing/body_base_routing/pre_auth_model_header/test_bbr_pre_auth_inference.py
```diff
@@ -89,8 +74,84 @@ def test_inference_no_api_key_returns_401_or_403(
...
+            json=streaming_payload,
+            timeout=60,
+            stream=True,
+        ) as response:
+            assert response.status_code == 200, f"Expected 200 for streaming BBR inference, got {response.status_code}"
+            content_type = response.headers.get("content-type", "")
+            assert "text/event-stream" in content_type, (
+                f"Expected text/event-stream content type for streaming response, got '{content_type}'"
+            )
+            chunks = [line for line, _ in zip(response.iter_lines(), range(10), strict=False) if line]
```
sheltoncyril, line 148: **Medium confidence** — The `LOGGER.info` line after the `with` block references `chunks`, but if an assertion inside the `with` block raises before the list comprehension completes, `chunks` will be unbound, causing a secondary `NameError` that masks the real failure. Consider moving this log line inside the `with` block.

tests/model_serving/maas_billing/body_base_routing/pre_auth_model_header/test_bbr_pre_auth_inference.py
sheltoncyril: **Resolved — false positive.** After checking the fixture chain locally: TestBBRPreAuthInferenceAuth tests (invalid key, no key) don't use bbr_api_key_headers or ocp_token_for_actor — they hardcode invalid/empty headers. The parametrize removal is intentional and correct for these negative auth paths.

tests/model_serving/maas_billing/body_base_routing/pre_auth_model_header/test_bbr_pre_auth_inference.py
sheltoncyril: **Partially resolved.** The docstring (lines 106-110) explains why 404 is expected: BBR pre-auth sets X-Gateway-Model-Name from the body, so post-auth routing returns 404 for a mismatched model. Original comment missed this context. The naming suggestion still has minor merit but is not blocking.

tests/model_serving/maas_billing/body_base_routing/pre_auth_model_header/test_bbr_pre_auth_inference.py
sheltoncyril: **Resolved — false positive.** The assert on line 143 raises AssertionError if status != 200, which stops execution before chunk parsing. Pytest assertions are exceptions — they do gate subsequent code. A 500 response would not reach the SSE assertions.

tests/model_serving/maas_billing/body_base_routing/pre_auth_model_header/test_bbr_pre_auth_inference.py
dbasunag: They would be executing once per module, vs once per class. Since all the Maas fixtures this seems like a right move to me. I feel this is a false positive.

dbasunag (approved):

tests/model_serving/maas_billing/body_base_routing/pre_auth_model_header/test_bbr_pre_auth_inference.py
SB159: Yes, intentional. Auth tests don’t need ocp_token_for_actor — they only assert 401/403 with invalid/missing credentials. Class-level parametrize is on TestBBRPreAuthInference only.

tests/model_serving/maas_billing/body_base_routing/pre_auth_model_header/test_bbr_pre_auth_inference.py
SB159: Good point — edge case with blank keep-alive lines. Not seen in our runs; happy to add a clearer assert in a follow-up if needed.

tests/model_serving/maas_billing/body_base_routing/pre_auth_model_header/test_bbr_pre_auth_inference.py
SB159: Docstring covers why 404 is expected (auth passes, routing fails for wrong body model). Haven’t seen 400/422 for this case; happy to adjust naming in a follow-up if needed.

tests/model_serving/maas_billing/body_base_routing/pre_auth_model_header/test_bbr_pre_auth_inference.py
SB159: Intentionally minimal for smoke — just verify we get SSE data chunks. Happy to use data: or parse JSON if we need stricter validation later.

tests/model_serving/maas_billing/body_base_routing/pre_auth_model_header/test_bbr_pre_auth_inference.py
SB159: With the current assert on status code, chunk parsing only runs after 200. Haven’t seen a false pass on error bodies

tests/model_serving/maas_billing/body_base_routing/pre_auth_model_header/test_bbr_pre_auth_inference.py
SB159: Assert failures stop the test before that log line, so haven’t seen this in practice.

sheltoncyril (approved):
/lgtm

PR #1869, opened 2026-06-26
test: base pr for mcp server preview

dbasunag:
# Pull Request

## Related Issues

- Fixes: 
- JIRA: 

- [ ] Locally
- [ ] Jenkins

## Additional Requirements

## Summary by CodeRabbit

* **New Features**
  * Added automated MCP preview test coverage for YAML-path mode, including included/excluded glob filtering and `filterStatus` (all/included/excluded) summary validation.
  * Introduced shared test helpers for building preview requests, validating preview counts, and checking item-level `included` flags.

* **Refactor**
  * Centralized include/exclude pattern matching and authenticated multipart POST behavior into shared helpers; reused a common catalog-container constant across tests.

* **Bug Fixes**
  * Strengthened assertions to catch mismatched totals or incorrect per-item inclusion results earlier.


tests/ai_hub/mcp_servers/preview/conftest.py
```diff
@@ -0,0 +1,82 @@
...
+from kubernetes.dynamic import DynamicClient
+from ocp_resources.route import Route
+
+from tests.ai_hub.mcp_servers.preview.utils import build_mcp_preview_config
+from tests.ai_hub.utils import execute_authenticated_post, get_model_catalog_pod
+from utilities.infra import get_openshift_token
+
+LOGGER = structlog.get_logger(name=__name__)
+
+CATALOG_CONTAINER: str = "catalog"
```
fege: this is already present in https://github.com/opendatahub-io/opendatahub-tests/blob/51c40d0de0c636496d3db92847275d6869abd2fa/tests/ai_hub/model_catalog/constants.py#L21 . Probably a good moment to move up and import in this file

tests/ai_hub/mcp_servers/preview/test_mcp_preview.py
```diff
@@ -0,0 +1,128 @@
...
+            excluded_patterns=excluded_patterns,
+            expected_server_names=catalog_server_names,
+        )
+
+    @pytest.mark.parametrize(
+        "mcp_preview_result",
+        [pytest.param({}, id="test_no_filters")],
+        indirect=True,
+    )
+    def test_mcp_preview_no_filters(
```
fege: why not a parameter of the previous test?

fege (dismissed):
/lgtm

sheltoncyril (commented):
## Review Summary (AI-assisted, multi-persona)

_This review was generated by an AI agent using three reviewer personas (chill, grumpy, unhinged) with confidence scoring. Findings flagged by multiple personas are higher confidence._

- **High confidence (3/3 agreement):** 3 findings (2 errors, 1 warning)
- **Medium confidence (2/3 agreement):** 9 findings (5 warnings, 1 nit)
- **Low confidence (1/3):** 4 findings (omitted from inline comments)

Main concerns: cross-component import violating isolation rules, and `filter_status` dual-source parametrize that could silently test the wrong thing.

tests/ai_hub/mcp_servers/preview/conftest.py
```diff
@@ -0,0 +1,82 @@
+from typing import Any
+
+import pytest
+import structlog
+import yaml
+from kubernetes.dynamic import DynamicClient
+from ocp_resources.route import Route
+
+from tests.ai_hub.mcp_servers.preview.utils import build_mcp_preview_config
+from tests.ai_hub.model_catalog.constants import CATALOG_CONTAINER
```
sheltoncyril: **[High confidence — 3/3 reviewers]**

`CATALOG_CONTAINER` is imported from `tests.ai_hub.model_catalog.constants`, which creates a cross-component dependency between `mcp_servers` and `model_catalog`. Project convention is to avoid imports between test components — shared code should go in `tests/ai_hub/constants.py` or `utilities/`.

Since the constant is just `"catalog"`, either promote it to `tests/ai_hub/constants.py` (where both components can import it) or define it locally. There's already a local definition in `tests/ai_hub/mcp_servers/config/test_invalid_yaml.py` as precedent.

tests/ai_hub/mcp_servers/preview/test_mcp_preview.py
```diff
@@ -0,0 +1,114 @@
...
+        )
+        validate_preview_items(
+            result=result,
+            included_patterns=included_patterns,
+            excluded_patterns=excluded_patterns,
+            expected_server_names=catalog_server_names,
+        )
+
+    @pytest.mark.parametrize(
+        "mcp_preview_result, filter_status",
```
sheltoncyril, line 60: **[High confidence — 3/3 reviewers]**

`filter_status` is passed in two places: as a key inside the indirect `mcp_preview_result` param dict (used by the fixture to build the API query) and as a separate direct parameter to the test function. If someone updates one parametrize value without updating the other, the test silently validates against the wrong filter status.

Consider having the fixture return the `filter_status` it used (or derive it from the result), so there's a single source of truth.

tests/ai_hub/mcp_servers/preview/conftest.py
```diff
@@ -0,0 +1,82 @@
...
+
+LOGGER = structlog.get_logger(name=__name__)
+
+MCP_CATALOG_FILE: str = "/shared-data/redhat-mcp-servers-catalog.yaml"
+
+MODEL_CATALOG_API_PATH: str = "/api/model_catalog/v1alpha1/"
+
+
+@pytest.fixture(scope="class")
+def model_catalog_preview_url(model_registry_namespace: str, admin_client: DynamicClient) -> str:
```
sheltoncyril, line 22: **[High confidence — 3/3 reviewers]**

Two observations:
1. The fixture name `model_catalog_preview_url` is a bit misleading in an MCP servers preview module.
2. The Route-lookup logic (find by label selector, extract host, build URL) is the same pattern used by existing fixtures like `model_catalog_rest_url` in `tests/ai_hub/model_catalog/conftest.py`. Consider reusing the existing Route discovery and just swapping the API path suffix, or extracting a shared helper.

tests/ai_hub/mcp_servers/preview/conftest.py
```diff
@@ -0,0 +1,82 @@
...
+from ocp_resources.route import Route
+
+from tests.ai_hub.mcp_servers.preview.utils import build_mcp_preview_config
+from tests.ai_hub.model_catalog.constants import CATALOG_CONTAINER
+from tests.ai_hub.utils import execute_authenticated_post, get_model_catalog_pod
+from utilities.infra import get_openshift_token
+
+LOGGER = structlog.get_logger(name=__name__)
+
+MCP_CATALOG_FILE: str = "/shared-data/redhat-mcp-servers-catalog.yaml"
```
sheltoncyril, line 16: **[Medium confidence — 2/3 reviewers]**

`MODEL_CATALOG_API_PATH` is already defined in `tests/ai_hub/test_security.py:11` with the same value. Consider promoting to `tests/ai_hub/constants.py` to avoid divergence.

tests/ai_hub/mcp_servers/preview/conftest.py
```diff
@@ -0,0 +1,82 @@
...
+    admin_client: DynamicClient,
+    model_registry_namespace: str,
+) -> dict[str, Any]:
+    """Fetch and parse MCP server catalog YAML from the catalog pod."""
+    model_catalog_pod = get_model_catalog_pod(client=admin_client, model_registry_namespace=model_registry_namespace)[0]
+    raw = model_catalog_pod.execute(command=["cat", MCP_CATALOG_FILE], container=CATALOG_CONTAINER)
+    return yaml.safe_load(raw)
+
+
+@pytest.fixture(scope="class")
```
sheltoncyril, line 48: **[Medium confidence — 2/3 reviewers]**

`get_model_catalog_pod(...)[0]` — if the function returns an empty list, this raises an opaque `IndexError`. The `model_catalog_preview_url` fixture above has a proper assert for empty results. Consider adding a similar guard here for clearer error messages.

tests/ai_hub/mcp_servers/preview/conftest.py
```diff
@@ -0,0 +1,82 @@
...
+    return servers
+
+
+@pytest.fixture()
+def mcp_preview_result(
+    request: pytest.FixtureRequest,
+    model_catalog_preview_url: str,
+    preview_user_token: str,
+) -> tuple[dict[str, Any], list[str] | None, list[str] | None]:
+    """Execute the MCP preview API and return (result, included_patterns, excluded_patterns).
```
sheltoncyril, line 62: **[Medium confidence — 2/3 reviewers]**

The fixture returns `(result, included_servers, excluded_servers)` but the test already has access to the input patterns via `request.param`. Round-tripping the inputs through the return value couples the fixture return type to caller knowledge. A `NamedTuple` would at least make the destructuring self-documenting, or the test could extract patterns from `request.param` directly.

tests/ai_hub/mcp_servers/preview/test_mcp_preview.py
```diff
@@ -0,0 +1,114 @@
...
+    )
+    def test_mcp_preview_filter_status(
+        self: Self,
+        mcp_preview_result: tuple[dict[str, Any], list[str] | None, list[str] | None],
+        filter_status: str,
+    ):
+        """
+        Test the MCP preview API with filterStatus parameter.
+
+        Given an MCP server catalog with include/exclude filters that produce both included and excluded servers,
```
sheltoncyril, line 88: **[Medium confidence — 2/3 reviewers]**

`validate_preview_items` is called here without `expected_server_names`, so the catalog-membership check is silently skipped. In `test_mcp_preview_server_filters` above, it is passed. If this omission is intentional (filter_status tests only care about included/excluded status), it might be worth a brief comment explaining why.

tests/ai_hub/model_catalog/metadata/utils.py
```diff
@@ -1,42 +1,27 @@
...
-from tests.ai_hub.utils import execute_get_command, get_rest_headers
+from tests.ai_hub.utils import (
+    execute_authenticated_post,
+    execute_get_command,
+    get_rest_headers,
+    should_include_by_pattern,
+)
 
 LOGGER = structlog.get_logger(name=__name__)
 CATALOG_CONTAINER = "catalog"
```
sheltoncyril, line 19: **[Medium confidence — 2/3 reviewers]**

`execute_model_catalog_post_command` is now a pass-through wrapper with an identical signature to `execute_authenticated_post`. There are 5 callers in `test_catalog_preview.py` that still use this wrapper. Consider migrating those callers to use the shared function directly to complete the refactor.

tests/ai_hub/mcp_servers/preview/utils.py
```diff
@@ -0,0 +1,101 @@
...
+        config_lines.append("excludedServers:")
+        config_lines.extend(f'  - "{pattern}"' for pattern in excluded_servers)
+
+    return "\n".join(config_lines)
+
+
+def validate_mcp_preview_counts(
+    result: dict[str, Any],
+    mcp_servers: list[dict[str, Any]],
+    included_patterns: list[str] | None = None,
```
sheltoncyril, line 38: **[Medium confidence — 2/3 reviewers]**

`validate_mcp_preview_counts` and `validate_preview_items` are structurally very similar to the existing model catalog equivalents in `metadata/utils.py`, with field names swapped (`Models` → `Assets`). Since the API response structure appears to be the same (same `summary`, `items`, `included` fields), consider whether a single parameterized validator set could serve both use cases.

tests/ai_hub/mcp_servers/preview/conftest.py
dbasunag: This is MCP catalog. Not applicable.

tests/ai_hub/mcp_servers/preview/conftest.py
dbasunag: Not the same and should stay separate.

tests/ai_hub/mcp_servers/preview/conftest.py
dbasunag: Not relevant since we have health check.

tests/ai_hub/mcp_servers/preview/conftest.py
dbasunag: Not making sense. Going to keep it simple.

tests/ai_hub/mcp_servers/preview/test_mcp_preview.py
dbasunag: This is a string param passed by the test. Current design should stay.

tests/ai_hub/mcp_servers/preview/test_mcp_preview.py
dbasunag: This is intentional. We don't want to hard code the server names here

fege (approved):
/lgtm

sheltoncyril (approved):
/lgtm

SB159 (approved):
/lgtm

PR #1888, opened 2026-06-30
feat(RHOAIENG-63118) - Enhance KServe upgrade tests with Kueue integration

sukumars321:
# Pull Request

## Summary
- Updated `pytest_runtest_setup` to conditionally enable KServe for upgrade tests based on item keywords.
- Introduced new utility functions for managing DSCI service mesh state during upgrades.
- Added configuration and test cases for KServe raw deployment with Kueue integration, including pre/post upgrade scenarios.
- Refactored existing utility functions to support the new upgrade tests and ensure proper state management.

This commit improves the testing framework for KServe upgrades, ensuring that Kueue integration is properly validated.

## Related Issues

- Fixes: 
- JIRA: https://redhat.atlassian.net/browse/RHOAIENG-63118
- This code change addresses Task 5 of [[RHOAIENG-63117] Kueue Upgrade Test Coverage Audit Report](https://docs.google.com/document/d/1ON66oE8MuKhHsT_YaG5hxTdAjhcHbYOvd5pskj9uPyg/edit?usp=sharing)

- [x] Locally
- [ ] Jenkins

## Additional Requirements

## Summary by CodeRabbit

* **New Features**
  * Added support for KServe raw-deployment upgrades with Kueue, including preserved service settings and post-upgrade inference checks.
  * Added configurable inference timeouts for prediction calls.

* **Bug Fixes**
  * Improved upgrade handling so KServe and Kueue settings are restored more reliably after tests.
  * Made component updates wait longer and more consistently for reconciliation.

* **Tests**
  * Expanded upgrade coverage for scaling, gating, restarts, and post-upgrade serving behavior.

[RHOAIENG-63117]: https://redhat.atlassian.net/browse/RHOAIENG-63117?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


conftest.py
```diff
@@ -425,7 +425,10 @@ def pytest_runtest_setup(item: Item) -> None:
             LOGGER.error(f"Database error: {db_exception}. Must-gather collection may not be accurate")
 
     if KServeDeploymentType.RAW_DEPLOYMENT.lower() in item.keywords:
-        item.fixturenames.insert(0, "enabled_kserve_in_dsc")
+        if "kueue" in item.keywords and ("pre_upgrade" in item.keywords or "post_upgrade" in item.keywords):
+            item.fixturenames.insert(0, "ensure_kserve_enabled_for_upgrade")
+        else:
+            item.fixturenames.insert(0, "enabled_kserve_in_dsc")
```
chambridge:
```python
if KServeDeploymentType.RAW_DEPLOYMENT.lower() in item.keywords:
    if "kueue" in item.keywords and ("pre_upgrade" in item.keywords or "post_upgrade" in item.keywords):
        item.fixturenames.insert(0, "ensure_kserve_enabled_for_upgrade")
    else:
        item.fixturenames.insert(0, "enabled_kserve_in_dsc")
```

This injects `ensure_kserve_enabled_for_upgrade` for **any** test with `rawdeployment` + `kueue` + `pre_upgrade`/`post_upgrade` markers, regardless of where in the repo it lives. If someone adds those markers to a test outside the upgrade directory (where the fixture is defined), pytest will fail with `fixture 'ensure_kserve_enabled_for_upgrade' not found`.

More importantly, `ensure_kserve_enabled_for_upgrade` modifies global cluster state (DSCI serviceMesh patching). Auto-injecting it based on markers alone risks affecting unrelated tests in the same session.

**Suggestion**: Either:
- Scope the condition to the upgrade path: add `and "upgrade" in str(item.fspath)` to the check
- Or (better) use `@pytest.mark.usefixtures("ensure_kserve_enabled_for_upgrade")` directly on the `TestKserveKueueRawPreUpgrade` / `TestKserveKueueRawPostUpgrade` classes, removing the root conftest change entirely

tests/model_serving/model_server/upgrade/conftest.py
```diff
@@ -1430,3 +1468,880 @@ def _capture_and_save_llmd_baseline(
...
+        pytestconfig=pytestconfig,
+        admin_client=admin_client,
+        dsc_resource=dsc_resource,
+        dsci_resource=dsci_resource,
+        kserve_kueue_upgrade_namespace=kserve_kueue_upgrade_namespace,
+        teardown_resources=teardown_resources,
+    )
+
+
+def _ensure_kueue_available_for_upgrade(
```
chambridge: **File**: `tests/model_serving/model_server/upgrade/conftest.py:1629` vs existing at `conftest.py:1239`

Both do the same thing: check operator installed → save managementState to ConfigMap → patch to Unmanaged → wait for CRDs → restore on teardown. The new version is parameterized by namespace and adds idempotency (reusing existing ConfigMaps). The existing one is not refactored to delegate.

tests/model_serving/model_server/upgrade/test_upgrade_kserve_kueue_raw.py
```diff
@@ -0,0 +1,316 @@
...
+        assert total_copies == KSERVE_KUEUE_EXPECTED_RUNNING_PODS, (
+            f"InferenceService should have {KSERVE_KUEUE_EXPECTED_RUNNING_PODS} total model copy, got {total_copies}"
+        )
+        LOGGER.info(
+            event=f"[PRE-UPGRADE] PASS: Kueue gating active — {running_pods} running, "
+            f"{gated_pods} gated, totalCopies={total_copies}"
+        )
+
+
+class TestKserveKueueRawPostUpgrade:
```
chambridge, line 205: **File**: `tests/model_serving/model_server/upgrade/test_upgrade_kserve_kueue_raw.py:189-314`

`TestKserveKueueRawPostUpgrade` verifies:
- ISVC exists (line 209)
- LocalQueue exists (line 223)
- Kueue integration stats unchanged (line 238)
- Total copies unchanged (line 261)
- Generation unchanged (line 279)
- Pods not restarted (line 295)

But it never sends an actual inference request. The existing `test_upgrade.py` tests verify inference both pre and post upgrade. Confirming that Kueue gating survived is not meaningful if the model no longer serves predictions.

**Suggestion**: Add a `test_inference_post_upgrade` test that sends a prediction request to the ISVC and validates the response.

tests/model_serving/model_server/upgrade/test_upgrade_kserve_kueue_raw.py
```diff
@@ -0,0 +1,316 @@
...
+        kserve_kueue_upgrade_inference_service: InferenceService,
+        baseline: dict,
+    ) -> None:
+        """Test steps:
+
+        1. Verify ISVC status.totalCopies matches the pre-upgrade baseline.
+        """
+        isvc = kserve_kueue_upgrade_inference_service
+        isvc.get()
+        total_copies = isvc.instance.status.modelStatus.copies.totalCopies
```
chambridge: **File**: `tests/model_serving/model_server/upgrade/test_upgrade_kserve_kueue_raw.py:271` and `utils.py:554`

```python
total_copies = isvc.instance.status.modelStatus.copies.totalCopies
```

No null checks on `modelStatus`, `copies`, or `totalCopies`. If the ISVC hasn't fully reconciled, this chain throws `AttributeError`. Either add a guard or wrap in a try/except with a clear message.

tests/model_serving/model_server/upgrade/conftest.py
```diff
@@ -1430,3 +1468,880 @@ def _capture_and_save_llmd_baseline(
...
+@pytest.fixture(scope="session")
+def kserve_kueue_upgrade_model_storage(pytestconfig: pytest.Config) -> dict[str, str]:
+    """Return model storage path and version for the configured S3 backend."""
+    if _s3_credentials_available(pytestconfig=pytestconfig):
+        return {
+            "storage_path": ModelStoragePath.OPENVINO_EXAMPLE_MODEL,
+            "model_version": ModelVersion.OPSET13,
+        }
+    return {
+        "storage_path": "serving/kserve/openvino-age-gender-recognition",
```
chambridge:
```python
"storage_path": "serving/kserve/openvino-age-gender-recognition"
```

This should be a constant in `kserve_kueue_upgrade_config.py` alongside the other `KSERVE_KUEUE_*` constants.

tests/model_serving/model_server/upgrade/test_upgrade_kserve_kueue_raw.py
```diff
@@ -0,0 +1,316 @@
...
+        )
+        deployment = deployments[0]
+
+        status_replicas = None
+        try:
+            for status_replicas in TimeoutSampler(
+                wait_timeout=120,
+                sleep=5,
+                func=lambda: _get_deployment_status_replicas(deployment),
+            ):
```
chambridge, line 158: **File**: `tests/model_serving/model_server/upgrade/test_upgrade_kserve_kueue_raw.py:136,159`

```python
TimeoutSampler(wait_timeout=120, sleep=5, ...)
```

Other tests use `Timeout.TIMEOUT_*` constants from `utilities.constants`. Use those for consistency.

chambridge (commented):
Come comments based on my review. Also looks like this branch may need a rebase.

tests/model_serving/model_server/upgrade/conftest.py
```diff
@@ -1430,3 +1468,880 @@ def _capture_and_save_llmd_baseline(
...
+def _restore_dsci_servicemesh_state(
+    admin_client: DynamicClient,
+    dsci_resource: DSCInitialization,
+    namespace: str,
+    dsci_servicemesh_state_cm_name: str,
+) -> None:
+    """Restore original DSCI serviceMesh state from the saved ConfigMap."""
+    state_cm = ConfigMap(client=admin_client, name=dsci_servicemesh_state_cm_name, namespace=namespace)
+    if not state_cm.exists:
+        return
```
dbasunag: Should this not fail?

tests/model_serving/model_server/upgrade/conftest.py
```diff
@@ -1430,3 +1468,880 @@ def _capture_and_save_llmd_baseline(
...
+        restore_dsci_servicemesh_from_upgrade_state(
+            dsci_resource=dsci_resource,
+            original_servicemesh=original_servicemesh,
+        )
+    state_cm.clean_up()
+
+
+def _read_secret_string(secret: Secret, key: str) -> str:
+    """Decode a string value from a Kubernetes Secret data entry."""
+    secret.get()
```
dbasunag: This is not a legit call.

tests/model_serving/model_server/upgrade/conftest.py
```diff
@@ -1430,3 +1468,880 @@ def _capture_and_save_llmd_baseline(
...
+    kserve_kueue_upgrade_namespace: Namespace,
+    teardown_resources: bool,
+) -> Generator[None, Any, Any]:
+    """Enable KServe for raw-deployment Kueue upgrade tests.
+
+    Pre-upgrade patches DSCI serviceMesh (only when absent) and DSC kserve for raw deployment.
+    The DSCI patch is reverted after post-upgrade tests. DSC kserve changes are not reverted so they survive
+    the upgrade under test.
+    """
+    dsc_resource.get()
```
dbasunag: ditto

tests/model_serving/model_server/upgrade/conftest.py
```diff
@@ -1430,3 +1468,880 @@ def _capture_and_save_llmd_baseline(
...
+    spec_kserve = dsc_resource.instance.spec.components.get(DscComponents.KSERVE)
+
+    if pytestconfig.option.post_upgrade:
+        if not _is_kserve_ready(dsc_resource=dsc_resource):
+            pytest.skip(
+                "KServe is not ready after upgrade. "
+                "Ensure pre-upgrade tests enabled KServe and the cluster upgrade completed successfully."
+            )
```
dbasunag: This should fail.

tests/model_serving/model_server/upgrade/conftest.py
```diff
@@ -1430,3 +1468,880 @@ def _capture_and_save_llmd_baseline(
...
+        yield
+        _restore_dsci_servicemesh_state(
+            admin_client=admin_client,
+            dsci_resource=dsci_resource,
+            namespace=kserve_kueue_upgrade_namespace.name,
+            dsci_servicemesh_state_cm_name=UPGRADE_DSCI_SERVICEMESH_STATE_CM_NAME,
+        )
+        return
```
dbasunag: No need to have a return. Instead please use if/else

tests/model_serving/model_server/upgrade/conftest.py
```diff
@@ -1430,3 +1468,880 @@ def _capture_and_save_llmd_baseline(
...
+
+    dsci_resource.get()
```
dbasunag: Please remove this line.

tests/model_serving/model_server/upgrade/conftest.py
```diff
@@ -1430,3 +1468,880 @@ def _capture_and_save_llmd_baseline(
...
+    servicemesh_patch = build_kserve_raw_deployment_dsci_servicemesh_patch(spec_service_mesh=spec_service_mesh)
+    if servicemesh_patch:
+        state_cm = ConfigMap(
+            client=admin_client,
+            name=UPGRADE_DSCI_SERVICEMESH_STATE_CM_NAME,
+            namespace=kserve_kueue_upgrade_namespace.name,
+            data=capture_dsci_servicemesh_state_for_upgrade(spec_service_mesh=spec_service_mesh),
+        )
+        if state_cm.exists:
+            LOGGER.info(f"Reusing existing DSCI serviceMesh state ConfigMap '{state_cm.name}'")
```
dbasunag: This should only happen in post upgrade? Please have clear expectation, that way we can fail properly.

tests/model_serving/model_server/upgrade/conftest.py
```diff
@@ -1430,3 +1468,880 @@ def _capture_and_save_llmd_baseline(
...
+        pre_patch_time = ready_condition.get("lastTransitionTime") if ready_condition else None
+        dsc_resource.update(
+            resource_dict={
+                "metadata": {"name": dsc_resource.name},
+                "spec": {"components": {DscComponents.KSERVE: kserve_patch}},
+            }
+        )
+        wait_for_dsc_reconciliation(dsc=dsc_resource, baseline_time=pre_patch_time)
+
+    dsc_resource.get()
```
dbasunag: Remove please.

tests/model_serving/model_server/upgrade/conftest.py
```diff
@@ -1430,3 +1468,880 @@ def _capture_and_save_llmd_baseline(
...
+    admin_client: DynamicClient,
+    dsc_resource: DataScienceCluster,
+    namespace: str,
+    teardown_resources: bool,
+) -> Generator[None, Any, Any]:
+    """Ensure Kueue is Unmanaged and ready; save/restore DSC state via ConfigMap in ``namespace``."""
+    kueue_dsc_state_cm_name = "upgrade-kueue-dsc-state"
+
+    if not pytestconfig.option.post_upgrade:
+        from tests.model_serving.model_server.conftest import _is_kueue_operator_installed
```
dbasunag: Please import at the top of this file.

tests/model_serving/model_server/upgrade/conftest.py
```diff
@@ -1430,3 +1468,880 @@ def _capture_and_save_llmd_baseline(
...
+
+        if not _is_kueue_operator_installed(admin_client):
+            pytest.skip("Kueue operator is not installed, skipping Kueue upgrade tests")
```
dbasunag: Should this not cause failure? Upgrade lanes should be running with Kueue installed.

tests/model_serving/model_server/upgrade/conftest.py
```diff
@@ -1430,3 +1468,880 @@ def _capture_and_save_llmd_baseline(
...
+            client=admin_client,
+            name=kueue_dsc_state_cm_name,
+            namespace=namespace,
+            data={"original_management_state": kueue_management_state},
+        )
+        if state_cm.exists:
+            LOGGER.info(f"Reusing existing Kueue DSC state ConfigMap '{kueue_dsc_state_cm_name}'")
+        else:
+            LOGGER.info(f"Saving original Kueue managementState '{kueue_management_state}' to ConfigMap")
+            state_cm.deploy()
```
dbasunag: Same comment. We should have clear expectation when something should exist vs not. Otherwise, we silently hide issues (infra or product).

tests/model_serving/model_server/upgrade/conftest.py
```diff
@@ -1430,3 +1468,880 @@ def _capture_and_save_llmd_baseline(
...
+                    "spec": {
+                        "components": {
+                            DscComponents.KUEUE: {"managementState": DscComponents.ManagementState.UNMANAGED}
+                        }
+                    },
+                }
+            )
+            wait_for_dsc_reconciliation(dsc=dsc_resource, baseline_time=pre_patch_time)
+        else:
+            LOGGER.info("Kueue already Unmanaged, no patch needed")
```
dbasunag: Same comment here.

tests/model_serving/model_server/upgrade/conftest.py
```diff
@@ -1430,3 +1468,880 @@ def _capture_and_save_llmd_baseline(
...
+        if local_queue.exists:
+            LOGGER.info(
+                f"Reusing existing Kueue resources for namespace '{namespace}' (LocalQueue '{local_queue_name}')"
+            )
+            yield local_queue
+            if teardown_resources:
+                local_queue.clean_up()
+                ClusterQueue(client=admin_client, name=cluster_queue_name).clean_up()
+                ResourceFlavor(client=admin_client, name=resource_flavor_name).clean_up()
+            return
```
dbasunag: Same comment here.

tests/model_serving/model_server/upgrade/conftest.py
```diff
@@ -1430,3 +1468,880 @@ def _capture_and_save_llmd_baseline(
...
+) -> Generator[Namespace, Any, Any]:
+    """Namespace for KServe raw ISVC + Kueue upgrade tests."""
+    ns = Namespace(client=admin_client, name=KSERVE_KUEUE_UPGRADE_NAMESPACE)
+
+    if pytestconfig.option.post_upgrade:
+        if not ns.exists:
+            pytest.skip(
+                f"[POST-UPGRADE] Namespace '{ns.name}' not found. "
+                "Ensure pre-upgrade KServe+Kueue tests completed successfully."
+            )
```
dbasunag: We should fail here.

tests/model_serving/model_server/upgrade/conftest.py
```diff
@@ -1430,3 +1468,880 @@ def _capture_and_save_llmd_baseline(
...
+        yield ns
+        if teardown_resources:
+            ns.clean_up()
+    else:
+        if ns.exists:
+            LOGGER.info(f"Reusing existing namespace '{ns.name}' from a prior pre-upgrade run")
+            yield ns
+            if teardown_resources:
+                ns.clean_up()
```
dbasunag: Same. We should fail.

tests/model_serving/model_server/upgrade/conftest.py
```diff
@@ -1430,3 +1468,880 @@ def _capture_and_save_llmd_baseline(
...
+@pytest.fixture(scope="session")
+def kserve_kueue_upgrade_minio_namespace(
+    pytestconfig: pytest.Config,
+    admin_client: DynamicClient,
+    teardown_resources: bool,
+) -> Generator[Namespace | None, Any, Any]:
+    """Dedicated namespace for MinIO without Kueue admission when external S3 is unavailable."""
+    if _s3_credentials_available(pytestconfig=pytestconfig):
+        yield None
+        return
```
dbasunag: Please use if/else, instead of using yield and return.

tests/model_serving/model_server/upgrade/conftest.py
```diff
@@ -1430,3 +1468,880 @@ def _capture_and_save_llmd_baseline(
...
+
+    ns = Namespace(client=admin_client, name=KSERVE_KUEUE_UPGRADE_MINIO_NAMESPACE)
+    if pytestconfig.option.post_upgrade:
+        if not ns.exists:
+            pytest.skip(
+                f"[POST-UPGRADE] MinIO namespace '{ns.name}' not found. "
+                "Ensure pre-upgrade tests completed with MinIO when external S3 is unavailable."
+            )
```
dbasunag: Please fail.

tests/model_serving/model_server/upgrade/conftest.py
```diff
@@ -1430,3 +1468,880 @@ def _capture_and_save_llmd_baseline(
...
+        if teardown_resources:
+            ns.clean_up()
+        return
+
+    if ns.exists:
+        LOGGER.info(f"Reusing existing MinIO namespace '{ns.name}'")
+        yield ns
+        if teardown_resources:
+            ns.clean_up()
+        return
```
dbasunag: Please fail.

tests/model_serving/model_server/upgrade/conftest.py
```diff
@@ -1430,3 +1468,880 @@ def _capture_and_save_llmd_baseline(
...
+def kserve_kueue_upgrade_minio_credentials_secret(
+    pytestconfig: pytest.Config,
+    admin_client: DynamicClient,
+    kserve_kueue_upgrade_minio_namespace: Namespace | None,
+    teardown_resources: bool,
+) -> Generator[Secret | None, Any, Any]:
+    """Store MinIO root credentials in a Secret when external S3 is unavailable."""
+    if kserve_kueue_upgrade_minio_namespace is None:
+        yield None
+        return
```
dbasunag: Please use if/else.

tests/model_serving/model_server/upgrade/conftest.py
```diff
@@ -1430,3 +1468,880 @@ def _capture_and_save_llmd_baseline(
...
+    if pytestconfig.option.post_upgrade:
+        if not secret.exists:
+            pytest.skip(
+                f"[POST-UPGRADE] MinIO credentials Secret '{secret.name}' not found in namespace "
+                f"'{secret.namespace}'. Ensure pre-upgrade tests completed successfully."
+            )
+        yield secret
+        if teardown_resources:
+            secret.clean_up()
+        return
```
dbasunag: Same comment here.

tests/model_serving/model_server/upgrade/conftest.py
```diff
@@ -1430,3 +1468,880 @@ def _capture_and_save_llmd_baseline(
...
+
+    if secret.exists:
+        LOGGER.info(f"Reusing existing MinIO credentials Secret '{secret.name}' in namespace '{secret.namespace}'")
+        yield secret
+        if teardown_resources:
+            secret.clean_up()
+        return
```
dbasunag: same comment

conftest.py
sukumars321: Addressed

tests/model_serving/model_server/upgrade/test_upgrade_kserve_kueue_raw.py
sukumars321: Addressed.

tests/model_serving/model_server/upgrade/conftest.py
sukumars321: removed the function itself, as part of removing minio dependency.

tests/model_serving/model_server/upgrade/conftest.py
sukumars321: done

tests/model_serving/model_server/upgrade/conftest.py
sukumars321: Addressed.

sukumars321:
@dbasunag @chambridge Addressed all the comments, and refactored the test to work with existing CI S3 env vars rather than deploying own minio deployment. PTAL.

tests/model_serving/model_server/upgrade/conftest.py
```diff
@@ -1430,3 +1402,608 @@ def _capture_and_save_llmd_baseline(
...
+                f"'{secret.namespace}'. Ensure pre-upgrade tests completed successfully."
+            )
+        yield secret
+        if teardown_resources:
+            secret.clean_up()
+    elif secret.exists:
+        LOGGER.info(f"Reusing existing S3 secret '{secret.name}' in namespace '{secret.namespace}'")
+        yield secret
+        if teardown_resources:
+            secret.clean_up()
```
dbasunag: This should fail. Because if we are not in post upgrade lane, the only situation that would lead to this would be due to bad previous run. Which needs to be triaged and possibly handled.

tests/model_serving/model_server/upgrade/conftest.py
```diff
@@ -1430,3 +1402,608 @@ def _capture_and_save_llmd_baseline(
...
+            )
+            model_runtime.delete()
+            model_runtime = ServingRuntime(**runtime_kwargs)
+            with ServingRuntimeFromTemplate(
+                **_kserve_kueue_upgrade_runtime_template_kwargs(
+                    runtime_kwargs=runtime_kwargs,
+                    teardown_resources=teardown_resources,
+                ),
+            ) as model_runtime:
+                yield model_runtime
```
dbasunag: Same comment.

chambridge (approved):
/lgtm
Thanks for addressing my feedback. 

Certainly @dbasunag is the expert here, so execution on her feedback is critical

dbasunag (dismissed):

sukumars321:
@mwaykole @threcc could you review this?

tests/model_serving/model_server/upgrade/conftest.py
```diff
@@ -1177,37 +1207,47 @@ def llmisvc_upgrade_auth_and_kueue(
...
-    dsc_resource,
+    dsc_resource: DataScienceCluster,
     llmisvc_auth_and_kueue_namespace: Namespace,
     teardown_resources: bool,
 ) -> Generator[None, Any, Any]:
     """Ensure Kueue is available for upgrade tests.
 
     Pre-upgrade:
-      1. Skip if kueue operator is not installed.
+      1. Fail if kueue operator is not installed.
```
mwaykole: why u are changing the core funcanality of fixture

mwaykole (changes requested):

tests/model_serving/model_server/upgrade/conftest.py
dbasunag: Can you please move this to the top of the file.

tests/model_serving/model_server/upgrade/conftest.py
dbasunag: Should this fail.

tests/model_serving/model_server/upgrade/conftest.py
```diff
@@ -1429,3 +1414,518 @@ def _capture_and_save_llmd_baseline(
...
+    namespace: str,
+    local_queue_name: str,
+    cluster_queue_name: str,
+    resource_flavor_name: str,
+    cpu_quota: int,
+    memory_quota: str,
+    teardown_resources: bool,
+) -> Generator[LocalQueue, Any, Any]:
+    """Create or look up Kueue resources for upgrade tests."""
+    from tests.model_serving.model_server.conftest import kueue_resource_groups
```
dbasunag: Please move this import to the top of the file.

tests/model_serving/model_server/upgrade/test_upgrade_kserve_kueue_raw.py
```diff
@@ -0,0 +1,344 @@
...
+                    break
+        except TimeoutExpiredError:
+            pytest.fail(
+                f"Timeout waiting for Kueue gating: expected "
+                f"{KSERVE_KUEUE_EXPECTED_RUNNING_PODS} running + "
+                f"{KSERVE_KUEUE_EXPECTED_GATED_PODS} gated, "
+                f"got {running_pods} running + {gated_pods} gated"
+            )
+
+        isvc.get()
```
dbasunag: Please remove this line.

tests/model_serving/model_server/upgrade/test_upgrade_kserve_kueue_raw.py
```diff
@@ -0,0 +1,344 @@
...
+        self,
+        kserve_kueue_upgrade_inference_service: InferenceService,
+        baseline: dict,
+    ) -> None:
+        """Test steps:
+
+        1. Verify ISVC status.totalCopies matches the pre-upgrade baseline.
+        """
+        isvc = kserve_kueue_upgrade_inference_service
+        isvc.get()
```
dbasunag: Please remove the get() call. It does not do anything.

tests/model_serving/model_server/upgrade/test_upgrade_kserve_kueue_raw.py
```diff
@@ -0,0 +1,344 @@
...
+            inference_config=OPENVINO_KSERVE_INFERENCE_CONFIG,
+            inference_type=Inference.INFER,
+            protocol=Protocols.HTTP,
+            use_default_query=True,
+            inference_timeout=KSERVE_KUEUE_INFERENCE_TIMEOUT,
+        )
+
+
+def _get_deployment_status_replicas(deployment: Deployment) -> int:
+    deployment.get()
```
dbasunag: Please remove the get call.

tests/model_serving/model_server/upgrade/test_upgrade_kserve_kueue_raw.py
```diff
@@ -0,0 +1,344 @@
...
+            inference_service=kserve_kueue_upgrade_inference_service,
+            inference_config=OPENVINO_KSERVE_INFERENCE_CONFIG,
+            inference_type=Inference.INFER,
+            protocol=Protocols.HTTP,
+            use_default_query=True,
+            inference_timeout=KSERVE_KUEUE_INFERENCE_TIMEOUT,
+        )
+
+
+def _get_deployment_status_replicas(deployment: Deployment) -> int:
```
dbasunag: Also, you don't need a utility just for the api call.

tests/model_serving/model_server/upgrade/utils.py
```diff
@@ -1,25 +1,257 @@
...
+    original_servicemesh: str,
+) -> None:
+    """Restore DSCI ``spec.serviceMesh`` to the value saved before pre-upgrade patching.
+
+    Args:
+        dsci_resource: DSCInitialization resource to update.
+        original_servicemesh: Serialized state from ``capture_dsci_servicemesh_state_for_upgrade``.
+    """
+    if original_servicemesh == DSCI_SERVICEMESH_STATE_ABSENT:
+        dsci_resource.get()
```
dbasunag: Please remove this call and any similar get() call.

tests/model_serving/model_server/upgrade/utils.py
```diff
@@ -1,25 +1,257 @@
...
+    state_cm = ConfigMap(client=admin_client, name=dsci_servicemesh_state_cm_name, namespace=namespace)
+    if not state_cm.exists:
+        pytest.fail(
+            f"DSCI serviceMesh state ConfigMap '{dsci_servicemesh_state_cm_name}' not found in namespace "
+            f"'{namespace}'. Ensure pre-upgrade tests saved the original state."
+        )
+
+    original_servicemesh = state_cm.instance.data.get(DSCI_SERVICEMESH_STATE_CM_KEY)
+    if original_servicemesh == DSCI_SERVICEMESH_STATE_NOT_PATCHED:
+        LOGGER.info("Pre-upgrade did not patch DSCI serviceMesh; skipping restore")
```
dbasunag: Should we not fail?

tests/model_serving/model_server/upgrade/utils.py
```diff
@@ -350,6 +601,60 @@ def capture_isvc_baseline(client: DynamicClient, isvc: InferenceService) -> ISVC
...
+        admin_client=client,
+    )
+    return {"running": running, "gated": gated}
+
+
+def capture_isvc_kueue_baseline(client: DynamicClient, isvc: InferenceService) -> ISVCKueueBaseline:
+    """Capture pre-upgrade baseline for a Kueue-integrated raw InferenceService."""
+    baseline = capture_isvc_baseline(client=client, isvc=isvc)
+    runtime_name = baseline["runtime_name"]
+    isvc.get()
```
dbasunag: Please remove.

dbasunag (changes requested):

conftest.py
```diff
@@ -425,7 +425,8 @@ def pytest_runtest_setup(item: Item) -> None:
             LOGGER.error(f"Database error: {db_exception}. Must-gather collection may not be accurate")
 
     if KServeDeploymentType.RAW_DEPLOYMENT.lower() in item.keywords:
-        item.fixturenames.insert(0, "enabled_kserve_in_dsc")
+        if not ("kueue" in item.keywords and ("pre_upgrade" in item.keywords or "post_upgrade" in item.keywords)):
+            item.fixturenames.insert(0, "enabled_kserve_in_dsc")
```
mwaykole: This condition is fragile. You're embedding knowledge of which fixture combinations handle KServe enablement into the root `pytest_runtest_setup`, which every contributor needs to understand. If someone adds a new `kueue + raw_deployment` upgrade test in the future without knowing about this carve-out, `enabled_kserve_in_dsc` silently won't run and the test will behave unpredictably.

A cleaner approach: have `_ensure_kserve_enabled_for_upgrade` fixture itself handle the conflict with `enabled_kserve_in_dsc` (e.g., by setting a marker or session flag that `enabled_kserve_in_dsc` can check to no-op). That way the logic stays close to where it matters rather than in the global hook.

tests/model_serving/model_server/upgrade/conftest.py
```diff
@@ -1429,3 +1414,518 @@ def _capture_and_save_llmd_baseline(
...
+        yield local_queue
+        if teardown_resources:
+            local_queue.clean_up()
+            ClusterQueue(client=admin_client, name=cluster_queue_name).clean_up()
+            ResourceFlavor(client=admin_client, name=resource_flavor_name).clean_up()
+    else:
+        local_queue = LocalQueue(
+            client=admin_client,
+            name=local_queue_name,
+            cluster_queue=cluster_queue_name,
```
mwaykole: +1 to dbasunag's comment. This and `_ensure_kueue_available_for_upgrade`, `_create_kueue_upgrade_resources`, `_capture_and_save_isvc_kueue_baseline` are all plain helper functions, not fixtures. Move them to `utils.py`. This conftest is already 1900+ lines, adding ~300 more lines of utility logic makes it harder to navigate.

tests/model_serving/model_server/upgrade/conftest.py
```diff
@@ -1429,3 +1414,518 @@ def _capture_and_save_llmd_baseline(
...
+                    ),
+                    teardown=teardown_resources,
+                ),
+                create_local_queue(
+                    client=admin_client,
+                    name=local_queue_name,
+                    cluster_queue=cluster_queue_name,
+                    namespace=namespace,
+                    teardown=teardown_resources,
+                ) as local_queue,
```
mwaykole: This changes existing behavior from `pytest.skip` to `pytest.fail` for the LLMD Kueue upgrade path (this fixture is used by the existing LLMD tests too, not just the new KServe tests). On a cluster without Kueue, existing LLMD upgrade tests that used to skip gracefully will now hard-fail. Was this intentional? If so, it should be a separate commit with its own justification since it affects more than just the new KServe+Kueue lane.

tests/model_serving/model_server/upgrade/conftest.py
```diff
@@ -1429,3 +1414,518 @@ def _capture_and_save_llmd_baseline(
...
+                isvc=isvc,
+            )
+
+
+@pytest.fixture
+def skip_if_not_raw_deployment(
+    kserve_kueue_upgrade_inference_service: InferenceService,
+) -> None:
+    """Skip tests when the Kueue upgrade ISVC is not deployed in RawDeployment mode."""
+    kserve_kueue_upgrade_inference_service.get()
```
mwaykole: Baseline capture happens inside the ISVC fixture teardown (after `yield`). This means if the pre-upgrade test `test_kueue_scale_and_gate` fails, the baseline is still captured with whatever state the ISVC is in at that point. The post-upgrade tests would then compare against a baseline from a failed pre-upgrade run.

Capture should happen explicitly in the last pre-upgrade test (after all assertions pass), or add a guard that skips capture if any pre-upgrade test failed.

tests/model_serving/model_server/upgrade/test_upgrade_kserve_kueue_raw.py
```diff
@@ -0,0 +1,344 @@
...
+                sleep=5,
+                func=lambda: _get_deployment_status_replicas(deployment),
+            ):
+                if status_replicas == KSERVE_KUEUE_SCALED_REPLICAS:
+                    break
+        except TimeoutExpiredError:
+            pytest.fail(
+                f"Timeout waiting for Deployment to scale to {KSERVE_KUEUE_SCALED_REPLICAS} replicas, "
+                f"got {status_replicas}"
+            )
```
mwaykole, line 165: The `isvc.update(isvc_dict)` call here passes the full object dict. This is risky -- it includes `status`, `metadata.resourceVersion`, and other server-managed fields. If there's any staleness (which there will be since the fixture is session-scoped), you'll get a 409 Conflict.

Use a targeted patch instead:
```python
isvc.update(
    resource_dict={
        "metadata": {"name": isvc.name},
        "spec": {"predictor": {"minReplicas": KSERVE_KUEUE_SCALED_REPLICAS}},
    }
)
```
This is also the pattern used everywhere else in this PR for DSC/DSCI updates.

tests/model_serving/model_server/upgrade/test_upgrade_kserve_kueue_raw.py
```diff
@@ -0,0 +1,344 @@
...
+            inference_type=Inference.INFER,
+            protocol=Protocols.HTTP,
+            use_default_query=True,
+            inference_timeout=KSERVE_KUEUE_INFERENCE_TIMEOUT,
+        )
+
+
+def _get_deployment_status_replicas(deployment: Deployment) -> int:
+    deployment.get()
+    return deployment.instance.status.replicas
```
mwaykole: The `.get()` here IS needed -- `_get_deployment_status_replicas` is called inside `TimeoutSampler` and must refresh the object each iteration. But dbasunag is right that a 2-line wrapper is unnecessary.

Just inline it:
```python
for _ in TimeoutSampler(
    wait_timeout=Timeout.TIMEOUT_2MIN,
    sleep=5,
    func=lambda: (deployment.get() or True) and deployment.instance.status.replicas,
):
```

Or better yet, use `deployment.wait_for_replicas()` if it supports waiting for a specific count.

tests/model_serving/model_server/upgrade/utils.py
```diff
@@ -1,25 +1,257 @@
...
+
+
+def _restore_dsci_servicemesh_state(
+    admin_client: DynamicClient,
+    dsci_resource: DSCInitialization,
+    namespace: str,
+    dsci_servicemesh_state_cm_name: str,
+) -> None:
+    """Restore original DSCI serviceMesh state from the saved ConfigMap."""
+    state_cm = ConfigMap(client=admin_client, name=dsci_servicemesh_state_cm_name, namespace=namespace)
```
mwaykole: Missing type annotation on `dsc_resource`. Should be `dsc_resource: DataScienceCluster`.

tests/model_serving/model_server/upgrade/utils.py
```diff
@@ -578,12 +888,25 @@ def verify_isvc_pods_not_restarted_against_baseline(
     baseline_pod_names = set(baseline_restart_counts.keys())
     missing_pods = baseline_pod_names - current_pod_names
     new_pods = current_pod_names - baseline_pod_names
-    if missing_pods or new_pods:
+    if missing_pods:
+        raise PodContainersRestartError(
+            f"Pod set changed after upgrade for {isvc.name}. missing={sorted(missing_pods)}, new={sorted(new_pods)}"
+        )
+    if new_pods and not allow_new_pods:
         raise PodContainersRestartError(
```
mwaykole, line 1004: These two blocks raise the same error with the same message. Consolidate:
```python
if missing_pods or (new_pods and not allow_new_pods):
    raise PodContainersRestartError(
        f"Pod set changed after upgrade for {isvc.name}. missing={sorted(missing_pods)}, new={sorted(new_pods)}"
    )
```

tests/model_serving/model_server/upgrade/utils.py
```diff
@@ -350,6 +601,60 @@ def capture_isvc_baseline(client: DynamicClient, isvc: InferenceService) -> ISVC
...
+    runtime_name = baseline["runtime_name"]
+    isvc.get()
+    total_copies = read_isvc_total_copies(isvc=isvc)
+    min_replicas = isvc.instance.spec.predictor.get("minReplicas", 1)
+    kueue_baseline: ISVCKueueBaseline = {
+        **baseline,
+        "kueue_integration_stats": get_isvc_kueue_integration_stats(
+            client=client,
+            isvc=isvc,
+            runtime_name=runtime_name,
```
mwaykole, line 536: The `isvc.get()` here is called right after `capture_isvc_baseline` which also accesses the isvc object. If `capture_isvc_baseline` doesn't refresh the ISVC (it reads pods and runtime, not the ISVC status), then this `.get()` is actually needed before `read_isvc_total_copies`. But it should have a brief comment explaining WHY, otherwise dbasunag will keep asking to remove it. Something like:
```python
# Refresh ISVC to get current modelStatus.copies after baseline pod capture
isvc.get()
```
Alternatively, have `read_isvc_total_copies` call `.get()` internally so callers don't need to remember.

tests/model_serving/model_server/upgrade/test_upgrade_kserve_kueue_raw.py
```diff
@@ -0,0 +1,344 @@
...
+        )
+        isvc_name = kserve_kueue_upgrade_inference_service.name
+        assert isvc_name in baselines, f"ISVC '{isvc_name}' not in baseline. Available: {list(baselines.keys())}"
+        return baselines[isvc_name]
+
+    @pytest.mark.post_upgrade
+    @pytest.mark.dependency(name="kserve_kueue_isvc_exists")
+    def test_isvc_exists_post_upgrade(
+        self,
+        kserve_kueue_upgrade_inference_service: InferenceService,
```
mwaykole, line 227: nit: `baseline` fixture returns `dict` but could be `ISVCKueueBaseline` for better IDE support and type safety. You defined the TypedDict for exactly this purpose.

tests/model_serving/model_server/upgrade/conftest.py
```diff
@@ -1429,3 +1414,518 @@ def _capture_and_save_llmd_baseline(
...
+            client=admin_client,
+            name=local_queue_name,
+            cluster_queue=cluster_queue_name,
+            namespace=namespace,
+        )
+        cluster_queue = ClusterQueue(client=admin_client, name=cluster_queue_name)
+        resource_flavor = ResourceFlavor(client=admin_client, name=resource_flavor_name)
+        missing_resources = [
+            resource_label
+            for resource_label, resource in (
```
mwaykole: The `dsc_resource.get()` here before `_is_kserve_ready` IS needed -- you're checking readiness after a potential patch and reconciliation, so the object needs to be current. But two lines later you call `dsc_resource.wait_for_condition()` which will internally refresh the object anyway.

So the flow is: get() -> check ready -> if not ready, wait_for_condition(). The `get()` + `_is_kserve_ready` check is an optimization to skip the 5-minute wait if already ready. That's fine but needs a comment, because without context it looks redundant.

utilities/inference_utils.py
```diff
@@ -454,7 +458,48 @@ def run_inference(
             ValueError: If inference fails
 
         """
```
mwaykole, line 460: Good refactor. The closure-based retry with configurable timeout is cleaner than the previous class-level decorator. One thing: if `inference_timeout=0` is passed, `0 or Timeout.TIMEOUT_30SEC` evaluates to `Timeout.TIMEOUT_30SEC`. Consider `inference_timeout if inference_timeout is not None else Timeout.TIMEOUT_30SEC` for correctness.

mwaykole (commented):
Went through the full diff. Overall the test coverage for KServe raw + Kueue upgrade is solid and the pre/post upgrade split with ConfigMap-based state persistence is well thought out. A few things need addressing before this can merge:

**Must fix:**

1. **Root `pytest_runtest_setup` change (conftest.py)** — Hardcoding a kueue+upgrade exclusion in the global hook is brittle. This needs a better mechanism that doesn't require every future contributor to know about this carve-out.

2. **Non-fixture helpers in conftest.py** — `_ensure_kserve_enabled_for_upgrade`, `_ensure_kueue_available_for_upgrade`, `_create_kueue_upgrade_resources`, `_capture_and_save_isvc_kueue_baseline` should move to `utils.py`. dbasunag already flagged this and the repo convention is fixtures-only in conftest.

3. **Skip→Fail behavioral change for existing LLMD tests** — `_ensure_kueue_available_for_upgrade` (which backs `ensure_kueue_for_upgrade`) now calls `pytest.fail` instead of `pytest.skip` when Kueue is missing. This is a breaking change for existing LLMD upgrade tests that used to skip gracefully on non-Kueue clusters.

4. **Full-object update for ISVC scaling** — `isvc.update(isvc_dict)` in `test_kueue_scale_and_gate` sends the entire object including status and resourceVersion. Use a targeted merge patch like every other update in this PR.

5. **Commit message** — Has `Co-authored-by: Cursor <cursoragent@cursor.com>`. Must be removed per repo policy.

**Should fix:**

6. Baseline capture happens inside ISVC fixture teardown — if pre-upgrade tests fail, baseline is still captured with broken state.
7. `_is_kserve_ready` missing type annotation for `dsc_resource`.
8. Duplicate `PodContainersRestartError` raises should be consolidated.
9. `baseline` fixture return type should be `ISVCKueueBaseline`, not bare `dict`.

See inline comments for details.

tests/model_serving/model_server/upgrade/conftest.py
sukumars321: Moved necessary helpers to utils and removed dsc/dsci mutation code.

tests/model_serving/model_server/upgrade/conftest.py
sukumars321: Not intentional, reverted.

tests/model_serving/model_server/upgrade/utils.py
```diff
@@ -350,6 +497,281 @@ def capture_isvc_baseline(client: DynamicClient, isvc: InferenceService) -> ISVC
...
+    Args:
+        client: Kubernetes dynamic client.
+        isvc: The InferenceService to inspect.
+        runtime_name: ServingRuntime name used for pod label selection.
+
+    Returns:
+        Dict with ``running`` and ``gated`` pod counts.
+    """
+    from utilities.general import create_isvc_label_selector_str
+    from utilities.kueue_utils import check_gated_pods_and_running_pods
```
mwaykole: `create_isvc_label_selector_str` and `check_gated_pods_and_running_pods` have no circular dependency with this module — move these to the top-level imports. Inline imports here just make it harder to see what this module depends on.

utilities/data_science_cluster_utils.py
```diff
@@ -29,11 +40,11 @@ def update_components_in_dsc(
 
     """
     dsc_dict: dict[str, dict[str, dict[str, dict[str, str]]]] = {}
-    dsc_components = dsc.instance.spec.components
     component_to_reconcile = {}
 
     for component_name, desired_state in components.items():
```
mwaykole, line 45: This changes error behavior for callers. Before: `dsc_components[component_name]` raised `KeyError` if the component wasn't in DSC spec. Now: returns `None`, and `None != desired_state` is `True`, so we silently try to patch a component that may not exist. If that's intentional for new components added during upgrade, add a log line for the `None` case so it's visible. If not, raise early.

tests/model_serving/model_server/upgrade/kserve_kueue_upgrade_config.py
```diff
@@ -0,0 +1,45 @@
...
+KSERVE_KUEUE_CLUSTER_QUEUE: str = "upgrade-kserve-cluster-queue"
+KSERVE_KUEUE_RESOURCE_FLAVOR: str = "upgrade-kserve-flavor"
+
+# Pod resources sized for minimal OVMS raw deployment footprint.
+# Kueue admits on the sum of ALL container requests in the pod. Headed raw ISVCs
+# also inject kube-rbac-proxy (~100m CPU / 64Mi memory), so one pod costs about
+# 200m CPU + 1088Mi memory when the model container requests 100m / 1Gi.
+KSERVE_KUEUE_POD_CPU_REQUEST: str = "100m"
+KSERVE_KUEUE_POD_MEMORY_REQUEST: str = "1Gi"
+KSERVE_KUEUE_POD_CPU_LIMIT: int = 1
```
mwaykole: CPU request is `str` ("100m") but CPU limit is `int` (1). K8s handles both, but the type mismatch forces the `dict[str, str | int]` union and makes it easy to trip over when comparing or serializing. Pick one form — either `"1"` or `int` for both.

tests/model_serving/model_server/upgrade/conftest.py
```diff
@@ -1429,3 +1369,242 @@ def _capture_and_save_llmd_baseline(
...
+        namespace=kserve_kueue_upgrade_namespace.name,
+        teardown_resources=teardown_resources,
+    )
+
+
+@pytest.fixture(scope="session")
+def kserve_upgrade_kueue_resources(
+    pytestconfig: pytest.Config,
+    admin_client: DynamicClient,
+    ensure_kueue_for_kserve_upgrade,
```
mwaykole: Missing type annotation — should be `ensure_kueue_for_kserve_upgrade: None` to match the fixture's `Generator[None, ...]` yield type.

tests/model_serving/model_server/upgrade/utils.py
```diff
@@ -350,6 +497,281 @@ def capture_isvc_baseline(client: DynamicClient, isvc: InferenceService) -> ISVC
...
+            resource_label
+            for resource_label, resource in (
+                (f"LocalQueue '{local_queue_name}' in namespace '{namespace}'", local_queue),
+                (f"ClusterQueue '{cluster_queue_name}'", cluster_queue),
+                (f"ResourceFlavor '{resource_flavor_name}'", resource_flavor),
+            )
+            if not resource.exists
+        ]
+        if missing_resources:
+            pytest.fail(
```
mwaykole, line 674: Post-upgrade teardown creates new `ClusterQueue` and `ResourceFlavor` objects for cleanup when you already have `cluster_queue` and `resource_flavor` from the existence check above. Reuse those instead of constructing new instances.

mwaykole (commented):
Good progress since last round — helpers moved to `utils.py`, targeted merge patch for ISVC scaling, `pytest.skip` preserved for missing Kueue, Co-authored-by trailer removed. A few remaining items inline.

mwaykole (dismissed):

tests/model_serving/model_server/upgrade/conftest.py
```diff
@@ -1429,3 +1384,255 @@ def _capture_and_save_llmd_baseline(
...
+                )
+
+
+@pytest.fixture
+def skip_if_not_raw_deployment(
+    kserve_kueue_upgrade_inference_service: InferenceService,
+) -> None:
+    """Skip tests when the Kueue upgrade ISVC is not deployed in RawDeployment mode."""
+    # Refresh ISVC before reading deployment mode from status/annotations.
+    kserve_kueue_upgrade_inference_service.get()
```
dbasunag: get() is a class method of resource. So the ideal way of refreshing the kserve_kueue_upgrade_inference_service would be either via class_name.get(pass the right param) or via kserve_kueue_upgrade_inference_service.instance.

tests/model_serving/model_server/upgrade/test_upgrade_kserve_kueue_raw.py
```diff
@@ -0,0 +1,378 @@
...
+        deployment = deployments[0]
+
+        status_replicas = None
+        try:
+            for status_replicas in TimeoutSampler(
+                wait_timeout=Timeout.TIMEOUT_2MIN,
+                sleep=5,
+                # Refresh Deployment each poll; status.replicas is otherwise stale.
+                # wait_for_replicas() is not used: with Kueue gating, ready < desired.
+                func=lambda: (deployment.get() or True) and deployment.instance.status.replicas,
```
dbasunag: please use deployment.instance and please don't use `or True`, it can't possibly work with `deployment.instance.status.replicas`

tests/model_serving/model_server/upgrade/test_upgrade_kserve_kueue_raw.py
dbasunag: @sukumars321 please don't resolve comment, if that has not been addressed. Please remove this line and use isvc.instance instead.

tests/model_serving/model_server/upgrade/utils.py
```diff
@@ -112,11 +173,99 @@ def verify_model_status_loaded(isvc: InferenceService) -> None:
...
+    model_status = isvc.instance.status.modelStatus
+    if not model_status or model_status.copies is None:
+        raise AssertionError(f"modelStatus.copies not populated for InferenceService {isvc.name}")
+    return model_status.copies.totalCopies
+
+
+def _read_isvc_status_url(isvc: InferenceService) -> str:
+    """Refresh the ISVC and return status.url when present."""
+    # TimeoutSampler must re-fetch; instance alone can retain a stale empty URL.
+    isvc.get()
```
dbasunag: Please remove

dbasunag (changes requested):

tests/model_serving/model_server/upgrade/test_upgrade_kserve_kueue_raw.py
sukumars321: Addressed.

tests/model_serving/model_server/upgrade/test_upgrade_kserve_kueue_raw.py
sukumars321: Addressed.

tests/model_serving/model_server/upgrade/test_upgrade_kserve_kueue_raw.py
sukumars321: Addressed.

dbasunag (approved):

mwaykole (approved):

PR #1756, opened 2026-06-15
feat(nemo_guardrails):test TAI MCP gateway guardrails enablement

christinaexyou:
# Pull Request
## Summary
Adds tests and helper functions for TrustyAI Service Operator MCP Gateway guardrails enablement via the NeMo Guardrails controller

Checks that the NeMo Guardrails controller executes the following steps in order:
1) Discovers a EnvoyFilter in the `mcpGateway.namespace` with name `mcpGateway.name`. We create this resource in the tests but in practice it would be deployed by the BBR plugin
2) Deploys the NeMo Guardrails CR 
3) Creates a new EnvoyFilter named `mcp-sse-strip`
4) Exposes functional `v1/guardrail/checks` endpoint 

## Related Issues

- JIRA:  [RHAISTRAT-1721](https://redhat.atlassian.net/browse/RHAISTRAT-1721?atlOrigin=eyJpIjoiYjNhZjMxNmY1Y2YxNDU5OGEyNzg1OTMyMjA0YzNiZmUiLCJwIjoiaiJ9)

## Please review and indicate how it has been tested

- [x] Locally 
```
$ uv run pytest tests/ai_safety/nemo_guardrails/test_nemo_guardrails.py -k "mcp" -v
```
```
_________ 6 of 6 completed, 6 Pass, 0 Fail, 0 Skip, 0 XPass, 0 XFail, 0 Error, 0 ReRun _________
========================= 6 passed, 18 deselected in 164.33s (0:02:44) =========================

========================= 6 passed, 18 deselected in 164.34s (0:02:44) =========================
```
- [ ] Jenkins

## Additional Requirements

- [ ] If this PR introduces a new test image, did you create a PR to mirror it in disconnected environment?
- [ ] If this PR introduces new marker(s)/adds a new component, was relevant ticket created to update relevant Jenkins job?

## Summary by CodeRabbit

* **Tests**
  * Expanded NeMo Guardrails integration coverage for the MCP gateway end-to-end, including MCP CRD presence, MCP spec wiring, authenticated chat/check backend behavior, and operator-created EnvoyFilter expectations.
  * Added new MCP-focused integration test scenarios to verify route exposure and successful request handling.

* **Chores**
  * Improved test reliability by dynamically deriving TLS verification settings from the admin connection.
  * Refined healthcheck and readiness waiting to align with the EnvoyFilter-based setup.

[...]

PR #1933, opened 2026-07-06
Mcp gateway guardrails enablement fixes

Raul-Prisecaru:
# Pull Request

## Summary

Within the `test_nemo_guardrails_mcp_gateway` file, two of the tests fails:
- `test_nemo_mcp_cr_spec` - fails because spec.mcpGateway is empty, this is fixed by adjusting where it is actually located which is under `nemo_guardrails_mcp.kind_dict["spec"]["mcpGateway"]` as opposed to `nemo_guardrails_mcp.instance.spec.mcpGateway`
- `test_nemo_mcp_envoy_filter_created` - fails because it coould not find the correct EnvoyFilter in the specified namespace and the issue comes from `bbr_envoy_filter` fixatures. Inside the metadata contains the incorrect name to which was changed from `"name": BBR_ENVOY_FILTER_NAME,` to `"name": MCP_ENVOY_FILTER_NAME`

## Related Issues

- Fixes: https://github.com/opendatahub-io/opendatahub-tests/pull/1756
- JIRA: https://redhat.atlassian.net/browse/RHOAIENG-74756

- [x] Locally

## Additional Requirements

Screenshot From 2026-07-06 11-46-53

## Summary by CodeRabbit

* **New Features**
  * Added NeMo Guardrails integration with MCP Gateway routing.
  * Added support for authenticated chat and health-check requests through the MCP Gateway.
  * Added validation for MCP Gateway resources and Envoy configuration.

* **Bug Fixes**
  * Improved TLS certificate handling across managed and self-managed environments.
  * Enhanced deployment, route, and backend readiness checks for more reliable validation.


tests/ai_safety/nemo_guardrails/conftest.py
```diff
@@ -464,70 +536,320 @@ def nemo_guardrails_config_update_route(
...
+@pytest.fixture(scope="session")
+def gateway_crd(
+    admin_client: DynamicClient,
+) -> Generator[None, Any, Any]:
+    """Verify gateways.gateway.networking.k8s.io CRD is present; built-in on OCP 4.19+."""
+    crd_name = "gateways.gateway.networking.k8s.io"
+    crd = CustomResourceDefinition(client=admin_client, name=crd_name)
+
+    if not crd.exists:
+        pytest.skip(
```
dbasunag: We should not silently skip. Please fail if we expect it to be present and it does not exist.

tests/ai_safety/nemo_guardrails/conftest.py
```diff
@@ -464,70 +536,320 @@ def nemo_guardrails_config_update_route(
...
+) -> Generator[None, Any, Any]:
+    """Install OpenShift Service Mesh 3 if not already present; yield if Istio CRDs already exist."""
+    envoy_filter_crd = CustomResourceDefinition(
+        client=admin_client,
+        name="envoyfilters.networking.istio.io",
+    )
+
+    if envoy_filter_crd.exists:
+        yield
+        return
```
dbasunag: Please use if/else. Then you won't need both yield and return

tests/ai_safety/nemo_guardrails/conftest.py
```diff
@@ -464,70 +536,320 @@ def nemo_guardrails_config_update_route(
...
+        yield
+
+        uninstall_operator(
+            admin_client=admin_client,
+            name=operator_name,
+            operator_namespace=operator_namespace,
+            clean_up_namespace=False,
+        )
+    else:
+        yield
```
dbasunag, line 632: either we expect the operator to be installed or not. If testops is required to have this installed, then we should fail. Else we should do the installation. Otherwise it gets hard to figure out who owns what.

tests/ai_safety/nemo_guardrails/conftest.py
```diff
@@ -464,70 +536,320 @@ def nemo_guardrails_config_update_route(
...
+) -> Generator[None, Any, Any]:
+    """Install the mcp-gateway operator via OLM if not already present."""
+    operator_name = "mcp-gateway"
+    operator_namespace = "openshift-operators"
+
+    subscription = Subscription(
+        client=admin_client,
+        namespace=operator_namespace,
+        name=operator_name,
+    )
```
dbasunag, line 694: Same comment here

dbasunag (changes requested):

tests/ai_safety/nemo_guardrails/conftest.py
kpunwatk: I agree with Debarati, @Raul-Prisecaru. I don't think this operator should be installed by default as part of the RHOAI installation, especially since Service Mesh is already installed.
Could you please remove this fixture?

kpunwatk (changes requested):

kpunwatk:
@Raul-Prisecaru fix the pre-commit and conflicts

tests/ai_safety/nemo_guardrails/conftest.py
Raul-Prisecaru: > I agree with Debarati, [@Raul-Prisecaru](https://github.com/Raul-Prisecaru). I don't think this operator should be installed by default as part of the RHOAI installation, especially since Service Mesh is already installed. Could you please remove this fixture?

I believe the installation of the operator is already removed now with the latest commits I pushed. Now we simply just check to see if it's installed and if not, we fail the test and prompting the user to install it

Thought I should remember to update the description of the function to reflect that

tests/ai_safety/nemo_guardrails/conftest.py
```diff
@@ -467,61 +538,280 @@ def verify_guardrails_healthcheck(
...
+        token=None,
+        ca_bundle_file=get_tls_verify(client=admin_client),
+    )
+
+
+@pytest.fixture(scope="session")
+def installed_istio(
+    admin_client: DynamicClient,
+) -> Generator[None, Any, Any]:
+    """Install OpenShift Service Mesh 3 if not already present; yield if Istio CRDs already exist."""
```
kpunwatk: change the description of the fixture

tests/ogx/vector_io/test_vector_stores_evaluation.py
```diff
@@ -3,7 +3,7 @@
 import pytest
 import structlog
 from ragas import EvaluationDataset, SingleTurnSample, evaluate
-from ragas.metrics import AnswerRelevancy, ContextPrecision, ContextRecall, Faithfulness
+from ragas.metrics.collections import AnswerRelevancy, ContextPrecision, ContextRecall, Faithfulness
```
kpunwatk: remove this @Raul-Prisecaru no need to change this

kpunwatk (changes requested):

Raul-Prisecaru:
Screenshot From 2026-07-27 16-37-36

kpunwatk (approved):

dbasunag (approved):

PR #1941, opened 2026-07-06
test: cover agent catalog plugin in plugin_arch tests

dbasunag:
# Pull Request

## Related Issues

- Fixes: 
- JIRA: 

- [ ] Locally
- [ ] Jenkins

## Additional Requirements

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved catalog endpoint readiness checks and startup validation for a broader set of plugin types.
  * Updated catalog base URL handling to work with route-based access, helping readiness checks resolve the service more reliably.

* **Tests**
  * Added coverage for model catalog pod lookup behavior.
  * Expanded unified catalog server test coverage to include an additional plugin registration and health check scenario.


fege (approved):
/lgtm

tests/ai_hub/plugin_arch/conftest.py
```diff
@@ -15,9 +16,10 @@
 
 
 @pytest.fixture(scope="class")
-def catalog_base_url(model_catalog_rest_url: list[str]) -> str:
+def catalog_base_url(model_catalog_routes: list[Route]) -> str:
     """Base URL for the catalog server without the API path."""
-    return model_catalog_rest_url[0].split("/api/")[0]
+    assert model_catalog_routes, "Model catalog routes not found"
```
sheltoncyril, line 21: nitpick: I think this assertion should be better especially as this is a list, we should be explicit that we are checking for one valid route or that it contains at least one valid route.

sheltoncyril (approved):
/lgtm

tests/ai_hub/plugin_arch/conftest.py
dbasunag: That is exactly what we are doing. We are ensuring that the list is not empty - which translates to having at least one route. :)

PR #1942, opened 2026-07-06
test: custom agent catalog tests

dbasunag:
# Pull Request

## Related Issues

- Fixes: 
- JIRA: 

- [ ] Locally
- [ ] Jenkins

## Additional Requirements

## Summary by CodeRabbit

* **Tests**
  * Expanded AI Hub agent catalog coverage for search, filtering, pagination, sort order, and source availability.
  * Added checks for agent details, required metadata, filter options, and consistent results across pages.
  * Improved validation of catalog behavior for matching keywords, labels, and framework-based queries.


tests/ai_hub/agent_catalog/utils.py
```diff
@@ -0,0 +1,128 @@
...
+
+        sort_order = order_params.get("sortOrder", "ASC")
+        if seen_ids:
+            if sort_order == "ASC":
+                assert int(agent_id) > int(seen_ids[-1]), (
+                    f"Page {page_idx} id '{agent_id}' is not greater than previous id '{seen_ids[-1]}'"
+                )
+            elif sort_order == "DESC":
+                assert int(agent_id) < int(seen_ids[-1]), (
+                    f"Page {page_idx} id '{agent_id}' is not less than previous id '{seen_ids[-1]}'"
```
lugi0, line 302: The ordering check casts IDs to int for comparison. The API defines id as a string. This works today because the MLMD-backed implementation generates numeric IDs, but it's an implementation detail. Safer to just verify monotonicity of the string-as-received or check that the returned order matches a separately-sorted list.

tests/ai_hub/agent_catalog/utils.py
```diff
@@ -0,0 +1,128 @@
...
+        headers=headers,
+        params={"filterQuery": filter_query},
+    )
+    total_items = all_response.get("size", 0)
+    LOGGER.info(f"Total items matching filter '{filter_query}': {total_items}")
+
+    items: list[dict[str, Any]] = []
+    next_page_token: str | None = None
+
+    for page_num in range(1, total_items + 1):
```
lugi0, line 237: CodeRabbit already flagged this. range(1, total_items + 1) iterates total_items times — correct only when page_size=1. The fix is math.ceil(total_items / page_size) for the loop bound and adjusting the last-page item count assertion. Currently masked but will break the moment someone calls with page_size=2.

tests/ai_hub/utils.py
```diff
@@ -1058,6 +1058,67 @@ def wait_for_mcp_catalog_api(
...
+        LOGGER.info(
+            f"Agent catalog API returned {current_size} agents (stable: {stable_count}/{consecutive_stable_checks})"
+        )
+    final_size = data.get("size", 0)
+    if final_size < min_agents:
+        raise AssertionError(
+            f"Agent catalog API did not stabilize with at least {min_agents} agents "
+            f"within {wait_timeout}s (last size={final_size})"
+        )
+    return data
```
lugi0: When the TimeoutSampler loop exhausts without stabilization but final_size >= min_agents, the function returns the last (potentially fluctuating) payload instead of raising. The docstring promises an AssertionError. This also means min_agents=0 on teardown will succeed even if every API call failed (since data stays {}).

  Fix: always raise AssertionError after the loop exits without hitting the stabilization threshold. The early return data inside the loop is the only success path.

tests/ai_hub/agent_catalog/conftest.py
```diff
@@ -0,0 +1,133 @@
...
+    )
+    size = response.get("size", 0)
+    assert size >= TEST_AGENT_COUNT, f"Expected at least {TEST_AGENT_COUNT} agents after catalog patch, got size={size}"
+    return response
+
+
+@pytest.fixture(scope="class")
+def default_agents(agents_response: dict) -> dict:
+    """Return the agents list loaded from the test agent catalog source."""
+    return agents_response
```
lugi0: This is a pass-through alias. Tests that use default_agents could use agents_response directly. If the intent is to distinguish "test catalog agents" from "all agents," the fixture should actually filter to only the test source, but it doesn't.

tests/ai_hub/agent_catalog/utils.py
```diff
@@ -0,0 +1,128 @@
...
+        if seen_ids:
+            if sort_order == "ASC":
+                assert int(agent_id) > int(seen_ids[-1]), (
+                    f"Page {page_idx} id '{agent_id}' is not greater than previous id '{seen_ids[-1]}'"
+                )
+            elif sort_order == "DESC":
+                assert int(agent_id) < int(seen_ids[-1]), (
+                    f"Page {page_idx} id '{agent_id}' is not less than previous id '{seen_ids[-1]}'"
+                )
+        seen_ids.append(agent_id)
```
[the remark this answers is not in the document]
dbasunag: In https://github.com/kubeflow/hub/pull/2887/changes#diff-fde3e16776a6aa8c48187e73eaf5084632e2060f9ba592a32f467ca990a6a071
  - db_agent.go — GetByID(id int32), DeleteByID(id int32) in the repository interface
  - service/agent.go — strconv.FormatInt(int64(*dbAgent.GetID()), 10) to serialize to the API response
  - The underlying MLMD schema.Context table uses integer primary keys — this is shared by models, MCP servers, and agents

  The ID is structurally int32 in the database, serialized to string for the API. Same pattern as models and MCP servers, which also use int() casting in their test assertions.

tests/ai_hub/agent_catalog/utils.py
dbasunag: Based on https://github.com/kubeflow/hub/pull/2887/changes#diff-fde3e16776a6aa8c48187e73eaf5084632e2060f9ba592a32f467ca990a6a071

  - db_agent.go — GetByID(id int32), DeleteByID(id int32) in the repository interface
  - service/agent.go — strconv.FormatInt(int64(*dbAgent.GetID()), 10) to serialize to the API response
  - The underlying MLMD schema.Context table uses integer primary keys — this is shared by models, MCP servers, and agents

  The ID is structurally int32 in the database, serialized to string for the API. Same pattern as models and MCP servers, which also use int() casting in their test assertions.

lugi0: It is true that today the IDs are always ints, but we don't know for sure that this will always be the case. This is just a hardening measure for future-proofing, so I am ok with the current version as well.

tests/ai_hub/agent_catalog/utils.py
dbasunag: Changed page size to fixed value of 1, to be aligned with what we did in mcp areas.

fege (dismissed):
/lgtm

tests/ai_hub/agent_catalog/search/test_agent_catalog_filtering.py
```diff
@@ -0,0 +1,169 @@
+from typing import Any, Self
```
lugi0, line 1: this file is missing the tier and usefixtures marks

tests/ai_hub/agent_catalog/search/test_agent_catalog_pagination.py
```diff
@@ -0,0 +1,127 @@
+from typing import Self
```
lugi0, line 1: this file is missing the tier and usefixtures marks

tests/ai_hub/agent_catalog/search/test_agent_catalog_search.py
```diff
@@ -0,0 +1,78 @@
+from typing import Any, Self
```
lugi0, line 1: this file is missing the tier and usefixtures marks

tests/ai_hub/constants.py
```diff
@@ -77,6 +77,7 @@ class ModelRegistryEndpoints:
 CUSTOM_CATALOG_ID1: str = "sample_custom_catalog1"
 DEFAULT_MODEL_CATALOG_CM: str = "default-catalog-sources"
 MCP_CATALOG_API_PATH: str = "/api/mcp_catalog/v1alpha1/"
+AGENT_CATALOG_API_PATH: str = "/api/agent_catalog/v1alpha1/"
```
lugi0, line 80: this was already defined in #1955 https://github.com/opendatahub-io/opendatahub-tests/pull/1955/changes#diff-2a6ee3d60fa30e653528c86620f1ddc5786aabe8866606404078c8f992923323R2

tests/ai_hub/agent_catalog/constants.py
```diff
@@ -0,0 +1,96 @@
+# Custom test catalog injected via ConfigMap patch (mirrors MCP_CATALOG_SOURCE pattern).
```
lugi0: there is already an ai_hub/agent_catalog/config/constants.py file defined from #1955 (https://github.com/opendatahub-io/opendatahub-tests/pull/1955/changes#diff-2a6ee3d60fa30e653528c86620f1ddc5786aabe8866606404078c8f992923323R1) - can we use a single file?

tests/ai_hub/agent_catalog/search/test_agent_catalog_filtering.py
```diff
@@ -0,0 +1,169 @@
...
+        """
+        unfiltered = execute_get_command_with_retry(
+            url=f"{agent_catalog_rest_urls[0]}agents",
+            headers=model_registry_rest_headers,
+            params={"pageSize": "1000"},
+        )
+        filtered = execute_get_command_with_retry(
+            url=f"{agent_catalog_rest_urls[0]}agents",
+            headers=model_registry_rest_headers,
+            params={"sourceLabel": "Red Hat Agents", "pageSize": "1000"},
```
lugi0: it would be best to not hardcode the source name here

tests/ai_hub/agent_catalog/search/test_agent_catalog_filtering.py
dbasunag: tier2 is the default and the usefixtures is only needed for tier1. This is intentional.

tests/ai_hub/agent_catalog/search/test_agent_catalog_search.py
dbasunag: tier2 is the default and the usefixtures is only needed for tier1. This is intentional.

tests/ai_hub/agent_catalog/search/test_agent_catalog_pagination.py
dbasunag: tier2 is the default and the usefixtures is only needed for tier1. This is intentional.

tests/ai_hub/agent_catalog/constants.py
dbasunag: We can't have everything in a single file. But I will rearrange and ensure things are placed where they belong.

tests/ai_hub/agent_catalog/search/test_agent_catalog_filtering.py
dbasunag: Moving to a constant.

tests/ai_hub/constants.py
dbasunag: Same answer.

lugi0 (approved):

sheltoncyril (approved):
LGTM — solid test expansion. Leaving a few non-blocking comments for consideration.

**Main observations:**

1. **Duplicate `AGENT_CATALOG_API_PATH`** — defined in both `tests/ai_hub/agent_catalog/constants.py` and `tests/ai_hub/constants.py`. Pick one location.

2. **Missing `pytestmark`** on `TestAgentCatalogFiltering`, `TestAgentCatalogPagination`, and `TestAgentCatalogSearch` — no tier marker or `usefixtures`. Compare to `TestAgentCatalogFilterOptions` which has both. If these classes need `agent_catalog_configmap_patch`, they should declare it.

3. **`assert_paginated_agents_unique_and_filtered` assumes integer IDs** — `int(agent_id)` will crash if IDs are non-numeric. Consider string comparison or guarding.

4. **`all_agents` vs `agents_response`** — near-identical fixtures in `search/conftest.py` and `agent_catalog/conftest.py`. Could unify.

5. **`test_filter_valid_source_label`** — asserts `filtered == unfiltered` which assumes all agents share the default source label. If `agent_catalog_configmap_patch` runs first (from another class), test agents from "Test Agents" source would break this assertion. Fragile ordering dependency.

6. **`test_keyword_search`** with `criteria=None` — no assertion that results are non-empty. If "ReAct" matches nothing, empty == empty passes silently.

Details in inline comments.

tests/ai_hub/constants.py
sheltoncyril, line 80: nit: This constant is also defined in `tests/ai_hub/agent_catalog/constants.py` (line 1). Duplicate — pick one location to avoid drift.

tests/ai_hub/agent_catalog/search/test_agent_catalog_filtering.py
```diff
@@ -0,0 +1,170 @@
+from typing import Any, Self
```
sheltoncyril, line 1: Missing `pytestmark` — no tier marker or `usefixtures`. Same applies to `TestAgentCatalogPagination` and `TestAgentCatalogSearch`. Compare to `TestAgentCatalogFilterOptions` which declares `tier1` + `usefixtures`.

tests/ai_hub/agent_catalog/utils.py
```diff
@@ -0,0 +1,306 @@
...
+
+        agent_id = item["id"]
+        assert agent_id not in seen_ids, (
+            f"Page {page_idx} returned duplicate agent id '{agent_id}', already seen on a previous page"
+        )
+
+        sort_order = order_params.get("sortOrder", "ASC")
+        if seen_ids:
+            if sort_order == "ASC":
+                assert int(agent_id) > int(seen_ids[-1]), (
```
sheltoncyril, line 297: `int(agent_id)` assumes IDs are numeric. If agent IDs are UUIDs or non-numeric strings, this will raise `ValueError`. Consider string comparison or adding a guard.

tests/ai_hub/agent_catalog/search/test_agent_catalog_filtering.py
```diff
@@ -0,0 +1,170 @@
...
+    def test_filter_valid_source_label(
+        self: Self,
+        agent_catalog_rest_urls: list[str],
+        model_registry_rest_headers: dict[str, str],
+    ) -> None:
+        """Given agents are configured with a default source label
+        When filtering by a valid sourceLabel
+        Then the same agents are returned as without the filter
+        """
+        unfiltered = execute_get_command_with_retry(
```
sheltoncyril, line 139: Asserts `filtered == unfiltered`, which assumes all agents share `DEFAULT_AGENT_SOURCE_LABEL`. If `agent_catalog_configmap_patch` from another test class runs first and injects "Test Agents" source, those agents won't match this filter — assertion breaks. Consider making fixture dependency explicit or scoping the comparison.

tests/ai_hub/agent_catalog/search/conftest.py
```diff
@@ -0,0 +1,21 @@
+from typing import Any
+
+import pytest
+
+from tests.ai_hub.utils import execute_get_command_with_retry
+
+
+@pytest.fixture(scope="class")
+def all_agents(
+    agent_catalog_rest_urls: list[str],
```
sheltoncyril, line 10: Near-duplicate of `agents_response` fixture in `agent_catalog/conftest.py` (same URL, same `pageSize: 1000`). Difference: this returns `items` list, the other returns full dict. Could unify to reduce maintenance.

dbasunag: With other prs, will handle in a future pr.

tests/ai_hub/agent_catalog/search/test_agent_catalog_filtering.py
dbasunag: Will check in a future pr and address if needed.

tests/ai_hub/agent_catalog/utils.py
dbasunag: Id's are numeric so this comment is not right.

tests/ai_hub/agent_catalog/search/test_agent_catalog_filtering.py
dbasunag: tier2 markers are default for ai hub and use fixtures are not needed for non-tier1 tests. Hence this is fine.

tests/ai_hub/constants.py
dbasunag: Will remove duplicate in a follow up pr.

tests/ai_hub/agent_catalog/search/test_agent_catalog_filtering.py
dbasunag: This is actually not a problem, since agent_catalog_configmap_patch is not used in search files. So no action is needed here.

tests/ai_hub/agent_catalog/search/conftest.py
dbasunag: They have different purpose. One includes custom, the second only works with the default. So both are needed.

PR #1955, opened 2026-07-08
test: add default agent catalog related tests

dbasunag:
# Pull Request

## Related Issues

- Fixes: 
- JIRA: 

- [ ] Locally
- [ ] Jenkins

## Additional Requirements

## Summary by CodeRabbit

* **Tests**
  * Added new coverage for default agent catalog configuration, including expected ConfigMap contents and agent label definitions.
  * Introduced reusable fixtures to load default catalog sources from YAML, derive catalogs/label definitions, and access the user-editable catalog ConfigMap.
  * Added validation that the catalog deployment includes the expected volume mount and `--catalogs-path` argument.
  * Extended REST API checks to list agents and verify the `assetType=agents` sources response (including `available` status).

### DIFF
--- a/tests/ai_hub/agent_catalog/config/conftest.py
+++ b/tests/ai_hub/agent_catalog/config/conftest.py
@@ -1,0 +1,74 @@
+import pytest
+import yaml
+from kubernetes.dynamic import DynamicClient
+from ocp_resources.config_map import ConfigMap
+from ocp_resources.deployment import Deployment
+
+from tests.ai_hub.agent_catalog.config.constants import AGENT_CATALOG_SOURCES_CM
+from tests.ai_hub.constants import CATALOG_CONTAINER, DEFAULT_MODEL_CATALOG_CM
+from tests.ai_hub.model_catalog.constants import MODEL_CATALOG_DEPLOYMENT_NAME
+
+
+@pytest.fixture(scope="class")
+def default_agent_catalog_sources_data(
+    admin_client: DynamicClient,
+    model_registry_namespace: str,
+) -> dict:
+    """Return the parsed sources.yaml data from the default catalog sources ConfigMap."""
+    configmap = ConfigMap(
+        name=DEFAULT_MODEL_CATALOG_CM,
+        client=admin_client,
+        namespace=model_registry_namespace,
+    )
+    return yaml.safe_load(configmap.instance.data.get("sources.yaml", "{}") or "{}")
+
+
+@pytest.fixture(scope="class")
+def default_agent_catalogs(default_agent_catalog_sources_data: dict) -> list[dict]:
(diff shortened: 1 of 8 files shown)


tests/ai_hub/agent_catalog/config/conftest.py
```diff
@@ -0,0 +1,74 @@
...
+        client=admin_client,
+        namespace=model_registry_namespace,
+    )
+
+
+@pytest.fixture(scope="class")
+def catalog_container_spec(
+    admin_client: DynamicClient,
+    model_registry_namespace: str,
+) -> object:
```
lugi0: object is not really useful as a type annotation, this would be a container (or we could use Any)

tests/ai_hub/agent_catalog/config/constants.py
```diff
@@ -0,0 +1,23 @@
+AGENT_CATALOG_SOURCES_CM: str = "agent-catalog-sources"
```
lugi0, line 1: you already defined constants in #1942 , would it make sense to have a single constants file for the agent catalog tests? this one is in a different location and defines different variables.

tests/ai_hub/agent_catalog/config/test_default_agent_catalog.py
```diff
@@ -0,0 +1,155 @@
...
+    def test_default_agent_source_available(
+        self: Self,
+        model_catalog_routes: list,
+        model_registry_rest_headers: dict[str, str],
+    ) -> None:
+        """Given the default agent catalog source is configured
+        When requesting GET /sources?assetType=agents
+        Then rh_agents source appears with status available
+        """
+        url = f"https://{model_catalog_routes[0].instance.spec.host}:443/api/model_catalog/v1alpha1/sources"
```
lugi0, line 140: perhaps it would make sense to move the url value to a variable in constants rather than hardcode it here? or you could reuse the fixture defined in #1942 (model_catalog_api_url)

tests/ai_hub/agent_catalog/config/constants.py
dbasunag: I would need to reorganize once both the prs are merged. This one is focused towards only the default catalog. Whichever gets merged first, would require the second one to be rebased and updated.

tests/ai_hub/agent_catalog/config/test_default_agent_catalog.py
dbasunag: It can't be a constant, but depending on whichever gets merged first. the second one would be reusing the fixtures.

lugi0 (approved):

SB159 (approved):
/lgtm

PR #1996, opened 2026-07-14
fix: use operator default network policy for OGX server

derekhiggins:
## Summary

- Remove custom network policy override from `create_ogx_server` and `_create_ogx_server`
- The custom policy used `kubernetes.io/metadata.name: openshift-ingress` which fails on clusters where the HAProxy router runs with `hostNetwork: true` - traffic appears from the node IP, not the pod namespace
- The operator default policy uses `network.openshift.io/policy-group: ingress`, which matches both `openshift-ingress` and `openshift-host-network`, covering both pod-networked and host-networked router configurations

## Test plan

- [ ] Deploy OGX server and verify route is accessible on a cluster with `hostNetwork: true` router
- [ ] Verify operator default network policy is applied

## Summary by CodeRabbit

* **Tests**
  * Updated OGX server test setup to use the standard/default networking configuration.
  * Removed custom ingress network access settings that previously scoped traffic to a specific ingress namespace and port.
  * Improved alignment between test deployments and the default server behavior to reduce environment-specific discrepancies.

### DIFF
--- a/tests/ogx/utils.py
+++ b/tests/ogx/utils.py
@@ -149,35 +149,12 @@ def create_ogx_server(
         teardown: Whether to delete the resource on exit.
     """
 
-    # Starting with RHOAI 3.3, pods in the 'openshift-ingress' namespace must be allowed
-    # to access the ogx-service. This is required for the ogx_test_route
-    # to function properly.
-    network: dict[str, Any] = {
-        "policy": {
-            "ingress": [
-                {
-                    "from": [
-                        {
-                            "namespaceSelector": {
-                                "matchLabels": {
-                                    "kubernetes.io/metadata.name": "openshift-ingress",
-                                },
-                            },
-                        },
-                    ],
-                    "ports": [{"protocol": "TCP", "port": 8321}],
-                },
-            ],
-        },
-    }
-
     with OgxServer(
         client=client,
         name=name,
         namespace=namespace,
         distribution=config["distribution"],
         workload=config.get("workload"),
-        network=network,
         tls=config.get("tls"),
         wait_for_resource=True,
         teardown=teardown,
--- a/tests/pipelines_components/autorag/conftest.py
+++ b/tests/pipelines_components/autorag/conftest.py
@@ -113,31 +113,12 @@ def _create_ogx_server(
     namespace: str,
     config: dict[str, Any],
 ) -> Generator[OgxServer, Any, Any]:
-    network: dict[str, Any] = {
-        "policy": {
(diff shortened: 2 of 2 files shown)


tests/ogx/utils.py
```diff
@@ -149,35 +149,12 @@ def create_ogx_server(
         teardown: Whether to delete the resource on exit.
     """
 
-    # Starting with RHOAI 3.3, pods in the 'openshift-ingress' namespace must be allowed
```
jgarciao, line 152: >The operator default policy uses network.openshift.io/policy-group: ingress, which matches both openshift-ingress and openshift-host-network, covering both pod-networked and host-networked router configurations

Can you verify this sentence is true in a cluster? If not, you could add an additional rule matching the openshift-host-network label

This worked for me with RHOAI 3.5 EA2 and the Entra cluster:

```
    # Starting with RHOAI 3.3, pods in the 'openshift-ingress' namespace must be allowed
    # to access the ogx-service. This is required for the ogx_test_route
    # to function properly.
    network: dict[str, Any] = {
        "policy": {
            "ingress": [
                {
                    "from": [
                        {
                            "namespaceSelector": {
                                "matchLabels": {
                                    "kubernetes.io/metadata.name": "openshift-ingress",
                                },
                            },
                        },
                    ],
                    "ports": [{"protocol": "TCP", "port": 8321}],
                },
                {
                    "from": [
                        {
                            "namespaceSelector": {
                                "matchLabels": {
                                    "kubernetes.io/metadata.name": "openshift-host-network",
                                },
                            },
                        },
                    ],
                    "ports": [{"protocol": "TCP", "port": 8321}],
                },
            ],
        },
    }
```

derekhiggins: > > The operator default policy uses network.openshift.io/policy-group: ingress, which matches both openshift-ingress and openshift-host-network, covering both pod-networked and host-networked router configurations
> 
> Can you verify this sentence is true in a cluster? If not, you could add an additional rule matching the openshift-host-network label

I've confirmed that the default networkpolicy works on both  pod-networked and host-networked router configurations  (I don't think its anything to do with  Entra clusters)

> This worked for me with RHOAI 3.5 EA2 and the Entra cluster:

AIUI what you used would work on  pod-networked routers

jgarciao (approved):
LGTM. 

Can you backport it to the 3.5ea2 branch? You can do it  adding a comment 
`/cherry-pick 3.5ea2` once the PR is merged

jiripetrlik (approved):

PR #2000, opened 2026-07-14
test: add KServe canary rollout P0 E2E tests (RHAISTRAT-1250)

mholder6:
## Summary

Adds initial P0 downstream E2E tests for KServe canary rollout in RawDeployment mode, from test plan [RHAISTRAT-1250](https://issues.redhat.com/browse/RHAISTRAT-1250).

New suite: `tests/model_serving/model_server/kserve/canary_rollout/`

Covers four P0 cases:
- **TC-CRD-001** — `canary` array field accepted on InferenceService (`trafficPercent` + named canary predictor)
- **TC-CTRL-001** — kserve-controller creates stable + canary Deployments
- **TC-ROUTE-001** — OpenShift Route `alternateBackends` 90/10 split, verified with live traffic
- **TC-E2E-001** — full canary lifecycle: split traffic → promote → single Deployment / cleared alternateBackends

Traffic split is checked by status-code fingerprinting (not pod metrics):
- stable: `gs://kfserving-examples/models/sklearn/1.0/model` → HTTP 200
- canary: `gs://kfserving-examples/models/sklearn/1.3/mixedtype` → HTTP 500
- same iris rows sent as V2 `/v2/models/{isvc}/infer` (MLServer on cluster does not expose V1 `:predict`)

No S3 credentials required. Gateway API HTTPRoute coverage is intentionally out of scope (upstream-only per the test plan).

Happy to extend this if there are other canary scenarios the suite should cover — this is the initial P0 set from the plan.

## Related Issues

- JIRA: [RHAISTRAT-1250](https://issues.redhat.com/browse/RHAISTRAT-1250)

- [x] Locally
- [ ] Jenkins

**Note:** Full suite run needs a cluster with canary-capable KServe. GitHub PR CI runs `tox` only.

Example local run (canary controller / skip flags as needed for pre-GA builds):
```bash
uv run pytest tests/model_serving/model_server/kserve/canary_rollout/ -v \
  --cluster-sanity-skip-check --skip-kserve-health-check \
  --tc=use_unprivileged_client:False
```

## Additional Requirements

- [x] If this PR introduces new marker(s)/adds a new component, was relevant ticket created to update relevant Jenkins job? _(Not yet — follow-up needed to wire suite into QE Jenkins tier jobs)_

[RHAISTRAT-1250]: https://redhat.atlassian.net/browse/RHAISTRAT-1250?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RHAISTRAT-1250]: https://redhat.atlassian.net/browse/RHAISTRAT-1250?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Tests**
  * Expanded KServe canary rollout coverage with new fixtures and helpers to create canary InferenceServices, wait for canary readiness/Route admission, and validate controller-created Deployments.
  * Added dedicated tests for CRD canary field persistence, controller deployment behavior, OpenShift Route `alternateBackends` traffic splitting, and an end-to-end canary promotion flow (including post-promotion traffic verification and cleanup).

### DIFF
--- a/tests/model_serving/model_server/kserve/canary_rollout/conftest.py
+++ b/tests/model_serving/model_server/kserve/canary_rollout/conftest.py
@@ -1,0 +1,157 @@
+"""Fixtures for KServe canary rollout (RawDeployment) tests."""
+
+from collections.abc import Generator
+from typing import Any
+
+import pytest
+from kubernetes.dynamic import DynamicClient
+from ocp_resources.inference_service import InferenceService
+from ocp_resources.namespace import Namespace
+from ocp_resources.secret import Secret
+from ocp_resources.service_account import ServiceAccount
+from ocp_resources.serving_runtime import ServingRuntime
+
+from tests.model_serving.model_server.kserve.canary_rollout.constants import (
+    CANARY_FEATURE_NAME,
+    CANARY_MODEL_DIR,
+    CANARY_MODEL_FORMAT,
+    CANARY_NAMESPACE_PREFIX,
+    DEFAULT_CANARY_TRAFFIC_PERCENT,
+    DEFAULT_DEPLOYMENT_MODE,
+    STABLE_MODEL_DIR,
+    STABLE_MODEL_FORMAT,
+)
+from tests.model_serving.model_server.kserve.canary_rollout.utils import create_canary_inference_service
+from utilities.constants import RuntimeTemplates
+from utilities.infra import create_ns, s3_endpoint_secret
+from utilities.serving_runtime import ServingRuntimeFromTemplate
+
+pytestmark = [pytest.mark.rawdeployment, pytest.mark.tier1, pytest.mark.usefixtures("valid_aws_config")]
+
+
+@pytest.fixture(scope="package")
+def canary_rollout_namespace(
+    admin_client: DynamicClient,
+    unprivileged_client: DynamicClient,
+) -> Generator[Namespace, Any, Any]:
+    """Shared namespace for canary rollout tests."""
+    with create_ns(
+        admin_client=admin_client,
+        unprivileged_client=unprivileged_client,
+        name=f"{CANARY_NAMESPACE_PREFIX}-ns",
+    ) as namespace:
+        yield namespace
+
+
+@pytest.fixture(scope="package")
+def canary_rollout_s3_secret(
+    unprivileged_client: DynamicClient,
+    canary_rollout_namespace: Namespace,
+    aws_access_key_id: str,
+    aws_secret_access_key: str,
+    ci_s3_bucket_name: str,
+    ci_s3_bucket_region: str,
+    ci_s3_bucket_endpoint: str,
+) -> Generator[Secret, Any, Any]:
+    """S3 credentials secret for canary rollout model storage."""
+    with s3_endpoint_secret(
+        client=unprivileged_client,
+        name="canary-models-bucket-secret",
+        namespace=canary_rollout_namespace.name,
+        aws_access_key=aws_access_key_id,
+        aws_secret_access_key=aws_secret_access_key,
+        aws_s3_region=ci_s3_bucket_region,
+        aws_s3_bucket=ci_s3_bucket_name,
+        aws_s3_endpoint=ci_s3_bucket_endpoint,
+    ) as secret:
+        yield secret
+
+
+@pytest.fixture(scope="package")
+def canary_rollout_service_account(
+    admin_client: DynamicClient,
+    canary_rollout_namespace: Namespace,
+    canary_rollout_s3_secret: Secret,
+) -> Generator[ServiceAccount, Any, Any]:
+    """Service account linked to the canary rollout S3 secret."""
+    with ServiceAccount(
+        client=admin_client,
+        namespace=canary_rollout_namespace.name,
+        name="canary-models-bucket-sa",
+        secrets=[{"name": canary_rollout_s3_secret.name}],
+    ) as service_account:
+        yield service_account
+
+
+@pytest.fixture(scope="package")
+def canary_mlserver_runtime(
+    admin_client: DynamicClient,
+    canary_rollout_namespace: Namespace,
+    mlserver_runtime_image: str,
+) -> Generator[ServingRuntime, Any, Any]:
+    """MLServer ServingRuntime for RawDeployment canary tests."""
+    with ServingRuntimeFromTemplate(
+        client=admin_client,
+        name=f"{CANARY_FEATURE_NAME}-runtime",
+        namespace=canary_rollout_namespace.name,
+        template_name=RuntimeTemplates.MLSERVER,
+        deployment_type=DEFAULT_DEPLOYMENT_MODE,
+        runtime_image=mlserver_runtime_image,
+    ) as model_runtime:
+        yield model_runtime
+
+
+def _model_storage_uri(bucket_name: str, model_dir: str) -> str:
+    return f"s3://{bucket_name}/{model_dir.rstrip('/')}/"
+
+
+@pytest.fixture(scope="package")
+def canary_sklearn_inference_service(
+    admin_client: DynamicClient,
+    canary_rollout_namespace: Namespace,
+    canary_mlserver_runtime: ServingRuntime,
+    ci_s3_bucket_name: str,
+    canary_rollout_service_account: ServiceAccount,
+) -> Generator[InferenceService, Any, Any]:
+    """InferenceService with a canary array entry at 10% traffic."""
+    with create_canary_inference_service(
+        client=admin_client,
+        name=f"{CANARY_NAMESPACE_PREFIX}-10",
+        namespace=canary_rollout_namespace.name,
+        runtime=canary_mlserver_runtime.name,
+        stable_model_format=STABLE_MODEL_FORMAT,
+        stable_storage_uri=_model_storage_uri(ci_s3_bucket_name, STABLE_MODEL_DIR),
+        canary_model_format=CANARY_MODEL_FORMAT,
+        canary_storage_uri=_model_storage_uri(ci_s3_bucket_name, CANARY_MODEL_DIR),
+        canary_traffic_percent=DEFAULT_CANARY_TRAFFIC_PERCENT,
+        deployment_mode=DEFAULT_DEPLOYMENT_MODE,
+        external_route=True,
+        model_service_account=canary_rollout_service_account.name,
+    ) as isvc:
+        yield isvc
+
+
+@pytest.fixture(scope="package")
+def canary_ctrl_inference_service(
+    admin_client: DynamicClient,
+    canary_rollout_namespace: Namespace,
+    canary_mlserver_runtime: ServingRuntime,
+    ci_s3_bucket_name: str,
+    canary_rollout_service_account: ServiceAccount,
+) -> Generator[InferenceService, Any, Any]:
+    """InferenceService with 20% canary traffic for controller behavior tests."""
+    with create_canary_inference_service(
+        client=admin_client,
+        name=f"{CANARY_NAMESPACE_PREFIX}-ctrl",
+        namespace=canary_rollout_namespace.name,
+        runtime=canary_mlserver_runtime.name,
+        stable_model_format=STABLE_MODEL_FORMAT,
+        stable_storage_uri=_model_storage_uri(ci_s3_bucket_name, STABLE_MODEL_DIR),
+        canary_model_format=CANARY_MODEL_FORMAT,
+        canary_storage_uri=_model_storage_uri(ci_s3_bucket_name, CANARY_MODEL_DIR),
+        canary_traffic_percent=20,
+        deployment_mode=DEFAULT_DEPLOYMENT_MODE,
+        external_route=True,
+        model_service_account=canary_rollout_service_account.name,
+    ) as isvc:
+        yield isvc
--- a/tests/model_serving/model_server/kserve/canary_rollout/conftest.py
+++ b/tests/model_serving/model_server/kserve/canary_rollout/conftest.py
@@ -7,26 +7,28 @@ import pytest
 from kubernetes.dynamic import DynamicClient
 from ocp_resources.inference_service import InferenceService
 from ocp_resources.namespace import Namespace
-from ocp_resources.secret import Secret
-from ocp_resources.service_account import ServiceAccount
 from ocp_resources.serving_runtime import ServingRuntime
 
+from tests.model_serving.model_runtime.mlserver.probes.utils import (
+    MLSERVER_LIVENESS_PROBE,
+    MLSERVER_READINESS_PROBE,
+)
 from tests.model_serving.model_server.kserve.canary_rollout.constants import (
     CANARY_FEATURE_NAME,
-    CANARY_MODEL_DIR,
     CANARY_MODEL_FORMAT,
     CANARY_NAMESPACE_PREFIX,
+    CANARY_STORAGE_URI,
     DEFAULT_CANARY_TRAFFIC_PERCENT,
     DEFAULT_DEPLOYMENT_MODE,
-    STABLE_MODEL_DIR,
     STABLE_MODEL_FORMAT,
+    STABLE_STORAGE_URI,
 )
 from tests.model_serving.model_server.kserve.canary_rollout.utils import create_canary_inference_service
-from utilities.constants import RuntimeTemplates
-from utilities.infra import create_ns, s3_endpoint_secret
+from utilities.constants import Containers, RuntimeTemplates
+from utilities.infra import create_ns
 from utilities.serving_runtime import ServingRuntimeFromTemplate
 
-pytestmark = [pytest.mark.rawdeployment, pytest.mark.tier1, pytest.mark.usefixtures("valid_aws_config")]
+pytestmark = [pytest.mark.rawdeployment, pytest.mark.tier1]
 
 
 @pytest.fixture(scope="package")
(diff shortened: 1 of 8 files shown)


VedantMahabaleshwarkar (commented):
Superseded — see inline review below.

tests/model_serving/model_server/kserve/canary_rollout/test_e2e_kserve_canary_rollout.py
```diff
@@ -0,0 +1,77 @@
...
+            client=admin_client,
+            isvc=canary_e2e_inference_service,
+            runtime_name=runtime_name,
+            expected_count=1,
+        )
+        assert len(post_promotion_deployments) == 1
+
+        route = get_model_route(client=admin_client, isvc=canary_e2e_inference_service)
+        alternate_backends = route.instance.spec.get("alternateBackends") or []
+        assert not alternate_backends, "Expected Route alternateBackends to be cleared after promotion"
```
VedantMahabaleshwarkar, line 78: After promotion we check infra state but never confirm traffic actually hits the promoted model. Worth adding a post-promotion inference check — e.g. all 500s from mixedtype, and maybe wait for Ready again.

tests/model_serving/model_server/kserve/canary_rollout/utils.py
```diff
@@ -0,0 +1,347 @@
...
+        patches={
+            isvc: {
+                "spec": {
+                    "predictor": promoted_predictor,
+                    "canary": [],
+                },
+            },
+        },
+    ).update()
+    wait_for_inference_deployment_replicas(
```
VedantMahabaleshwarkar, line 424: This only waits for deployment count — Route `alternateBackends` can lag. A polled wait here (like `wait_for_route_admitted`) would help, and `PROMOTION_WAIT_TIMEOUT` looks like it was meant for this.

tests/model_serving/model_server/kserve/canary_rollout/utils.py
```diff
@@ -0,0 +1,347 @@
...
+def wait_for_canary_ready_condition(isvc: InferenceService, timeout: int = Timeout.TIMEOUT_15MIN) -> None:
+    """Wait until a canary-related Ready condition is True, if present."""
+
+    def _canary_ready() -> bool:
+        conditions = isvc.instance.status.get("conditions") or []
+        for condition in conditions:
+            condition_type = condition.get("type", "")
+            if "canary" in condition_type.lower() and condition.get("status") == "True":
+                return True
+        return isvc.instance.status.get("ready", False) is True
```
VedantMahabaleshwarkar: If a canary condition exists but isn't True, we fall through to generic `status.ready` which can mask a not-yet-ready canary. Fixture already waits for READY + 2 deployments + route admission — I'd fix this logic or drop the helper.

tests/model_serving/model_server/kserve/canary_rollout/utils.py
```diff
@@ -0,0 +1,347 @@
...
+    host = inference.get_inference_url()
+    predict_model = model_name or isvc.name
+    rest_url = f"https://{host}{V2_INFER_PATH_TEMPLATE.format(model_name=predict_model)}"
+
+    stable_hits = 0
+    canary_hits = 0
+    other_statuses: dict[int, int] = {}
+
+    for _ in range(sample_size):
+        response = requests.post(url=rest_url, json=TRAFFIC_INFERENCE_INPUT, verify=False, timeout=60)
```
VedantMahabaleshwarkar: New TCP+TLS per request here — `requests.Session` would be faster (see `maas_billing`). Also any single 502/503/504 fails the whole run; a short warmup + bounded retry for transient 5xx would help. (`verify=False` is fine, matches rest of repo.)

tests/model_serving/model_server/kserve/canary_rollout/utils.py
```diff
@@ -0,0 +1,347 @@
...
+    stable_weight: int,
+    canary_weight: int,
+) -> None:
+    """Assert OpenShift Route primary and alternate backend weights."""
+    route = get_model_route(client=isvc.client, isvc=isvc)
+    route_spec = route.instance.spec
+    primary_weight = route_spec["to"]["weight"]
+    alternate_backends = route_spec.get("alternateBackends") or []
+
+    assert primary_weight == stable_weight, f"Expected stable Route weight {stable_weight}, got {primary_weight}"
```
VedantMahabaleshwarkar: We check weights but not that `to.name` / `alternateBackends[0].name` point at the right Services. Worth asserting `{isvc}-predictor` vs `{isvc}-canary-predictor` so we don't pass on correct weights with wrong backends.

tests/model_serving/model_server/kserve/canary_rollout/constants.py
```diff
@@ -0,0 +1,39 @@
...
+            "name": "predict",
+            "shape": [len(_TRAFFIC_ROWS), len(_TRAFFIC_ROWS[0])],
+            "datatype": "FP32",
+            "data": _TRAFFIC_ROWS,
+        }
+    ]
+}
+V2_INFER_PATH_TEMPLATE: str = "/v2/models/{model_name}/infer"
+
+PROMOTION_WAIT_TIMEOUT: int = 120
```
VedantMahabaleshwarkar, line 42: Defined but never used — wire into post-promotion Route wait or drop it.

tests/model_serving/model_server/kserve/canary_rollout/utils.py
```diff
@@ -0,0 +1,347 @@
...
+    runtime: str,
+    model_format: str,
+    timeout: int = Timeout.TIMEOUT_15MIN,
+) -> None:
+    """Promote canary by replacing stable storage and clearing the canary array.
+
+    Uses a permanent patch (not a ResourceEditor context manager) so promotion is
+    not undone when the helper returns.
+    """
+    promoted_predictor = build_predictor_spec(
```
VedantMahabaleshwarkar, line 408: `create_canary_inference_service` supports `model_service_account` but promotion rebuilds the predictor without it. Nothing uses it today, but worth threading through to avoid a silent break later.

tests/model_serving/model_server/kserve/canary_rollout/conftest.py
```diff
@@ -0,0 +1,147 @@
+"""Fixtures for KServe canary rollout (RawDeployment) tests."""
```
VedantMahabaleshwarkar, line 1: Minor consistency — `model_cache/` has an `__init__.py` and this directory doesn't. Package-scoped fixtures work fine without it, but an empty one would align with that sibling.

tests/model_serving/model_server/kserve/canary_rollout/conftest.py
```diff
@@ -0,0 +1,147 @@
...
+    DEFAULT_DEPLOYMENT_MODE,
+    STABLE_MODEL_FORMAT,
+    STABLE_STORAGE_URI,
+)
+from tests.model_serving.model_server.kserve.canary_rollout.utils import create_canary_inference_service
+from utilities.constants import Containers, RuntimeTemplates
+from utilities.infra import create_ns
+from utilities.serving_runtime import ServingRuntimeFromTemplate
+
+pytestmark = [pytest.mark.rawdeployment, pytest.mark.tier1]
```
VedantMahabaleshwarkar, line 31: ROUTE test fires 1000 HTTP requests — might be worth `@pytest.mark.slow` so CI can filter if runtime becomes an issue.

VedantMahabaleshwarkar (commented):
Nice work on this — the status-code fingerprinting approach is clever and the fixture scoping is well thought out. Left a few inline comments, mostly around E2E post-promotion coverage and hardening the traffic sampling loop.

tests/model_serving/model_server/kserve/canary_rollout/test_e2e_kserve_canary_rollout.py
mholder6: Addressed in 97752cf3 — after promotion we wait for Ready, then assert all sampled requests return 500 (promoted mixedtype) with `expected_percent=100`.

tests/model_serving/model_server/kserve/canary_rollout/utils.py
mholder6: Addressed — `promote_canary_to_stable` now waits for Ready and polls until Route `alternateBackends` are cleared via `wait_for_route_alternate_backends_cleared` using `PROMOTION_WAIT_TIMEOUT`.

tests/model_serving/model_server/kserve/canary_rollout/utils.py
mholder6: Fixed — the helper no longer falls through to generic Ready. If canary conditions exist, we require one to be True; if none exist yet, we keep waiting.

tests/model_serving/model_server/kserve/canary_rollout/utils.py
mholder6: Addressed — sampling now uses `requests.Session`, discards a short warmup, and retries 502/503/504 a few times before classifying the response.

tests/model_serving/model_server/kserve/canary_rollout/utils.py
mholder6: Addressed — we now assert primary `{isvc}-predictor` and alternate `{isvc}-canary-predictor` in addition to the weights.

tests/model_serving/model_server/kserve/canary_rollout/constants.py
mholder6: Wired in — `PROMOTION_WAIT_TIMEOUT` is the default for `wait_for_route_alternate_backends_cleared` after promotion.

tests/model_serving/model_server/kserve/canary_rollout/utils.py
mholder6: Addressed — `promote_canary_to_stable` accepts `model_service_account` and defaults to preserving the existing predictor `serviceAccountName` when unset.

tests/model_serving/model_server/kserve/canary_rollout/conftest.py
mholder6: Added an empty `__init__.py` for consistency with `model_cache/`.

tests/model_serving/model_server/kserve/canary_rollout/conftest.py
mholder6: Added `@pytest.mark.slow` on the ROUTE test module (1000-sample traffic check).

tests/model_serving/model_server/kserve/canary_rollout/utils.py
```diff
@@ -0,0 +1,426 @@
...
+
+
+def wait_for_canary_ready_condition(isvc: InferenceService, timeout: int = Timeout.TIMEOUT_15MIN) -> None:
+    """Wait until a canary-related Ready condition is True.
+
+    Does not fall back to generic Ready — a present-but-False canary condition must stay False.
+    """
+
+    def _canary_ready() -> bool:
+        conditions = isvc.instance.status.get("conditions") or []
```
mwaykole: `isvc.instance` is a cached snapshot from the last API interaction — it's not a live property. `TimeoutSampler` will keep calling `_canary_ready()` but read the same stale conditions every iteration, so this always times out unless the canary condition happened to be True already at fixture yield time.

We hit the same pattern in `negative/utils.py` where `_model_status()` needed `isvc.update()` before reading `status.modelStatus`. Add a refresh here:

```python
def _canary_ready() -> bool:
    isvc.api.get()  # or isvc.update() depending on ocp_resources version
    conditions = isvc.instance.status.get("conditions") or []
    ...
```

Right now you get away with it because `create_canary_inference_service` already calls `wait_for_condition(READY)` before yielding, so `isvc.instance` happens to have a fresh snapshot where canary is True. But if the fixture ever drops that pre-wait, or if this helper is reused for a different flow, it'll silently regress to a 15-minute timeout.

tests/model_serving/model_server/kserve/canary_rollout/utils.py
```diff
@@ -0,0 +1,426 @@
...
+            model_service_account=model_service_account,
+            name=predictor_name,
+        ),
+        "trafficPercent": canary_traffic_percent,
+    }
+
+
+def deployment_contains_storage_uri(deployment: Deployment, storage_uri: str) -> bool:
+    """Return True if the deployment pod spec references the storage URI."""
+    deployment_json = json.dumps(deployment.instance.to_dict())
```
mwaykole: This serializes the entire Deployment (including metadata, status, all labels, annotations, etc.) to JSON just to do a substring search. Works, but it's heavyweight and could false-positive on a match in an unrelated field (e.g. an annotation containing the storage fragment).

More targeted:
```python
def deployment_contains_storage_uri(deployment: Deployment, storage_uri: str) -> bool:
    for container in deployment.instance.spec.template.spec.get("containers", []):
        for env in container.get("env", []):
            if env.get("value") and storage_uri in env["value"]:
                return True
    for init_container in deployment.instance.spec.template.spec.get("initContainers", []):
        for env in init_container.get("env", []):
            if env.get("value") and storage_uri in env["value"]:
                return True
    return False
```

Not a blocker — the current version is correct for these tests.

tests/model_serving/model_server/kserve/canary_rollout/test_ctrl_kserve_canary_rollout.py
```diff
@@ -0,0 +1,46 @@
...
+            expected_count=2,
+        )
+        assert len(deployments) == 2
+
+        canary_storage_fragment = CANARY_STORAGE_URI.rstrip("/").split("/")[-1]
+        assert any(
+            deployment_contains_storage_uri(deployment, canary_storage_fragment) for deployment in deployments
+        ), f"Expected one Deployment to reference canary storage ({canary_storage_fragment})"
+
+        running_pods = [
```
mwaykole: Nit: `running_pods` is a list of `Deployment` objects, not pods. Something like `deployments_with_ready_pods` would be clearer. The assertion message also says "canary Deployment pod" but this checks both stable and canary deployments.

tests/model_serving/model_server/kserve/canary_rollout/test_e2e_kserve_canary_rollout.py
```diff
@@ -0,0 +1,86 @@
...
+
+        # Promoted stable is mixedtype — fingerprint status is 500 for every request.
+        assert_canary_traffic_by_status_codes(
+            isvc=canary_e2e_inference_service,
+            expected_percent=100,
+            sample_size=50,
```
mwaykole, line 85: This is the right call to validate post-promotion, but `expected_percent=100` with `tolerance_percent=0` on 50 samples means a single 200 response from a terminating old pod fails the run.

`promote_canary_to_stable` already waits for single deployment + Ready + Route cleared, so stale responses *shouldn't* happen. But on slow CI clusters where pod termination grace period overlaps with Route convergence, you could see a straggler. Something to keep an eye on if this flakes in nightly — bumping to `tolerance_percent=2` (i.e. allow 1 out of 50) would be a safe fallback.

tests/model_serving/model_server/kserve/canary_rollout/utils.py
```diff
@@ -0,0 +1,426 @@
...
+    with (
+        InferenceService(
+            client=client,
+            name=name,
+            namespace=namespace,
+            annotations=annotations,
+            label=labels,
+            predictor=predictor,
+            teardown=teardown,
+        ) as isvc,
```
mwaykole, line 363: There's an inherent race between `InferenceService.__enter__()` (creates the ISVC with stable-only predictor) and `ResourceEditor.__enter__()` (patches in the canary array). If the controller reconciles between those two calls, it briefly sees a non-canary ISVC.

In practice the controller takes >1s for first reconciliation so the patch wins the race. But this is worth a comment at the call site — future readers might wonder why canary isn't in the initial spec.

tests/model_serving/model_server/kserve/canary_rollout/constants.py
```diff
@@ -0,0 +1,39 @@
...
+
+CANARY_FEATURE_NAME: str = "kserve-canary-rollout"
+CANARY_NAMESPACE_PREFIX: str = "kserve-canary"
+
+# Both revisions are sklearn; canary mixedtype returns HTTP 500 for TRAFFIC_INFERENCE_INPUT.
+STABLE_MODEL_FORMAT: str = ModelFormat.SKLEARN
+CANARY_MODEL_FORMAT: str = ModelFormat.SKLEARN
+
+STABLE_STORAGE_URI: str = "gs://kfserving-examples/models/sklearn/1.0/model"
+CANARY_STORAGE_URI: str = "gs://kfserving-examples/models/sklearn/1.3/mixedtype"
```
mwaykole, line 16: The test correctness depends on this bucket reliably returning 500 for the traffic payload. If Google ever updates or removes `sklearn/1.3/mixedtype`, all four tests break silently (they'd see unexpected status codes rather than the 200/500 fingerprint).

Might be worth a brief doc comment noting this dependency, or long-term moving to a self-hosted model artifact.

mwaykole (commented):
Nice initial P0 set — the status-code fingerprinting approach is a clever way to validate traffic split without needing pod-level metrics, and the fixture scoping (package for shared ISVCs, function for E2E promotion) is well thought out. Good that Vedant's feedback on Session/warmup/retries, Route backend name assertions, and serviceAccountName preservation got addressed in the latest push.

One real issue: `wait_for_canary_ready_condition` polls a stale `isvc.instance` — it works today only because the fixture pre-waits for Ready, but the helper itself is broken as a standalone utility. The rest are nits/things to watch in CI.

LGTM with the `isvc` refresh fix.

tests/model_serving/model_server/kserve/canary_rollout/utils.py
mholder6: Addressed in 79832315 — `_canary_ready` now does a live `isvc.api.get(...)` each poll instead of reading a held snapshot.

tests/model_serving/model_server/kserve/canary_rollout/utils.py
mholder6: Addressed — now only scans container/initContainer env values for the storage URI fragment.

tests/model_serving/model_server/kserve/canary_rollout/test_ctrl_kserve_canary_rollout.py
mholder6: Renamed to `deployments_with_ready_pods` and clarified the assertion message.

tests/model_serving/model_server/kserve/canary_rollout/test_e2e_kserve_canary_rollout.py
mholder6: Bumped to `tolerance_percent=2` so a single straggler 200 during pod termination won't fail the run.

tests/model_serving/model_server/kserve/canary_rollout/utils.py
mholder6: Added a call-site comment explaining why canary is patched in after create and noting the brief reconcile race.

tests/model_serving/model_server/kserve/canary_rollout/constants.py
mholder6: Documented the public GCS fingerprint dependency on the storage URI constants, including the note to prefer self-hosted artifacts long-term.

VedantMahabaleshwarkar (approved):
LGTM

dbasunag (approved):

dbasunag:
Merging based on @VedantMahabaleshwarkar 's request as it is P0 test pre for Kserve and FF is today.

PR #2008, opened 2026-07-15
Enhance KServe upgrade tests with Kueue integration

sukumars321:
# Pull Request

## Summary

- Update pytest fixture logic to conditionally insert "enabled_kserve_in_dsc" based on KServe deployment type and upgrade status.
- Introduce new configuration for KServe raw deployment with Kueue, including resource specifications and expected pod states.
- Implement pre/post-upgrade tests for InferenceService with Kueue admission control, verifying deployment readiness and scaling behavior.
- Add utility functions for managing DSCI service mesh state during upgrades.
- Ensure proper handling of inference timeouts in utility functions.

This commit improves the robustness of upgrade tests and integrates Kueue functionality into the KServe deployment process.

## Related Issues

Fixes:
JIRA: https://redhat.atlassian.net/browse/RHOAIENG-63118
This code change addresses Task 5 of [[RHOAIENG-63117](https://redhat.atlassian.net/browse/RHOAIENG-63117?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ) Kueue Upgrade Test Coverage Audit Report](https://docs.google.com/document/d/1ON66oE8MuKhHsT_YaG5hxTdAjhcHbYOvd5pskj9uPyg/edit?usp=sharing)

- [x] Locally
- [ ] Jenkins

## Additional Requirements

[RHOAIENG-63117]: https://redhat.atlassian.net/browse/RHOAIENG-63117?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -445,12 +445,6 @@ def enabled_modelmesh_in_dsc(
 def enabled_kserve_in_dsc(
     dsc_resource: DataScienceCluster,
 ) -> Generator[DataScienceCluster, Any, Any]:
-    # Set by ensure_kserve_enabled_for_upgrade so upgrade lanes own DSC/DSCI KServe setup
-    # and this fixture does not conflict (or restore) that state on teardown.
-    if py_config.get("kserve_enablement_handled_by_upgrade"):
-        yield dsc_resource
-        return
-
     with update_components_in_dsc(
         dsc=dsc_resource,
         components={DscComponents.KSERVE: DscComponents.ManagementState.MANAGED},
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -445,6 +445,12 @@ def enabled_modelmesh_in_dsc(
 def enabled_kserve_in_dsc(
     dsc_resource: DataScienceCluster,
 ) -> Generator[DataScienceCluster, Any, Any]:
+    # Set by ensure_kserve_enabled_for_upgrade so upgrade lanes own DSC/DSCI KServe setup
+    # and this fixture does not conflict (or restore) that state on teardown.
+    if py_config.get("kserve_enablement_handled_by_upgrade"):
+        yield dsc_resource
+        return
+
     with update_components_in_dsc(
         dsc=dsc_resource,
         components={DscComponents.KSERVE: DscComponents.ManagementState.MANAGED},
--- a/tests/model_serving/model_server/upgrade/conftest.py
+++ b/tests/model_serving/model_server/upgrade/conftest.py
@@ -1,7 +1,11 @@
 from typing import Any, Generator
 
 import pytest
+from _pytest.config import Config
+from _pytest.nodes import Item
+from _pytest.runner import CallInfo
 from kubernetes.dynamic import DynamicClient
+from kubernetes.dynamic.exceptions import ResourceNotFoundError
 from ocp_resources.inference_service import InferenceService
 from ocp_resources.namespace import Namespace
 from ocp_resources.role import Role
@@ -24,8 +28,46 @@ from utilities.inference_utils import create_isvc
 from utilities.infra import create_inference_token, create_isvc_view_role, create_ns, s3_endpoint_secret
 from utilities.serving_runtime import ServingRuntimeFromTemplate
 
+from pytest_testconfig import config as py_config
+
+from tests.model_serving.model_runtime.vllm.utils import skip_if_not_deployment_mode
+from tests.model_serving.model_server.upgrade.kserve_kueue_upgrade_config import (
+    KSERVE_KUEUE_CLUSTER_QUEUE,
+    KSERVE_KUEUE_CPU_QUOTA,
+    KSERVE_KUEUE_ISVC_LABELS,
+    KSERVE_KUEUE_ISVC_RESOURCES,
+    KSERVE_KUEUE_LOCAL_QUEUE,
+    KSERVE_KUEUE_MAX_REPLICAS,
+    KSERVE_KUEUE_MEMORY_QUOTA,
+    KSERVE_KUEUE_MIN_REPLICAS,
+    KSERVE_KUEUE_RESOURCE_FLAVOR,
+    KSERVE_KUEUE_UPGRADE_ISVC_NAME,
+    KSERVE_KUEUE_UPGRADE_NAMESPACE,
+    KSERVE_KUEUE_UPGRADE_RUNTIME_NAME,
+    KSERVE_KUEUE_UPGRADE_S3_SECRET,
+)
+from tests.model_serving.model_server.upgrade.utils import (
+    _capture_and_save_isvc_kueue_baseline,
+    _create_kueue_upgrade_resources,
+    _kserve_kueue_upgrade_runtime_template_kwargs,
+)
+from utilities.kueue_utils import LocalQueue
+
 
 LOGGER = get_logger(name=__name__)
+
+
+@pytest.hookimpl(tryfirst=True, hookwrapper=True)
+def pytest_runtest_makereport(item: Item, call: CallInfo[None]) -> Generator[None, Any, Any]:
+    """Track pre-upgrade test failures to prevent baseline capture on failure."""
+    outcome = yield
+    report = outcome.get_result()
+
+    # Only track failures during the actual test execution (not setup/teardown)
+    if call.when == "call" and report.failed:
+        if "pre_upgrade" in item.keywords:
+            # Mark that a pre-upgrade test failed so baseline capture is skipped
+            item.config._pre_upgrade_test_failed = True  # type: ignore[attr-defined]
 
 
 @pytest.fixture(scope="session")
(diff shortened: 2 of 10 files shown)


conftest.py
```diff
@@ -373,7 +373,8 @@ def pytest_runtest_setup(item: Item) -> None:
         item.fixturenames.insert(0, "enabled_kserve_in_dsc")
 
     elif KServeDeploymentType.RAW_DEPLOYMENT.lower() in item.keywords:
-        item.fixturenames.insert(0, "enabled_kserve_in_dsc")
+        if not ("kueue" in item.keywords and ("pre_upgrade" in item.keywords or "post_upgrade" in item.keywords)):
```
mwaykole: Same concern as PR #1888: this hardcodes knowledge of which fixture combinations handle KServe enablement into the root `pytest_runtest_setup`. If someone adds a new `kueue + raw_deployment` upgrade test without knowing about this carve-out, `enabled_kserve_in_dsc` silently won't inject and the test will misbehave.

The upgrade fixture should handle this conflict internally rather than modifying the global hook.

tests/model_serving/model_server/upgrade/utils.py
```diff
@@ -47,7 +294,7 @@ def verify_inference_generation(isvc: InferenceService, expected_generation: int
         ResourceMismatch: If inference generation is not equal to expected generation
     """
     if isvc.instance.status.observedGeneration != expected_generation:
-        ResourceMismatchError(f"Inference service {isvc.name} was modified")
+        raise ResourceMismatchError(f"Inference service {isvc.name} was modified")
```
mwaykole, line 110: Good catch fixing the missing `raise` -- this was a real bug where `verify_inference_generation` silently succeeded even when the generation didn't match. However, this is an unrelated bugfix that affects all upgrade tests, not just the new Kueue lane. It should be a separate commit (or even a separate small PR) so it can be reviewed and cherry-picked independently.

tests/model_serving/model_server/upgrade/utils.py
```diff
@@ -35,6 +61,227 @@ def verify_pod_containers_not_restarted(client: DynamicClient, component_name: s
...
+    """
+    return {
+        "managementState": DscComponents.ManagementState.REMOVED,
+        "controlPlane": {
+            "name": KSERVE_DSCI_SERVICEMESH_CONTROL_PLANE_NAME,
+            "namespace": KSERVE_DSCI_SERVICEMESH_CONTROL_PLANE_NAMESPACE,
+            "metricsCollection": "Istio",
+        },
+        "auth": {"audiences": [KSERVE_DSCI_SERVICEMESH_AUTH_AUDIENCE]},
+    }
```
mwaykole: This uses `Headed` while the `main` branch PR (#1888) uses `Headless`. Which is correct for 2.25?

On 2.25 the ISVC is created with `external_route=True` and inference goes through the external HTTPS route. On main it's `external_route=False` and uses port-forward. The service config needs to match -- `Headed` creates a ClusterIP Service (required for routes/ingress), `Headless` creates a headless Service. If you're using external routes, `Headed` is the right choice here.

But this means the two PRs are testing fundamentally different things -- one tests port-forward inference, the other tests external route inference. Is that intentional, or was main supposed to be `Headed` + `external_route=True` too?

tests/model_serving/model_server/upgrade/conftest.py
```diff
@@ -24,10 +25,98 @@
...
 
+from ocp_resources.cluster_service_version import ClusterServiceVersion
+from ocp_resources.config_map import ConfigMap
+from ocp_resources.data_science_cluster import DataScienceCluster
+from ocp_resources.dsc_initialization import DSCInitialization
+from pytest_testconfig import config as py_config
+
+from tests.model_serving.model_runtime.vllm.utils import skip_if_not_deployment_mode
+from tests.model_serving.model_server.upgrade.kserve_kueue_upgrade_config import (
+    KSERVE_KUEUE_CLUSTER_QUEUE,
```
mwaykole, line 32: `_is_kueue_operator_installed`, `kueue_resource_groups`, `_create_kueue_upgrade_resources`, `_capture_and_save_isvc_kueue_baseline`, `_ensure_kserve_enabled_for_upgrade`, `_ensure_kueue_available_for_upgrade`, `_restore_kueue_dsc_state` — these are all standalone helper functions, not fixtures.

`conftest.py` should only contain fixtures (and the hooks that pytest discovers). Move all `_`-prefixed helpers to `utils.py`.

tests/model_serving/model_server/upgrade/conftest.py
```diff
@@ -495,3 +584,545 @@ def ovms_authenticated_serverless_inference_service_scope_session(
...
+    """S3 connection secret for the KServe Kueue upgrade ISVC."""
+    _ = valid_aws_config
+    secret_kwargs = {
+        "client": admin_client,
+        "name": KSERVE_KUEUE_UPGRADE_S3_SECRET,
+        "namespace": kserve_kueue_upgrade_namespace.name,
+    }
+    secret = Secret(**secret_kwargs)
+
+    if pytestconfig.option.post_upgrade:
```
mwaykole, line 556: This reads `dsc_resource.instance.spec.components[DscComponents.KUEUE].managementState` without guarding against `None`.

If the `kueue` component doesn't exist in the DSC spec (e.g., on an older RHOAI version that doesn't have the Kueue component), this will `KeyError` or `AttributeError` instead of giving a clear failure message.

Suggest:
```python
kueue_spec = dsc_resource.instance.spec.components.get(DscComponents.KUEUE)
if not kueue_spec:
    pytest.fail("Kueue component not found in DSC spec. Upgrade lanes require Kueue DSC integration.")
kueue_management_state = kueue_spec.managementState
```

tests/model_serving/model_server/upgrade/test_upgrade_kserve_kueue_raw.py
```diff
@@ -0,0 +1,371 @@
...
+        isvc = kserve_kueue_upgrade_inference_service
+        LOGGER.info(f"[PRE-UPGRADE] Checking ISVC '{isvc.name}' exists in namespace '{isvc.namespace}'")
+        assert isvc.exists, f"InferenceService {isvc.name} does not exist"
+        LOGGER.info(f"[PRE-UPGRADE] PASS: ISVC '{isvc.name}' is deployed")
+
+    @pytest.mark.pre_upgrade
+    def test_initial_deployment_ready(
+        self,
+        admin_client: DynamicClient,
+        kserve_kueue_upgrade_inference_service: InferenceService,
```
mwaykole, line 87: This does a full-object update with `isvc.update(isvc_dict)` which sends back the entire ISVC spec+status. This is a known race condition vector -- if the controller reconciles between `isvc.instance.to_dict()` and `isvc.update()`, you'll get a 409 Conflict.

Use a targeted patch instead:
```python
isvc.update(resource_dict={
    "metadata": {"name": isvc.name},
    "spec": {"predictor": {"minReplicas": KSERVE_KUEUE_SCALED_REPLICAS}},
})
```

tests/model_serving/model_server/upgrade/test_upgrade_kserve_kueue_raw.py
```diff
@@ -0,0 +1,371 @@
...
+        kserve_kueue_upgrade_serving_runtime: ServingRuntime,
+    ) -> None:
+        """Test steps:
+
+        1. Locate the ISVC Deployment.
+        2. Wait for the initial replica to be deployed.
+        """
+        isvc = kserve_kueue_upgrade_inference_service
```
mwaykole, line 95: The baseline capture happens in the fixture teardown (after `yield`), but the `test_kueue_scale_and_gate` test mutates the ISVC replicas during the test body. This means the baseline captures the *scaled-up* state (2 replicas), not the initial state.

Is that intentional? Post-upgrade tests assert `kueue_integration_stats` and `total_copies` match the baseline -- so they're effectively asserting against the scaled state. That's probably correct for this use-case (verifying the gated state survives upgrade), but it would be worth a comment in the fixture to make that clear.

Also: if `test_kueue_scale_and_gate` fails mid-scale, the baseline may capture an inconsistent state.

utilities/inference_utils.py
```diff
@@ -429,6 +433,29 @@ def run_inference(
             ValueError: If inference fails
 
         """
+        wait_timeout = inference_timeout or Timeout.TIMEOUT_30SEC
```
mwaykole: `inference_timeout or Timeout.TIMEOUT_30SEC` treats `inference_timeout=0` as falsy, falling back to 30s. If someone passes `inference_timeout=0` (meaning "no retry"), they'd get 30s instead.

Use:
```python
wait_timeout = inference_timeout if inference_timeout is not None else Timeout.TIMEOUT_30SEC
```

tests/model_serving/model_server/upgrade/utils.py
```diff
@@ -62,4 +309,368 @@ def verify_serving_runtime_generation(isvc: InferenceService, expected_generatio
...
+        AssertionError: If ConfigMap does not exist or has no baseline data
+    """
+    cm = ConfigMap(
+        client=client,
+        name=cm_name,
+        namespace=namespace,
+    )
+
+    if not cm.exists:
+        raise AssertionError(
```
mwaykole, line 326: The `except Exception` here swallows all non-409 errors silently and re-raises. But the check `if "409" in str(exc) or "Conflict" in str(exc)` is string-matching against the exception message which is fragile.

The `kubernetes` library raises `kubernetes.client.exceptions.ApiException` with a `status` attribute. Use:
```python
except ApiException as exc:
    if exc.status == 409:
        last_conflict = exc
        continue
    raise
```

Also, the typo on line 569: `AssertionError` should be `AssertionError` (it happens to be valid Python since it's just a string argument, but it's a typo in the error message context).

tests/model_serving/model_server/upgrade/utils.py
```diff
@@ -62,4 +309,368 @@ def verify_serving_runtime_generation(isvc: InferenceService, expected_generatio
...
+                f"Container set changed after upgrade for pod {pod.name}: "
+                f"missing containers {sorted(missing_containers)}"
+            )
+
+        for container in statuses:
+            if container.name not in pod_baseline:
+                raise PodContainersRestartError(
+                    f"Container set changed after upgrade for pod {pod.name}: new container '{container.name}'"
+                )
+            pre_count = pod_baseline[container.name]
```
mwaykole, line 405: The two `raise PodContainersRestartError(...)` messages for `missing_pods` (line 588) and `new_pods` (line 591) are identical strings but represent different failure modes. The `missing_pods` case is pods that disappeared; the `new_pods` case is unexpected extra pods.

Consider distinct messages:
```python
if missing_pods:
    raise PodContainersRestartError(
        f"Baseline pods missing after upgrade for {isvc.name}: {sorted(missing_pods)}"
    )
if new_pods and not allow_new_pods:
    raise PodContainersRestartError(
        f"Unexpected new pods appeared after upgrade for {isvc.name}: {sorted(new_pods)}"
    )
```

uv.lock
```diff
@@ -1,5 +1,5 @@
 version = 1
-revision = 3
+revision = 2
 requires-python = "==3.13.*"
```
mwaykole, line 4: This changes `revision` from 3 to 2 -- a downgrade. `uv.lock` revision tracks the lockfile schema version and is set by the `uv` resolver. A downgrade usually means this was generated with a different (older) version of `uv` than what the 2.25 branch currently uses.

This could break lock resolution for anyone using the newer `uv` version. Was this intentional? If not, regenerate with the same `uv` version the branch currently uses.

tests/model_serving/model_server/upgrade/utils.py
```diff
@@ -62,4 +309,368 @@ def verify_serving_runtime_generation(isvc: InferenceService, expected_generatio
...
+            f"ISVC '{inference_service.name}' is not exposed (networking.kserve.io/visibility={visibility!r}). "
+            "Set external_route=True when creating the ISVC and re-run pre-upgrade."
+        )
+
+    wait_for_isvc_inference_url(isvc=inference_service)
+
+    verify_inference_response(
+        inference_service=inference_service,
+        inference_config=inference_config,
+        inference_type=inference_type,
```
mwaykole, line 481: Unreachable code after `pytest.fail` (which raises). This `return last_url` only satisfies mypy, but `pytest.fail` already returns `NoReturn`. A cleaner pattern:
```python
    except TimeoutExpiredError:
        pytest.fail(
            f"Timeout waiting for inference URL on ISVC '{isvc.name}'..."
        )
    # pytest.fail raises; TimeoutSampler exhaustion is the only path here
    raise AssertionError(f"ISVC '{isvc.name}' has no status.url")
```

tests/model_serving/model_server/upgrade/test_upgrade_kserve_kueue_raw.py
```diff
@@ -0,0 +1,371 @@
...
+            )
+
+        pod_labels = [
+            create_isvc_label_selector_str(
+                isvc=isvc,
+                resource_type="pod",
+                runtime_name=runtime_name,
+            )
+        ]
+        running_pods = 0
```
mwaykole, line 168: `baseline` fixture return type is `dict` -- too vague for what's actually an `ISVCKueueBaseline`. The post-upgrade tests rely on specific keys (`kueue_integration_stats`, `total_copies`, `isvc_observed_generation`, `pod_restart_counts`) which won't be caught by type checkers with a bare `dict`.

```python
def baseline(self, ...) -> ISVCKueueBaseline:
```

mwaykole (changes requested):
Thanks for the 2.25 backport. This is a larger PR than the `main` counterpart (#1888) because it also adds missing utilities (`get_dsc_ready_condition`, `wait_for_dsc_reconciliation`, `wait_for_kueue_crds_available`, `DscComponents.KUEUE`, etc.) that already exist on `main`. That's expected, but it makes the diff harder to review.

### Blocking issues

1. **Fragile root `conftest.py` hook** -- Same issue as #1888. The `pytest_runtest_setup` carve-out is implicit and fragile. The upgrade fixture should handle the KServe enablement conflict internally.

2. **Full-object ISVC update** -- `isvc.update(isvc_dict)` sends the entire object back including status, which is a 409 Conflict waiting to happen. Use a targeted patch.

3. **`uv.lock` revision downgrade** (3 -> 2) -- Likely generated with an older `uv` version. This can break lock resolution. Please regenerate with the correct `uv` version.

4. **Helpers in `conftest.py`** -- ~8 standalone functions in conftest. Move to `utils.py`.

### Important issues

5. **`Headed` vs `Headless` discrepancy** -- This PR uses `rawDeploymentServiceConfig: Headed` + `external_route=True`, while #1888 on main uses `Headless` + `external_route=False`. These test completely different inference paths. Is that intentional?

6. **`inference_timeout or` treats 0 as falsy** -- Use `inference_timeout if inference_timeout is not None else ...`.

7. **Unrelated `raise` bugfix** mixed in -- The fix for `verify_inference_generation` and `verify_serving_runtime_generation` is correct and needed, but should be a separate commit for clean backport tracking.

8. **Fragile 409 handling** -- String-matching `"409" in str(exc)` instead of checking `ApiException.status`.

### Minor

9. Duplicate error messages for missing vs new pods (different failures, same message)
10. `baseline` fixture return type should be `ISVCKueueBaseline`, not `dict`
11. Unreachable code after `pytest.fail` in `wait_for_isvc_inference_url`
12. Unguarded `dsc_resource.instance.spec.components[DscComponents.KUEUE]` -- will `KeyError` on DSCs without Kueue

mwaykole:
fix tox

tests/conftest.py
```diff
@@ -445,6 +445,12 @@ def enabled_modelmesh_in_dsc(
 def enabled_kserve_in_dsc(
     dsc_resource: DataScienceCluster,
 ) -> Generator[DataScienceCluster, Any, Any]:
+    # Set by ensure_kserve_enabled_for_upgrade so upgrade lanes own DSC/DSCI KServe setup
+    # and this fixture does not conflict (or restore) that state on teardown.
+    if py_config.get("kserve_enablement_handled_by_upgrade"):
```
mwaykole: This guard reads `py_config.get("kserve_enablement_handled_by_upgrade")` but nowhere in this PR (or the upgrade conftest/fixtures) is this key ever set to `True`.

As-is, this is dead code — the early-return never fires, and `enabled_kserve_in_dsc` will still modify the DSC in upgrade lanes, causing teardown conflicts with whatever the upgrade fixture does to KServe.

Either:
1. Show the fixture/hook that sets `py_config["kserve_enablement_handled_by_upgrade"] = True` before this fixture runs, or
2. Remove this guard if upgrade lanes handle DSC state a different way.

utilities/constants.py
```diff
@@ -159,10 +159,12 @@ class DscComponents:
     KSERVE: str = "kserve"
     MODELREGISTRY: str = "modelregistry"
     LLAMASTACKOPERATOR: str = "llamastackoperator"
+    KUEUE: str = "kueue"
```
mwaykole, line 162: `KUEUE` is added here but is **not** added to `COMPONENT_MAPPING` below.

On 2.25, `COMPONENT_MAPPING` maps component names to their DSC condition types. If any code path calls `update_components_in_dsc(components={DscComponents.KUEUE: ...})` with `wait_for_components_state=True` (the default), it will hit:
```
KeyError: 'kueue'
```
at `DscComponents.COMPONENT_MAPPING[component]`.

Either add the mapping:
```python
COMPONENT_MAPPING: dict[str, str] = {
    ...
    KUEUE: ConditionType.KUEUE_READY,  # add the actual condition name
}
```
or ensure all Kueue `update_components_in_dsc` calls pass `wait_for_components_state=False`.

tests/model_serving/model_server/upgrade/utils.py
```diff
@@ -62,4 +122,483 @@ def verify_serving_runtime_generation(isvc: InferenceService, expected_generatio
...
     runtime = get_inference_serving_runtime(isvc=isvc)
     if runtime.instance.metadata.generation != expected_generation:
-        ResourceMismatchError(f"Serving runtime {runtime.name} was modified")
+        raise ResourceMismatchError(f"Serving runtime {runtime.name} was modified")
+
+
+def read_isvc_total_copies(isvc: InferenceService) -> int:
+    """Return totalCopies from ISVC status without requiring a fully Loaded model state.
+
+    Use for Kueue gating scenarios where additional replicas are pending admission and
```
mwaykole, line 131: Stale object read — `isvc.instance` is the cached representation from when the object was last fetched. After `test_kueue_scale_and_gate` patches `minReplicas` and waits for pod gating, the ISVC status (including `modelStatus.copies.totalCopies`) has been updated by the controller, but this code reads from the **old** cached object.

Without refreshing, this will either:
- Return stale pre-scale data and pass incorrectly, or
- Fail with `model_status.copies is None` because the old object didn't have copies populated yet.

Add a refresh before reading:
```python
def read_isvc_total_copies(isvc: InferenceService) -> int:
    isvc.reload()  # or the equivalent refresh call
    model_status = isvc.instance.status.modelStatus
    ...
```

tests/model_serving/model_server/upgrade/test_upgrade_kserve_kueue_raw.py
```diff
@@ -0,0 +1,371 @@
...
+        verify_kserve_kueue_upgrade_inference(
+            inference_service=kserve_kueue_upgrade_inference_service,
+            inference_config=OPENVINO_KSERVE_INFERENCE_CONFIG,
+            inference_type=Inference.INFER,
+            inference_timeout=KSERVE_KUEUE_INFERENCE_TIMEOUT,
+        )
+
+
+def _get_deployment_status_replicas(deployment: Deployment) -> int:
+    return deployment.instance.status.replicas
```
mwaykole: Two problems:

1. **No reload**: `deployment.instance.status.replicas` reads the cached object. Without `deployment.reload()`, the `TimeoutSampler` in `test_kueue_scale_and_gate` polls the same stale value every 5s and either times out or returns immediately depending on when the object was first populated.

2. **Null safety**: `status.replicas` is `None` during initial rollout (before the Deployment controller sets it). This returns `None`, and the sampler comparison `None == KSERVE_KUEUE_SCALED_REPLICAS` will never be `True`, causing a timeout with a confusing error message.

Fix:
```python
def _get_deployment_status_replicas(deployment: Deployment) -> int:
    deployment.reload()
    return deployment.instance.status.replicas or 0
```

mwaykole (commented):
Inline comments on the three blocking issues:

1. `py_config["kserve_enablement_handled_by_upgrade"]` is never set — the fixture guard is dead code.
2. `DscComponents.KUEUE` missing from `COMPONENT_MAPPING` — runtime KeyError if `update_components_in_dsc` is called for kueue with wait.
3. Stale object reads in `read_isvc_total_copies` and `_get_deployment_status_replicas` — will cause flakes or false passes in CI.

Please address these before merge.

tests/model_serving/model_server/upgrade/utils.py
```diff
@@ -62,4 +122,483 @@ def verify_serving_runtime_generation(isvc: InferenceService, expected_generatio
...
+    cluster_queue_name: str,
+    resource_flavor_name: str,
+    cpu_quota: int,
+    memory_quota: str,
+    teardown_resources: bool,
+) -> Generator[LocalQueue, Any, Any]:
+    """Create or look up Kueue resources for upgrade tests."""
+    if pytestconfig.option.post_upgrade:
+        local_queue = LocalQueue(
+            client=admin_client,
```
mwaykole, line 505: Tested on cluster (RHOAI 2.25.9) — `test_inference_post_upgrade` fails here because `insecure=True` is not passed to `verify_inference_response`.

The docstring above says:
> External routes use curl with `--insecure` on managed clusters instead

But the call doesn't actually do that. Without `insecure=True`, `get_ca_bundle()` returns the istio/knative CA cert (`istio_knative.crt`) which is irrelevant for a raw deployment OpenShift Route — curl gets an empty response or TLS mismatch.

Fix:
```python
verify_inference_response(
    inference_service=inference_service,
    inference_config=inference_config,
    inference_type=inference_type,
    protocol=Protocols.HTTPS,
    use_default_query=True,
    insecure=True,  # ← add this
    inference_timeout=inference_timeout,
)
```

Confirmed: all other post-upgrade tests pass (6/7), only this inference call fails due to wrong CA.

mwaykole (commented):
Ran pre+post upgrade on RHOAI 2.25.9 cluster. Pre-upgrade 3/3 pass, post-upgrade 6/7 pass. The one failure (`test_inference_post_upgrade`) is a missing `insecure=True` in `verify_kserve_kueue_upgrade_inference` — see inline comment.

tests/model_serving/model_server/upgrade/utils.py
```diff
@@ -62,4 +122,483 @@ def verify_serving_runtime_generation(isvc: InferenceService, expected_generatio
...
+                ),
+                create_cluster_queue(
+                    client=admin_client,
+                    name=cluster_queue_name,
+                    resource_groups=kueue_resource_groups(
+                        flavor_name=resource_flavor_name,
+                        cpu_quota=cpu_quota,
+                        memory_quota=memory_quota,
+                    ),
+                    teardown=teardown_resources,
```
mwaykole: This checks for a stale `LocalQueue` (namespaced) before pre-upgrade, but does **not** check for stale `ClusterQueue` or `ResourceFlavor` (cluster-scoped). 

Confirmed on cluster: when pre-upgrade is re-run after a partial failure or without `--delete-pre-upgrade-resources`, the namespace gets cleaned up but `upgrade-kserve-cluster-queue` and `upgrade-kserve-flavor` survive (they're cluster-scoped). The next run hits `ConflictError` on `ResourceFlavor` creation.

Add guards for both cluster-scoped resources:
```python
cluster_queue = ClusterQueue(client=admin_client, name=cluster_queue_name)
resource_flavor = ResourceFlavor(client=admin_client, name=resource_flavor_name)
stale = []
if local_queue.exists:
    stale.append(f"LocalQueue '{local_queue_name}'")
if cluster_queue.exists:
    stale.append(f"ClusterQueue '{cluster_queue_name}'")
if resource_flavor.exists:
    stale.append(f"ResourceFlavor '{resource_flavor_name}'")
if stale:
    pytest.fail(
        f"Stale upgrade resources found: {', '.join(stale)}. "
        "Clear them before pre-upgrade or use --delete-pre-upgrade-resources."
    )
```

mwaykole (commented):
One additional blocking issue found during cluster testing: cluster-scoped Kueue resources (`ClusterQueue`, `ResourceFlavor`) are not guarded against stale leftovers on pre-upgrade re-runs, causing `ConflictError`. See inline comment.

utilities/constants.py
sukumars321: Added Kueue to COMPONENT_MAPPING

tests/model_serving/model_server/upgrade/test_upgrade_kserve_kueue_raw.py
sukumars321: Addressed but used the repo convention of .get() call to refresh the object.

mwaykole (approved):

PR #2011, opened 2026-07-15
fix(ogx): label odh-trusted-ca-bundle for operator cache on disconnected clusters

derekhiggins:
## Summary

- On disconnected clusters with `HTTPS_PROXY`, the OGX operator cache filters ConfigMaps by `ogx.io/watch=true`, so `odh-trusted-ca-bundle` must be labeled before `OGXServer` creation or the operator cannot find the CA bundle and HTTPS through the proxy fails

## Test plan

- Run OGX tests against a disconnected cluster with `HTTPS_PROXY` set

## Summary by CodeRabbit

* **Tests**
  * Improved test environment setup for disconnected Kubernetes clusters using HTTPS proxies.
  * Added validation for trusted certificate configuration and automatic readiness handling.
  * Enhanced coverage of resource labeling and monitoring behavior in air-gapped environments.


tests/ogx/conftest.py
```diff
@@ -145,6 +147,15 @@ def ogx_server(
...
 
+    if is_disconnected_cluster and HTTPS_PROXY:
+        cm = ConfigMap(
+            client=unprivileged_client,
+            name="odh-trusted-ca-bundle",
+            namespace=unprivileged_model_namespace.name,
+        )
+        cm.wait(timeout=5)
+        ResourceEditor(patches={cm: {"metadata": {"labels": {"ogx.io/watch": "true"}}}}).update()
+
```
dbasunag, line 158: Do you need to undo this or wait for stabilization once this change is propagated?

tests/ogx/conftest.py
```diff
@@ -145,6 +147,15 @@ def ogx_server(
         ogx_srv.clean_up()
         return
 
+    if is_disconnected_cluster and HTTPS_PROXY:
```
jgarciao, line 150: This is needed anytime the tls field is set when creating the OGXServer. This is more future proof I think
```suggestion
    if ogx_server_config.get("tls"):
```

Also, beware this code won't be run when pytestconfig.option.post_upgrade. Not sure if it's needed in that phase

tests/ogx/conftest.py
derekhiggins: No need to undo, the configmap is cleaned up after the test, I'm not sure about waiting for stabilization, I didn't need to wait when testing locally but maybe I just got lucky..

dbasunag: May be @Artemon-line can chime in here?

tests/ogx/conftest.py
derekhiggins: done

tests/ogx/conftest.py
```diff
@@ -145,6 +147,15 @@ def ogx_server(
...
+        cm = ConfigMap(
+            client=unprivileged_client,
+            name="odh-trusted-ca-bundle",
+            namespace=unprivileged_model_namespace.name,
+        )
+        cm.wait(timeout=5)
+        ResourceEditor(patches={cm: {"metadata": {"labels": {"ogx.io/watch": "true"}}}}).update()
```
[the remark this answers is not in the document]
derekhiggins: cleaned up during teardown unless teardown_resources=False the label is preserved for post-test inspection.

Artemon-line (approved):
LGTM!

dbasunag (approved):

PR #2012, opened 2026-07-15
Fix guardrails retry logic to handle HTTP ReadTimeout

Sandeep20013:
# Pull Request

## Summary

The `@retry` decorator on guardrails test functions only retries on Python's built-in `TimeoutError`, but when an HTTP request exceeds the 30s timeout, `requests` raises `ReadTimeout`. This causes the test to fail immediately instead of retrying within the 120s budget.

Fixed by catching `ReadTimeout` in `_send_guardrails_orchestrator_post_request` and re-raising it as `TimeoutError` so the existing retry logic handles it.

image


sheltoncyril (dismissed):
/lgtm

tests/ai_safety/guardrails/utils.py
```diff
@@ -267,13 +267,16 @@ def _send_guardrails_orchestrator_post_request(
...
-    )
+    try:
+        response = requests.post(
+            url=url,
+            headers=get_auth_headers(token=token),
+            json=payload,
+            verify=ca_bundle_file,
+            timeout=30,
+        )
+    except requests.exceptions.ReadTimeout as e:
```
dbasunag: This is not the right approach. Where you are using the retry decorator, in that exception list you need to add requests.exceptions.ReadTimeout

sheltoncyril (approved):
/lgtm

dbasunag (approved):

PR #2020, opened 2026-07-16
fix(trustyai): prevent replica race condition in LMEval fixture

swattamw2024:
## Summary
  - Fix race condition in `patched_dsc_lmeval_allow_all` fixture where `trustyai_operator_deployment.replicas` returns `None` when read during operator
  reconciliation
  - Default to `1` when replicas is `None` to avoid setting `replicas: null` on the deployment
  - Add `wait_for_replicas()` after scale-down to ensure the scale-down completes before scaling back up
  
  ## Problem
  When the fixture patches the DSC to enable LMEval flags, it reads the current replica count, scales to 0, then scales back. On slow/busy post-upgrade clusters
  (e.g., 2.25.0 → 2.25.9), the replica read returns `None` (operator is mid-reconciliation). Setting `replicas: None` removes the field from the deployment spec,
  relying on the operator to reconcile it back — which may not happen within the 240s timeout.
  
  This leaves TrustyAI at 0 replicas → `TrustyAIReady: False` → `DSC Ready: False` → cascading failure across all test suites that use the
  `cluster_sanity_scope_session` fixture (42+ tests in model-server, workbenches, etc.).
  
  ## Test plan
  - [x] Verified on broken cluster (aws-up1-2259): `oc scale --replicas=1` restored TrustyAI and DSC Ready status
  - [ ] Run `pytest tests/ai_safety/lm_eval/upgrade/ -k test_lmeval_job_pod_lifecycle` on a 2.25.x cluster 
  - [ ] Verify log shows `Set deployment replicas: 1` (not `None`)
  - [ ] Verify TrustyAI deployment is healthy after test: `replicas=1 ready=1`
  - [ ] Verify DSC `Ready: True` after test completes

**Jira:** https://redhat.atlassian.net/browse/RHOAIENG-77602


tests/fixtures/trustyai.py
```diff
@@ -46,8 +46,5 @@ def patched_dsc_lmeval_allow_all(
             }
         }
     ):
-        num_replicas: int = trustyai_operator_deployment.replicas
```
dbasunag, line 49: Your problem seems to be 
```
(e.g., 2.25.0 → 2.25.9), the replica read returns None (operator is mid-reconciliation). Setting replicas: None removes the field from the deployment spec,
```
The solution would be to ensure operator is not `mid-reconciliation`.  May be collect `num_replicas: int = trustyai_operator_deployment.replicas` before the patch?

dbasunag (changes requested):

tests/fixtures/trustyai.py
swattamw2024: Updated. Moved num_replicas read before the ResourceEditor block so the operator is stable when we read it — returns 1 instead of None. Added or 1 as safety
  fallback. Scale-down/up logic preserved as-is.

sheltoncyril (approved):
/lgtm but I think the fix is in the operator too which we're looking at

kpunwatk (approved):

tests/fixtures/trustyai.py
dbasunag: Thank you!

dbasunag (approved):

PR #2027, opened 2026-07-17
fix(ogx): set network.externalAccess.enabled and pass network config to OgxServer

derekhiggins:
## Summary

- Set `network.externalAccess.enabled: true` in the OgxServer spec built by `build_ogx_server_config`
- Pass the `network` key through `create_ogx_server` to the `OgxServer` constructor (it was being silently dropped)

Without `externalAccess` enabled, the operator-managed NetworkPolicy does not include the OpenShift router namespace (`network.openshift.io/policy-group: ingress`), so the public route returns 503.

## Test plan

- [ ] Run `tests/ogx/vector_io/` smoke tests and verify the OgxServer route is reachable
- [ ] Confirm operator-generated NetworkPolicy includes the ingress namespace selector

## Summary by CodeRabbit

* **New Features**
  * Enabled external network access in OGX server configurations.
  * Added support for passing optional network settings when creating OGX servers.

### DIFF
--- a/tests/ogx/server_config.py
+++ b/tests/ogx/server_config.py
@@ -144,6 +144,7 @@ def build_ogx_server_config(
 
     config: dict[str, Any] = {
         "distribution": {"name": "rh"},
+        "network": {"externalAccess": {"enabled": True}},
         "workload": {
             "resources": {
                 "requests": {"cpu": cpu_requests, "memory": "1Gi"},
--- a/tests/ogx/utils.py
+++ b/tests/ogx/utils.py
@@ -155,6 +155,7 @@ def create_ogx_server(
         namespace=namespace,
         distribution=config["distribution"],
         workload=config.get("workload"),
+        network=config.get("network"),
         tls=config.get("tls"),
         wait_for_resource=True,
         teardown=teardown,


eoinfennessy (approved):

jgarciao:
>Without externalAccess enabled, the operator-managed NetworkPolicy does not include the OpenShift router namespace (network.openshift.io/policy-group: ingress), so the public route returns 503.

Isn't this a consequence of removing it here? https://github.com/opendatahub-io/opendatahub-tests/pull/1997

What will be the consequence of enabling externalAccess? Will the ogx operator create a route?  We are already creating a test route with a random url for testing in conftest.py

Will this solution support regular an clusters using Entra OIDCs?

jgarciao (commented):
Reviewed it and added some questions

derekhiggins:
> > Without externalAccess enabled, the operator-managed NetworkPolicy does not include the OpenShift router namespace (network.openshift.io/policy-group: ingress), so the public route returns 503.
> 
> Isn't this a consequence of removing it here? #1997
> 
> What will be the consequence of enabling externalAccess? Will the ogx operator create a route? We are already creating a test route with a random url for testing in conftest.py
> 
> Will this solution support regular an clusters using Entra OIDCs?

Thats part of the story, #1997 essentially removed this label
          "kubernetes.io/metadata.name": "openshift-ingress",
which was allowing the route to work, but it only worked in clusters with pod-networked router configurations

adding externalAccess.enabled uses a different selector in the networkpolicy
          network.openshift.io/policy-group: ingress
which AIUI works on clusters with both pod-networked and host networked router configurations

I can't explain why/how my tests on the disconnected/Entra cluster didn't need this and without a disconnected cluster I can't check at the moment

It does not create a route

jgarciao (approved):

jgarciao:
Thanks for the info, Derek!

Artemon-line:
Thanks @derekhiggins this should work (enabling auto merge):
Checked the operator source (ogx-k8s-operator):

1. **No route conflict.** `externalAccess.enabled` creates a K8s Ingress, not a Route. The test route in conftest.py is separate — they coexist fine.

2. **The real fix is the NetworkPolicy.** When `spec.network` is set, the operator adds `network.openshift.io/policy-group: ingress` to NetworkPolicy peers, letting the router reach the pods. Without it → 503 on clusters that enforce NetworkPolicies (PSI, AWS). GCP passed because it doesn't enforce them.

3. **Entra OIDC — no issue.** The operator has no auth logic. The Ingress is plain HTTP, no auth annotations. No conflict with Entra OIDC.

Artemon-line (approved):

PR #2036, opened 2026-07-19
feat: add PVC storage e2e tests for EvalHub evaluation jobs 

kpunwatk:
Add integration tests for PVC as a storage source for evaluation provider test data, covering PVC mount, sub-path loading, missing PVC failure, read-only mount verification, and multiple providers on the same PVC.

# Pull Request

## Related Issues

- Fixes: 
- JIRA: 

- [ ] Locally
- [ ] Jenkins

## Additional Requirements

## Summary by CodeRabbit

* **New Features**
  * Updated EvalHub end-to-end tests to use tenant-scoped PVCs as provider-backed test data, including automatic population via a short-lived writer pod.
  * Added support for loading datasets from a specified `sub_path` within the same PVC.
  * Added stronger safety checks for PVC-backed job pods (e.g., read-only mounts and restricted security settings for the data writer).
* **Bug Fixes**
  * Improved coverage for the missing-PVC scenario to confirm a non-completed job outcome.
* **Tests**
  * Added a dedicated PVC storage test suite with validations for pod specs, mount flags, concurrent jobs on the same PVC, and absence of unexpected init containers/credential env vars.


tests/ai_safety/evalhub/test_evalhub_pvc_storage.py
```diff
@@ -0,0 +1,284 @@
...
+            payload=payload,
+        ) as job_id:
+            job_data = wait_for_evalhub_job(
+                host=evalhub_mt_route.host,
+                token=tenant_a_token,
+                ca_bundle_file=evalhub_mt_ca_bundle_file,
+                tenant=tenant_a_namespace.name,
+                job_id=job_id,
+            )
+            validate_evalhub_job_completed(job_data=job_data)
```
dbasunag: Should be a fixture.

tests/ai_safety/evalhub/test_evalhub_pvc_storage.py
```diff
@@ -0,0 +1,284 @@
...
```
dbasunag: Same comment.

tests/ai_safety/evalhub/test_evalhub_pvc_storage.py
```diff
@@ -0,0 +1,284 @@
...
+            tenant=tenant_a_namespace.name,
+            payload=payload,
+        ) as job_id:
+            batch_jobs = wait_for_evalhub_runtime_job_count(
+                admin_client=admin_client,
+                namespace=tenant_a_namespace.name,
+                evalhub_job_id=job_id,
+                minimum=1,
+            )
+            batch_job = batch_jobs[0]
```
dbasunag: Same comment. Also please consider parameterization if possible.

tests/ai_safety/evalhub/utils.py
```diff
@@ -873,6 +875,61 @@ def build_evalhub_job_payload(
...
+        host=host,
+        token=token,
+        ca_bundle_file=ca_bundle_file,
+        tenant=tenant,
+        payload=payload,
+    )
+    job_id = data["resource"]["id"]
+    try:
+        yield job_id
+    finally:
```
dbasunag: There is nothing to try/except here. Why not use a fixture instead or context manager?

tests/ai_safety/evalhub/conftest.py
```diff
@@ -1665,3 +1669,64 @@ def evalhub_otel_dual_sink_route(
...
+def evalhub_test_data_pvc(
+    admin_client: DynamicClient,
+    tenant_a_namespace: Namespace,
+) -> Generator[PersistentVolumeClaim, Any, Any]:
+    """Create a PVC for evaluation test data in the tenant namespace."""
+    with PersistentVolumeClaim(
+        client=admin_client,
+        name=PVC_TEST_DATA_NAME,
+        namespace=tenant_a_namespace.name,
+        accessmodes=PersistentVolumeClaim.AccessMode.RWO,
```
scheruku-rh, line 1686: ReadWriteMany or ReadOnlyMany access modes are recommended for PVCs used with EvalHub.

tests/ai_safety/evalhub/conftest.py
```diff
@@ -1665,3 +1669,64 @@ def evalhub_otel_dual_sink_route(
...
+
+@pytest.fixture(scope="class")
+def evalhub_test_data_populated(
+    admin_client: DynamicClient,
+    tenant_a_namespace: Namespace,
+    evalhub_test_data_pvc: PersistentVolumeClaim,
+) -> PersistentVolumeClaim:
+    """Populate the test data PVC with marker files at two sub-paths."""
+    populate_script = (
+        "mkdir -p /data/provider_a /data/provider_b && "
```
scheruku-rh, line 1701: The PVC is populated with dummy JSON files, not real dataset content. lm-eval needs HF arrow format files, can we atleast upload tokenizer files in the PVC and pass it through benchmark parameters. 
```
         "parameters": {
            "tokenizer":"/test_data/tokenizer"
         },
```
This way we can validate that the adapter pod reads PVC content and uses it during evaluation.

tests/ai_safety/evalhub/utils.py
```diff
@@ -873,6 +875,61 @@ def build_evalhub_job_payload(
...
+    claim_name: str,
+    sub_path: str | None = None,
+) -> dict:
+    """Build an EvalHub job payload with PVC-backed test data."""
+    payload = build_evalhub_job_payload(
+        model_service_name=model_service_name,
+        tenant_namespace=tenant_namespace,
+        job_name=job_name,
+    )
+    payload["test_data_ref"] = build_pvc_test_data_ref(claim_name=claim_name, sub_path=sub_path)
```
scheruku-rh: can you confirm that the `test_data_ref` is passed for each benchmark?

scheruku-rh (dismissed):

tests/ai_safety/evalhub/conftest.py
kpunwatk: On RWX: "The ROSA test cluster uses gp3-csi (EBS) which only supports ReadWriteOnce. RWX would require an EFS-backed StorageClass. Keeping RWO since tests run sequentially on the same node."

tests/ai_safety/evalhub/conftest.py
kpunwatk: Done. The build_vllm_arc_easy_benchmark() helper already sets "tokenizer": "google/flan-t5-small" in the benchmark parameters, so lm-eval downloads the tokenizer at runtime from HuggingFace — no need to pre-populate tokenizer files on the PVC. The PVC now only carries provider marker data for the sub-path tests, which is all the PVC storage feature needs to exercise (mount, sub-path routing, read-only verification). All 5 tests pass.

tests/ai_safety/evalhub/test_evalhub_pvc_storage.py
kpunwatk: Done

tests/ai_safety/evalhub/test_evalhub_pvc_storage.py
kpunwatk: Done

tests/ai_safety/evalhub/test_evalhub_pvc_storage.py
kpunwatk: Done

tests/ai_safety/evalhub/conftest.py
```diff
@@ -1665,3 +1671,111 @@ def evalhub_otel_dual_sink_route(
...
+    tenant_a_token: str,
+    tenant_a_namespace: Namespace,
+    evalhub_mt_ca_bundle_file: str,
+    evalhub_mt_route: Route,
+    evalhub_vllm_emulator_service: Service,
+) -> Generator[Callable[..., str], Any, Any]:
+    """Factory fixture: submit PVC evaluation jobs with guaranteed cleanup."""
+    job_ids: list[str] = []
+
+    def _submit(
```
dbasunag, line 1748: I don't understand the need for this function.

tests/ai_safety/evalhub/test_evalhub_pvc_storage.py
```diff
@@ -0,0 +1,246 @@
...
+        tenant_a_namespace: Namespace,
+        evalhub_mt_ca_bundle_file: str,
+        evalhub_mt_route: Route,
+        submit_pvc_job: Callable[..., str],
+        evalhub_test_data_populated: PersistentVolumeClaim,
+    ) -> None:
+        """Given a PVC with test data in the tenant namespace,
+        when an evaluation job is submitted with test_data_ref.pvc,
+        then the job completes successfully and results are persisted."""
+        job_id = submit_pvc_job(
```
dbasunag, line 39: submit_pvc_job yields the id, so you can simply directly use submit_pvc_job in place of job id.

tests/ai_safety/evalhub/test_evalhub_pvc_storage.py
```diff
@@ -0,0 +1,246 @@
...
+        evalhub_test_data_populated: PersistentVolumeClaim,
+    ) -> None:
+        """Given a PVC with data at a specific sub-path,
+        when an evaluation job specifies test_data_ref.pvc with sub_path,
+        then the job completes successfully using data from that sub-path."""
+        job_id = submit_pvc_job(
+            claim_name=evalhub_test_data_populated.name,
+            job_name="pvc-sub-path-test",
+            sub_path="provider_a",
+        )
```
dbasunag, line 102: Same comment here

tests/ai_safety/evalhub/conftest.py
kpunwatk: The factory pattern is intentional, tests pass different claim_name/sub_path args, and test_missing_pvc_job_fails uses claim_name="nonexistent-pvc".  It yields a callable so each test can pass different claim_name/sub_path arguments (e.g. test_missing_pvc_job_fails uses  claim_name="nonexistent-pvc", and test_multiple_providers_same_pvc submits two jobs). The inner function also tracks all job IDs for guaranteed cleanup in the teardown.

tests/ai_safety/evalhub/test_evalhub_pvc_storage.py
kpunwatk: submit_pvc_job is a factory fixture, it yields a callable, not the job_id directly. Each test calls it with different arguments (e.g. claim_name, sub_path) to submit a job and get the id back.

kpunwatk: therefore, you have to call it with arguments to get the id back.

tests/ai_safety/evalhub/test_evalhub_pvc_storage.py
kpunwatk: same as above

tests/ai_safety/evalhub/conftest.py
dbasunag: You can achieve the same with requests and parameterization

dbasunag (approved):

dbasunag (dismissed):

sheltoncyril (commented):
## Review Findings

### High Severity

1. **`tests/ai_safety/evalhub/conftest.py:1681` — RWO PVC with concurrent pods**
   `evalhub_test_data_pvc` uses `ReadWriteOnce` but `test_multiple_providers_same_pvc` submits two concurrent jobs that both mount it. On multi-node clusters, the second pod will fail with Multi-Attach error. Should be `ROX` (ReadOnlyMany) since eval jobs mount read-only.

2. **`tests/ai_safety/evalhub/test_evalhub_pvc_storage.py:150` — No-op assertion in `test_missing_pvc_job_fails`**
   Single GET + `assert state != "completed"` is trivially true for any in-progress job. Doesn't verify the job actually fails due to missing PVC. Should poll with bounded timeout then `assert state == "failed"`.

### Medium Severity

3. **`tests/ai_safety/evalhub/test_evalhub_pvc_storage.py:147` — Missing `raise_for_status()` before `.json()`**
   If the API returns 500/404, `.json()` throws `JSONDecodeError` with no indication of HTTP status. Add `.raise_for_status()` before `.json()`.

4. **`tests/ai_safety/evalhub/conftest.py:1769` — Teardown leak on partial deletion failure**
   `submit_pvc_job` cleanup loop iterates `job_ids` but doesn't catch per-iteration exceptions. If the first deletion raises, remaining jobs are never cleaned up.

### Low Severity

5. **`tests/ai_safety/evalhub/conftest.py:1718` — Security regression: `readOnlyRootFilesystem` dropped**
   Switching from `PVC_DATA_WRITER_SECURITY_CONTEXT` to `MINIO_UPLOADER_SECURITY_CONTEXT` silently drops `readOnlyRootFilesystem: True`. The data-writer pod only writes to the PVC mount, not rootfs.

6. **`tests/ai_safety/evalhub/conftest.py:1717` — Wrong image for task**
   MinIO MC image used for a pod that only runs `mkdir` and `echo`. The `mc` binary is never invoked. Original `ubi-minimal` was the right choice for shell-only ops.

tests/ai_safety/evalhub/conftest.py
```diff
@@ -1665,3 +1671,111 @@ def evalhub_otel_dual_sink_route(
...
     """CA bundle file for EvalHub OTEL routes."""
     return create_ca_bundle_file(client=admin_client)
+
+
+@pytest.fixture(scope="class")
+def evalhub_test_data_pvc(
+    admin_client: DynamicClient,
+    tenant_a_namespace: Namespace,
+) -> Generator[PersistentVolumeClaim, Any, Any]:
+    """Create a PVC for evaluation test data in the tenant namespace."""
```
sheltoncyril, line 1681: **[High]** `ReadWriteOnce` but `test_multiple_providers_same_pvc` submits two concurrent jobs that both mount this PVC. On multi-node clusters, second pod fails with Multi-Attach error. Should be `ROX` (ReadOnlyMany) since eval jobs mount read-only.

tests/ai_safety/evalhub/test_evalhub_pvc_storage.py
```diff
@@ -0,0 +1,246 @@
...
+        ]
+        assert len(pvc_volumes) >= 1, (
+            f"Expected PVC volume referencing nonexistent-pvc in pod spec, "
+            f"got volumes: {[v.name for v in spec.volumes]}"
+        )
+        assert pvc_volumes[0].persistentVolumeClaim.claimName == "nonexistent-pvc"
+
+        job_data = get_evalhub_job_http(
+            host=evalhub_mt_route.host,
+            token=tenant_a_token,
```
sheltoncyril, line 149: **[High]** No-op assertion: single GET + `!= "completed"` is trivially true for any in-progress job. Does not verify job actually fails from missing PVC. Should poll with bounded timeout then `assert state == "failed"`.

tests/ai_safety/evalhub/test_evalhub_pvc_storage.py
```diff
@@ -0,0 +1,246 @@
...
+
+        pvc_volumes = [
+            volume for volume in (spec.volumes or []) if getattr(volume, "persistentVolumeClaim", None) is not None
+        ]
+        assert len(pvc_volumes) >= 1, (
+            f"Expected PVC volume referencing nonexistent-pvc in pod spec, "
+            f"got volumes: {[v.name for v in spec.volumes]}"
+        )
+        assert pvc_volumes[0].persistentVolumeClaim.claimName == "nonexistent-pvc"
+
```
sheltoncyril, line 146: **[Medium]** Missing `raise_for_status()` before `.json()`. If API returns 500/404, this throws `JSONDecodeError` with no indication of HTTP status.

tests/ai_safety/evalhub/conftest.py
```diff
@@ -1665,3 +1671,111 @@ def evalhub_otel_dual_sink_route(
...
+        data = submit_evalhub_job(
+            host=evalhub_mt_route.host,
+            token=tenant_a_token,
+            ca_bundle_file=evalhub_mt_ca_bundle_file,
+            tenant=tenant_a_namespace.name,
+            payload=payload,
+        )
+        job_id = data["resource"]["id"]
+        job_ids.append(job_id)
+        return job_id
```
sheltoncyril, line 1769: **[Medium]** Teardown loop does not catch per-iteration exceptions. If first deletion raises (network timeout, 500, job already gone), remaining `job_ids` are never cleaned up — orphaned eval jobs and K8s batch Jobs left on cluster.

tests/ai_safety/evalhub/conftest.py
```diff
@@ -1665,3 +1671,111 @@ def evalhub_otel_dual_sink_route(
...
+        client=admin_client,
+        name=pod_name,
+        namespace=tenant_a_namespace.name,
+        restart_policy="Never",
+        volumes=[{"name": "test-data", "persistentVolumeClaim": {"claimName": evalhub_test_data_pvc.name}}],
+        containers=[
+            {
+                "name": "data-writer",
+                "image": MINIO_MC_IMAGE,
+                "command": ["/bin/sh", "-c"],
```
sheltoncyril, line 1718: **[Low]** Switching to `MINIO_UPLOADER_SECURITY_CONTEXT` silently drops `readOnlyRootFilesystem: True`. Data-writer pod only writes to PVC mount, not rootfs — the original constraint was correct defense-in-depth.

tests/ai_safety/evalhub/conftest.py
```diff
@@ -1665,3 +1671,111 @@ def evalhub_otel_dual_sink_route(
...
+    with Pod(
+        client=admin_client,
+        name=pod_name,
+        namespace=tenant_a_namespace.name,
+        restart_policy="Never",
+        volumes=[{"name": "test-data", "persistentVolumeClaim": {"claimName": evalhub_test_data_pvc.name}}],
+        containers=[
+            {
+                "name": "data-writer",
+                "image": MINIO_MC_IMAGE,
```
sheltoncyril, line 1717: **[Low]** MinIO MC image used for pod that only runs `mkdir` and `echo`. The `mc` binary is never invoked. Original `ubi-minimal` was the right choice for shell-only ops.

tests/ai_safety/evalhub/conftest.py
kpunwatk: done c81ff89b

tests/ai_safety/evalhub/conftest.py
kpunwatk: We're keeping MINIO_MC for now since it's already in AiSafetyImages and adding a new UBI_MINIMAL image constant would need a separate change.

tests/ai_safety/evalhub/conftest.py
kpunwatk: being addressed securityContext": {**MINIO_UPLOADER_SECURITY_CONTEXT, "readOnlyRootFilesystem": True},

dbasunag (approved):

tests/ai_safety/evalhub/conftest.py
scheruku-rh: The issue is that with the current approach, lm-eval never actually reads anything from the PVC — the dummy JSON marker files are never consumed by the adapter, and the tokenizer is pulled from HF independently. So the tests only validate the K8s changes (volume is mounted, readOnly is set) but not that the adapter successfully reads from the PVC during evaluation.
The idea was: have the writer pod download the tokenizer and write it to the PVC (e.g. /data/tokenizer), then pass it via:

```
"parameters": {
    "tokenizer": "/test_data/tokenizer"
}
```
This way `validate_evalhub_job_completed` actually proves the full end-to-end PVC read path i.e. adapter picks up the tokenizer from the PVC mount and uses it during evaluation. Can we update the writer pod to do this?

sheltoncyril (approved):
/lgtm

sheltoncyril:
Time sensitive PR was merged

PR #2043, opened 2026-07-20
test(upgrade): add AdmissionCheck upgrade test

cgoodfred:
# Pull Request

## Summary

Add pre/post upgrade test for Kueue AdmissionCheck CRD (Task 9 from the Kueue Upgrade Test Coverage Audit).

Pre-upgrade: creates an AdmissionCheck, associates it with a ClusterQueue, submits a Job, and verifies the Workload is blocked (QuotaReserved but not Admitted).

Post-upgrade: verifies the AdmissionCheck and Workload survived the upgrade and the gate still holds, then approves the AdmissionCheck and validates the Workload is admitted and the Job completes.

Refactors ensure_kueue_for_ac_upgrade to use the shared _ensure_kueue_available_for_upgrade helper from PR 1888, eliminating duplicated DSC state management code.

## Related Issues

- Fixes: 
- JIRA: [RHOAIENG-76938](https://redhat.atlassian.net/browse/RHOAIENG-76938)

## Please review and indicate how it has been tested

- [x] Locally
- [ ] Jenkins

## Additional Requirements

- [ ] If this PR introduces a new test image, did you create a PR to mirror it in disconnected environment?
- [ ] If this PR introduces new marker(s)/adds a new component, was relevant ticket created to update relevant Jenkins job?

[RHOAIENG-76938]: https://redhat.atlassian.net/browse/RHOAIENG-76938?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end upgrade coverage validating AdmissionCheck-gated Workloads stay blocked pre-upgrade, remain gated after upgrade, and are admitted after approval; verifies job completion.
  * Added coverage for AdmissionCheck persistence and that ClusterQueue continues referencing the AdmissionCheck.
* **Test Infrastructure**
  * Added utilities/fixtures to create and activate AdmissionChecks, patch admission/workload status during upgrades, and validate AdmissionCheck/Workload/ClusterQueue conditions.
  * Extended upgrade setup to ensure BatchJob framework is present and added an AdmissionCheck resource model for test generation.

PR #2068, opened 2026-07-22
test(upgrade): add AdmissionCheck upgrade test for 2.25

cgoodfred:
# Pull Request

## Summary

Backport of #2043:  add pre/post upgrade test for Kueue AdmissionCheck CRD to the 2.25 branch.

Pre-upgrade: creates an AdmissionCheck, associates it with a ClusterQueue, submits a Job, and verifies the Workload is blocked (QuotaReserved but not Admitted).

Post-upgrade: verifies the AdmissionCheck and Workload survived the upgrade and the gate still holds, then approves the AdmissionCheck and validates the Workload is admitted and the Job completes.

Adapted to 2.25 branch conventions: v1 DSC API, Kueue v1beta1, inline DSC state management with separate ConfigMap, simple_logger, and typing module imports.

## Related Issues

- Fixes: 
- JIRA: [RHOAIENG-76938](https://redhat.atlassian.net/browse/RHOAIENG-76938)

- [x] Locally
- [ ] Jenkins

## Additional Requirements

[RHOAIENG-76938]: https://redhat.atlassian.net/browse/RHOAIENG-76938?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/model_serving/model_server/upgrade/conftest.py
```diff
@@ -867,3 +898,315 @@ def skip_if_not_raw_deployment(
...
+                "metadata": {
+                    "name": AC_JOB_NAME,
+                    "namespace": namespace,
+                },
+                "spec": {
+                    "suspend": True,
+                    "backoffLimit": 0,
+                    "template": {
+                        "metadata": {
+                            "labels": {"kueue.x-k8s.io/queue-name": local_queue.name},
```
sukumars321: **Change request:** `kueue.x-k8s.io/queue-name` is on the wrong object.

Kueue’s Job integration expects this label on **Job `metadata.labels`**, not on `spec.template.metadata.labels` (pod template). As written, Kueue may never associate the Job with the LocalQueue / create a managed Workload, and the pre-upgrade wait for Workload + `QuotaReserved` will fail.

Per [Kueue Job docs](https://kueue.sigs.k8s.io/docs/tasks/run/jobs/):

```yaml
metadata:
  labels:
    kueue.x-k8s.io/queue-name: <local-queue>
spec:
  suspend: true
  template:
    # queue-name does not belong here
```

**Suggested fix:** move the label to Job metadata and drop it from the pod template:

```python
"metadata": {
    "name": AC_JOB_NAME,
    "namespace": namespace,
    "labels": {"kueue.x-k8s.io/queue-name": local_queue.name},
},
"spec": {
    "suspend": True,
    "backoffLimit": 0,
    "template": {
        "spec": {
            "restartPolicy": "Never",
            "containers": [...],
        },
    },
},
```

tests/model_serving/model_server/upgrade/conftest.py
```diff
@@ -867,3 +898,315 @@ def skip_if_not_raw_deployment(
...
+                        },
+                        "spec": {
+                            "restartPolicy": "Never",
+                            "containers": [
+                                {
+                                    "name": "test",
+                                    "image": "registry.access.redhat.com/ubi9/ubi-minimal:latest",
+                                    "command": ["echo", "admission-check-test"],
```
sukumars321, line 1097: **Change request:** Job container is missing resource requests.

Kueue needs `cpu`/`memory` requests to account quota against the ClusterQueue flavors. This container only sets `name`, `image`, and `command` — with no requests, quota accounting cannot work, and the pre-upgrade wait for `QuotaReserved=True` is likely to time out.

See [Kueue Job docs](https://kueue.sigs.k8s.io/docs/tasks/run/jobs/) (sample jobs always declare requests) and the existing KServe Kueue upgrade helpers in this repo, which set requests/limits via the upgrade utils.

**Suggested fix:** add modest requests aligned with `AC_CPU_QUOTA` / `AC_MEMORY_QUOTA` (and the ResourceFlavor covered resources), for example:

```python
{
    "name": "test",
    "image": "registry.access.redhat.com/ubi9/ubi-minimal:latest",
    "command": ["echo", "admission-check-test"],
    "resources": {
        "requests": {"cpu": "100m", "memory": "128Mi"},
        "limits": {"cpu": "200m", "memory": "256Mi"},
    },
}
```

Constants can live next to the other `AC_*` values in `admission_check_upgrade_config.py` if you want them shared/tunable.

tests/model_serving/model_server/upgrade/conftest.py
```diff
@@ -867,3 +898,315 @@ def skip_if_not_raw_deployment(
...
+            model_mesh_enabled=False,
+            add_dashboard_label=True,
+            add_kueue_label=True,
+            teardown=teardown_resources,
+        ) as ns:
+            yield ns
+
+
+@pytest.fixture(scope="session")
+def ensure_kueue_for_ac_upgrade(
```
sukumars321: **Change request:** Duplicate DSC Kueue Unmanaged ownership will collide if both Kueue upgrade suites run in one session.

`ensure_kueue_for_ac_upgrade` is nearly a copy of `ensure_kueue_for_kserve_upgrade`, but with a different ConfigMap (`upgrade-ac-kueue-dsc-state` vs `upgrade-kueue-dsc-state`). Both mutate the same DSC `components.kueue.managementState`.

If both fixtures run pre-upgrade:
1. First suite snapshots the real original state and patches → `Unmanaged`
2. Second suite reads current state (`Unmanaged`) and saves that as “original”
3. Restore order then depends on fixture teardown — final DSC state can be wrong

**Is a shared helper enough?** Partially.

- **Yes — extract a shared helper** (DRY) for save → patch Unmanaged → yield → restore. Something like `_ensure_kueue_unmanaged_for_upgrade(...)` used by both fixtures removes the copy/paste.
- **No — a helper alone does not fix the bug** if each suite still keeps its own ConfigMap. You need **single ownership** of the DSC snapshot/restore.

**Preferred approach:** one session-scoped fixture (or one shared helper + one well-known ConfigMap contract) that both KServe and AdmissionCheck upgrade suites depend on:

```python
def ensure_kueue_unmanaged_for_upgrade(...):
    # single CM (e.g. UPGRADE_KUEUE_DSC_STATE_CM_NAME)
    # snapshot once, patch once, restore once
    ...
```

Then:
- `ensure_kueue_for_kserve_upgrade` / AC resources fixtures just depend on it
- Drop `AC_DSC_STATE_CM` as a second source of truth

If the suites are guaranteed never to run together, say so explicitly in the docstring — but given they live in the same upgrade package, shared ownership is the safer default.

utilities/kueue_utils.py
```diff
@@ -159,6 +211,99 @@ def wait_for_deployments(labels: list[str], namespace: str, admin_client: Dynami
...
+    """Check if a Kueue Workload has QuotaReserved=True condition."""
+    conditions = getattr(workload.instance.status, "conditions", None) or []
+    return any(
+        (c.get("type") if isinstance(c, dict) else getattr(c, "type", None)) == "QuotaReserved"
+        and (c.get("status") if isinstance(c, dict) else getattr(c, "status", None)) == "True"
+        for c in conditions
+    )
+
+
+def activate_admission_check(
```
sukumars321, line 287: **Change request (option 2 — keep fake ACC, harden it):** The status-spoof helpers are a reasonable test double for a generic AdmissionCheck upgrade gate, but the contract should be explicit and post-upgrade should not assume `Active` survived.

Kueue expects a real AdmissionCheck Controller to own `AdmissionCheck.status` (`Active`) and Workload check state (`Ready`/`Pending`). This suite intentionally uses a fake `controllerName` (`upgrade-test-controller.example.com`) and admin status-patches (`activate_admission_check` / `approve_admission_check_on_workload`) so we can hold a workload gated across upgrade without deploying ProvisioningRequest/MultiKueue. That’s fine for this test’s goal — **please don’t treat it as a temp hack to delete** — but as written it is brittle:

1. Pre-upgrade activates AC once; nothing re-asserts `Active` after upgrade.
2. Post-upgrade only checks AC **exists**, then patches Workload `Ready`.
3. If reconcile/upgrade clears `Active`, CQ/admission behavior can break and the approve path becomes flaky.

**Please:**

1. **Document the test double** on the helpers / fixture (why fake `controllerName`, that the test acts as the ACC).
2. **Post-upgrade:** assert AC `Active=True` (or call `activate_admission_check` again before approve if not Active).
3. Optionally assert Workload admission-check state is still pending before approve.

Using a real ProvisioningRequest ACC is a larger follow-up (CA/ProvReq infra + poor fit for “hold pending across upgrade”) — out of scope for this PR.

utilities/kueue_utils.py
```diff
@@ -159,6 +211,99 @@ def wait_for_deployments(labels: list[str], namespace: str, admin_client: Dynami
...
+                        "lastTransitionTime": datetime.now(tz=UTC).strftime("%Y-%m-%dT%H:%M:%SZ"),
+                    }
+                ]
+            }
+        },
+        content_type="application/merge-patch+json",
+    )
+
+
+def approve_admission_check_on_workload(
```
sukumars321, line 321: **Change request:** Please document the merge-patch behavior (and its assumptions) on `approve_admission_check_on_workload`.

This uses `application/merge-patch+json` against `status.admissionChecks`. Per JSON Merge Patch, **arrays are replaced wholesale** — they are not merged by `name`. So this patch overwrites the entire `admissionChecks` list with the single Ready entry you send.

That is acceptable for this suite’s fake single-check path (no real ACC writing `podSetUpdates`), but it is easy to misuse later (multiple checks, ProvisioningRequest `podSetUpdates`, etc.).

**Please add a short docstring note**, e.g.:

- Uses JSON merge-patch, which **replaces** `status.admissionChecks` entirely
- Safe while there is exactly one AdmissionCheck and no reliance on sibling fields like `podSetUpdates`
- Callers that need to preserve existing entries/fields should read-modify-write instead

No functional change required for this PR unless you already plan to harden it.

tests/model_serving/model_server/upgrade/test_upgrade_admission_check.py
```diff
@@ -0,0 +1,167 @@
...
+        LOGGER.info(
+            f"[PRE-UPGRADE] PASS: Job is gated by AdmissionCheck, "
+            f"job={admission_check_job.name}, workload={admission_check_workload.name}"
+        )
+
+
+class TestAdmissionCheckPostUpgrade:
+    """Post-upgrade: verify AdmissionCheck still gates, then approve and validate admission."""
+
+    @pytest.mark.post_upgrade
```
sukumars321, line 93: **Change request:** Post-upgrade coverage is thinner than the suite’s stated goal.

Today post-upgrade checks:
1. AdmissionCheck **resource exists**
2. Workload still **not Admitted**
3. Spoof approve → Admitted + Job Complete

What’s missing for an AdmissionCheck **upgrade** regression:

- AdmissionCheck still **Active** (ties to the fake-ACC hardening comment)
- ClusterQueue still references the check via `admissionChecksStrategy` / expected check name (config survival, not just the AC object)
- Optionally: Workload `status.admissionChecks[].state` still `Pending` before approve (stronger than only `not Admitted`)

Existence + still-gated is useful, but a CQ that dropped the strategy (or an inactive AC) can still look “gated” for other reasons and hide an upgrade regression.

**Please add** at least:
- assert AC `Active=True` (or re-activate + assert), and
- assert CQ still lists `AC_ADMISSION_CHECK_NAME` in its admission-check strategy

Happy to keep approve/admit as the final functional proof once those survival checks pass.

sukumars321:
**Update on `admissionChecksStrategy`:** Confirmed it **does** exist on the 2.25 ClusterQueue CRD — earlier check was mistaken.

Please disregard the prior change request about switching to top-level `admissionChecks`. No API-field change needed for that item. Sorry for the noise.

tests/model_serving/model_server/upgrade/conftest.py
cgoodfred: Done — moved label to Job metadata.labels and dropped it from pod template.

tests/model_serving/model_server/upgrade/conftest.py
cgoodfred: Done — added requests/limits to the container and put the constants in admission_check_upgrade_config.py (AC_JOB_CPU_REQUEST, AC_JOB_MEMORY_REQUEST, AC_JOB_CPU_LIMIT, AC_JOB_MEMORY_LIMIT).

tests/model_serving/model_server/upgrade/conftest.py
cgoodfred: Done — extracted _ensure_kueue_dsc_for_upgrade shared helper in conftest.py. Both ensure_kueue_for_kserve_upgrade and ensure_kueue_for_ac_upgrade now yield from it with their own namespace. Both use the same CM name (UPGRADE_KUEUE_DSC_STATE_CM_NAME) stored in different namespaces, so no collision. Dropped AC_DSC_STATE_CM. Matches the pattern on main where _ensure_kueue_available_for_upgrade is shared in utils.py.

utilities/kueue_utils.py
cgoodfred: Done — added docstrings to activate_admission_check and approve_admission_check_on_workload documenting the fake ACC pattern and why we use it. Post-upgrade now asserts Active=True on the AdmissionCheck before proceeding to approve.

utilities/kueue_utils.py
cgoodfred: Done — added docstring to approve_admission_check_on_workload noting that merge-patch replaces status.admissionChecks entirely, is safe for single-check scenarios, and callers with multiple checks should read-modify-write.

tests/model_serving/model_server/upgrade/test_upgrade_admission_check.py
cgoodfred: Done — test_admission_check_exists now asserts AC Active=True via check_admission_check_active and asserts CQ still lists the AC in admissionChecksStrategy via check_cluster_queue_has_admission_check. Both helpers added to kueue_utils.py.

dbasunag (changes requested):
Same comments as the base PR. Going forward, it would be great, if we can let the base pr get merged first, before backporting.

dbasunag (dismissed):

dbasunag (approved):

cgoodfred:
Closing based on slack conversation with @threcc determining this wasn't necessary.

PR #2072, opened 2026-07-22
Discover spark operator RBAC and NetworkPolicy from operator namespace instead of creating hardcoded resources

alimaredia:
# Pull Request

## Summary

Instead of creating Role, RoleBinding, ServiceAccount, and NetworkPolicy from scratch with hardcoded rules, the upgrade test now discovers these resources from the applications namespace (where the spark operator runs) and recreates them in the upgrade-spark-operator test namespace. This ensures the test reflects the actual operator-provisioned resources, catching any RBAC or NetworkPolicy changes between upgrades.

## Related Issues

- Fixes: 
- JIRA: 

- [ ] Locally
- [ ] Jenkins

## Additional Requirements

## Summary by CodeRabbit

- **Tests**
  - Expanded Spark upgrade coverage for applications using multiple service accounts and role-based access resources.
  - Added validation that roles, role bindings, service accounts, and network policies are recreated correctly in the upgrade namespace.
  - Updated upgrade scenarios to use recreated service accounts when validating Spark application behavior.
  - Added coverage for preserving access controls and network policies during namespace transitions.


alimaredia:
@dbasunag Can you review and merge this PR?

tests/spark/upgrade/conftest.py
```diff
@@ -160,92 +169,157 @@ def spark_role_fixture(
...
+
+        created_roles = []
+        for source_role in source_roles:
+            role = recreate_role_in_namespace(
+                client=admin_client,
+                source_role=source_role,
+                target_namespace=spark_namespace_fixture.name,
+                teardown=teardown_resources,
+            )
+            created_roles.append(role)
```
dbasunag: Why not use context manager?

tests/spark/upgrade/conftest.py
```diff
@@ -160,92 +169,157 @@ def spark_role_fixture(
...
+    Pre-upgrade: Discovers Roles from operator's namespace, recreates in upgrade namespace
+    Post-upgrade: References existing Roles in upgrade namespace
     """
-    role_kwargs = {
-        "client": admin_client,
-        "name": "spark-operator-role",
-        "namespace": spark_namespace_fixture.name,
-    }
+    if pytestconfig.option.post_upgrade:
+        yield get_spark_roles(client=admin_client, namespace=spark_namespace_fixture.name)
```
dbasunag: You need to clean up the roles you created during pre-upgrade, at teardown time.

tests/spark/upgrade/conftest.py
```diff
@@ -160,92 +169,157 @@ def spark_role_fixture(
...
+        created_sas = []
+        for source_sa in src_sas:
+            sa = recreate_service_account_in_namespace(
+                client=admin_client,
+                source_sa=source_sa,
+                target_namespace=spark_namespace_fixture.name,
+                teardown=teardown_resources,
+            )
+            created_sas.append(sa)
+
```
dbasunag: Anything you create in pre-upgrade, must be cleaned up at the teardown stage of post upgrade.

tests/spark/upgrade/conftest.py
```diff
@@ -160,92 +169,157 @@ def spark_role_fixture(
...
+    Pre-upgrade: Discovers RoleBindings from operator's namespace, recreates in upgrade namespace
+    Post-upgrade: References existing RoleBindings in upgrade namespace
     """
-    sa_kwargs = {
-        "client": admin_client,
-        "name": SPARK_SERVICE_ACCOUNT,
-        "namespace": spark_namespace_fixture.name,
-    }
+    if pytestconfig.option.post_upgrade:
+        yield get_spark_role_bindings(client=admin_client, namespace=spark_namespace_fixture.name)
```
dbasunag: Same comment about teardown.

dbasunag (changes requested):

alimaredia:
@dbasunag Addressed your comments

tests/spark/upgrade/conftest.py
```diff
@@ -160,92 +169,158 @@ def spark_role_fixture(
...
     """
-    role_kwargs = {
-        "client": admin_client,
-        "name": "spark-operator-role",
-        "namespace": spark_namespace_fixture.name,
-    }
+    if pytestconfig.option.post_upgrade:
+        stale_roles = get_spark_roles(client=admin_client, namespace=spark_namespace_fixture.name)
+        for role in stale_roles:
+            role.clean_up()
```
dbasunag, line 181: Should not this be after yield? Otherwise, at the beginning of your post upgrade these would be cleaned up.

tests/spark/upgrade/conftest.py
```diff
@@ -160,92 +169,158 @@ def spark_role_fixture(
...
+
+    created_roles = []
+    for source_role in source_roles:
+        role = recreate_role_in_namespace(
+            client=admin_client,
+            source_role=source_role,
+            target_namespace=spark_namespace_fixture.name,
+            teardown=teardown_resources,
+        )
+        created_roles.append(role)
```
dbasunag, line 199: Should be under a clear else. That way this code does not execute for post upgrade.

tests/spark/upgrade/conftest.py
```diff
@@ -160,92 +169,158 @@ def spark_role_fixture(
...
+        stale_sas = get_spark_service_accounts(client=admin_client, namespace=spark_namespace_fixture.name)
+        for sa in stale_sas:
+            sa.clean_up()
+
+    apps_namespace = py_config["applications_namespace"]
+    src_sas = get_spark_service_accounts(client=admin_client, namespace=apps_namespace)
+    assert src_sas, (
+        f"No spark-related ServiceAccounts found in namespace {apps_namespace}. "
+        "Ensure the Spark Operator is enabled and has created RBAC resources."
+    )
```
dbasunag, line 226: Same comment here and in following fixtures.

dbasunag (changes requested):

tests/spark/upgrade/conftest.py
alimaredia: That is fine, because it's possible after the upgrade there could be entirely new RBAC resources that are created. Cleaning the old operators RBAC resources future proofs the new RBAC resources that might be created in a new release and ensures that SparkApplications aren't using anything that's stale.

tests/spark/upgrade/conftest.py
alimaredia: Since the roles need are being deleted pre upgrade the roles need to be created again. Can you explain why you don't think the code needs to execute post upgrade?

tests/spark/upgrade/conftest.py
alimaredia: Per my comments above all the fixtures need to discover and recreate resources in both
  pre-upgrade and post-upgrade phases

dbasunag: I don't understand why we need these in upgrade lane then. You can add these tests as install and ensure testops runs them on upgraded cluster. Upgrade tests typically involves checking that functionalities did not degrade after upgrade and that is only possible if you retain the resources after pre-upgrade and check them after upgrade is complete.

tests/spark/upgrade/conftest.py
```diff
@@ -266,14 +341,12 @@ def spark_application_fixture(
     if pytestconfig.option.post_upgrade:
         yield spark_app
     else:
-        # Create SparkApplication spec
         spec = create_spark_pi_application_spec(
```
sheltoncyril, line 344: An AI-assisted review flagged this and I wanted to check before treating it as actionable.

`service_account_fixture[0]` relies on the order of `get_spark_service_accounts`, which returns controller/spark/webhook SAs. Since `.get()` ordering isn't guaranteed, `[0]` may resolve to `spark-operator-controller` rather than the `-spark` job SA the driver needs, which could fail the job nondeterministically. Is the ordering guaranteed somewhere, or should this select the job SA explicitly? Thanks!

tests/spark/upgrade/conftest.py
dbasunag: You only want to create resources during pre-upgrade and delete them during post upgrade. That is the typical upgrade test flow.

tests/spark/upgrade/conftest.py
```diff
@@ -27,7 +40,44 @@
...
+    dsc_resource: DataScienceCluster,
+) -> Generator[DataScienceCluster, Any, Any]:
+    """Reset Spark before and after pre-upgrade runs when resource deletion is requested."""
+    if (
+        not pytestconfig.option.delete_pre_upgrade_resources
+        or not pytestconfig.option.pre_upgrade
+        or pytestconfig.option.post_upgrade
+    ):
+        yield dsc_resource
+        return
```
dbasunag, line 58: Please use if/else, that way this awkward return statement won't be needed.

tests/spark/upgrade/conftest.py
```diff
@@ -27,7 +40,44 @@
 LOGGER = structlog.get_logger(name=__name__)
 
 UPGRADE_NAMESPACE = "upgrade-spark-operator"
-SPARK_SERVICE_ACCOUNT = "spark-operator-spark"
+
+
+@pytest.fixture(scope="session")
+def dsc_resource(  # noqa: UFN001
```
dbasunag, line 46: Please don't reuse the fixture name. We already have a dsc_resource that that is why you are getting the not unique name error.

tests/spark/upgrade/conftest.py
```diff
@@ -27,7 +40,44 @@
...
+        LOGGER.info(f"Deleting previous Spark test namespace {UPGRADE_NAMESPACE}")
+        namespace.clean_up(wait=True, timeout=300)
+
+    LOGGER.info("Resetting Spark Operator to Removed before pre-upgrade setup")
+    editor = ResourceEditor(
+        patches={
+            dsc_resource: {
+                "spec": {"components": {"sparkoperator": {"managementState": DscComponents.ManagementState.REMOVED}}}
+            }
+        }
```
dbasunag, line 71: I don't understand the logic. if you are testing spark, why are you setting it to removed? Clean up should happen after post-upgrade run only.

tests/spark/upgrade/conftest.py
```diff
@@ -160,92 +191,158 @@ def spark_role_fixture(
...
     teardown_resources: bool,
-) -> Generator[Role, Any, Any]:
-    """Create or reference the Spark Role with necessary permissions.
+) -> Generator[list[Role], Any, Any]:
+    """Copy only spark-role from the applications namespace into the test namespace."""
+    if pytestconfig.option.post_upgrade:
+        resources = get_spark_roles(client=admin_client, namespace=spark_namespace_fixture.name)
+        assert resources, "Required pre-upgrade resources are missing from the test namespace"
+        yield resources
+        return
```
dbasunag, line 200: Not needed. if you use if/else.

dbasunag, line 200: Please do clean up after yield.

dbasunag (changes requested):
The clean up logic does not exist for post upgrade.

sheltoncyril:
@alimaredia bumping the above comments

alimaredia:
@aloganat In these tests I have a 4 hour timeout period for RHOAI upgrades to happen. Is that too long or too short.

alimaredia:
@sheltoncyril I should have addressed the comments you had

alimaredia:
@dbasunag can you take another look at my latest revisions?

PR #1874, opened 2026-06-29
fix(ogx): add tool_choice to upgrade RAG test, remove stale xfail

Artemon-line:
## Summary
- Add `tool_choice="required"` and `include=["file_search_call.results"]` to the upgrade RAG test `responses.create()` call — makes `file_citation` annotations deterministic
- Remove `@pytest.mark.xfail(reason="RHAIENG-3650")` from post-upgrade test — RHAIENG-3650 (PVC conflict on 3.2→3.3 rolling upgrade) does not apply to the 3.4→3.5 disruptive upgrade path

## Root Cause
Without `tool_choice="required"`, annotation generation depends on the model (`Qwen3.6-35B-A3B`, 3B active params) emitting correct `<|file-ID|>` citation markers in its response text. The server-side parser uses a strict regex that rejects malformed markers — a known limitation of small/medium models (RHAIENG-2732). This caused ~70% failure rate across all upgrade pipelines.

## Verified on cluster
Reproduced on ssw-gcp-upg1 (RHOAI 3.4.2, OCP 4.21.22, GCP):

| Parameter | file_search_call | annotations |
|-----------|-----------------|-------------|
| Default (no tool_choice) | completed, 10 results | **0** → FAILED |
| `tool_choice="required"` + `include` | completed, 10 results | **12** → PASSED |

## Test plan
- [x] Reproduced original failure on live cluster (3.4 branch, pre-upgrade)
- [x] Verified `tool_choice="required"` produces annotations deterministically
- [x] Pre-upgrade test passes with fix applied (hard assertions intact)
- [ ] CI upgrade pipeline passes with this change
- [x] Cherry-pick to `3.4` branch

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Tests**
  * Updated automated checks for vector-store-based RAG behavior to align with the latest expected response format.
  * Removed an expected-failure marker from the post-upgrade scenario, so the test now runs as a passing check.

PR #2073, opened 2026-07-22
fix(ogx): remove stale xfail from post-upgrade chat completions test

Artemon-line:
## Summary
- Remove `@pytest.mark.xfail(reason="RHAIENG-3650")` from the post-upgrade chat completions test
- The PVC multi-attach bug is resolved and the xfail is no longer needed
- This was missed in PR #1874 which removed the same xfail from the vector store RAG post-upgrade test

## Test plan
- [x] Verified the xfail is the only change
- [ ] CI upgrade pipeline passes with this change


dbasunag (approved):

SB159 (approved):
/lgtm

PR #2119, opened 2026-07-29
feat(mlserver): add GPU and upgrade test suites

Snomaan6846:
Signed-off-by: Syed Nomaan Ali <syedali@redhat.com>

rh-pre-commit.version: 2.3.2
rh-pre-commit.check-secrets: ENABLED

# Pull Request

## Summary

Add MLServer CUDA GPU tests (build, serving, fallback, perf, RBAC) and CPU upgrade survivability tests with shared fixture architecture.

- mlserver_serving_runtime selects CPU or CUDA template via 'gpu': True param; CPU tests unaffected by --supported-accelerator-type
- mlserver_inference_service supports gpu_count, enable_auth, wait, wait_for_predictor_pods; volumes always attached
- run_mlserver_inference uses external route with port-forward fallback; pod_name parameter removed
- Add MLSERVER_TEMPLATE_MAP, MLSERVER_ACCELERATOR_IDENTIFIER constants
- Add RuntimeTemplates.MLSERVER_CUDA, ModelInferenceRuntime.MLSERVER_CUDA_RUNTIME
- Add ONNX_RESNET50_REST_INPUT_QUERY, measure_average_latency helper
- Update ModelCarImage.MLSERVER_ONNX to ResNet-50 digest
- Consolidate GPU SA into shared mlserver_model_service_account

## Related Issues

- Fixes: 
- JIRA: https://redhat.atlassian.net/browse/RHOAIENG-71768

- [x] Locally
- [ ] Jenkins

## Additional Requirements

## Summary by CodeRabbit

* **New Features**
  * Expanded accelerator-aware GPU-enabled MLServer coverage, including CUDA runtime template support and bearer-token authenticated REST inference.
  * Added GPU vs CPU performance testing for ResNet-50 ONNX, with speedup threshold warnings.
  * Introduced an MLServer upgrade scenario test to verify inference continuity and restart stability.
* **Bug Fixes**
  * Improved REST inference routing to use externally exposed endpoints when available, with clearer fallback behavior.
  * Adjusted external routing defaults to be configurable instead of forced on.
* **Tests**
  * Added GPU build-artifact and template validation, GPU scheduling/fallback scenarios, RBAC enforcement, and concurrent inference checks.
  * Added a new `mlserver_nvidia_gpu` marker for GPU-specific runs.

### DIFF
--- a/tests/model_serving/model_runtime/mlserver/conftest.py
+++ b/tests/model_serving/model_runtime/mlserver/conftest.py
@@ -23,6 +23,9 @@ from ocp_resources.serving_runtime import ServingRuntime
 from syrupy.extensions.json import JSONSnapshotExtension
 
 from tests.model_serving.model_runtime.mlserver.constant import (
+    MLSERVER_ACCELERATOR_IDENTIFIER,
+    MLSERVER_RUNTIME_NAME_MAP,
+    MLSERVER_TEMPLATE_MAP,
     PREDICT_RESOURCES,
 )
 from utilities.constants import (
@@ -41,25 +44,40 @@ def mlserver_serving_runtime(
     request: pytest.FixtureRequest,
     admin_client: DynamicClient,
     model_namespace: Namespace,
-    mlserver_runtime_image: str,
+    supported_accelerator_type: str | None,
+    mlserver_runtime_image: str | None,
 ) -> Generator[ServingRuntime]:
-    """
-    Provides a ServingRuntime resource for MLServer using the pre-installed template.
+    """Provides a ServingRuntime, selecting CPU or CUDA template based on params.
+
+    Uses CUDA template only when request.param contains 'gpu': True.
+    Otherwise uses the CPU template regardless of --supported-accelerator-type.
 
     Args:
-        request (pytest.FixtureRequest): Pytest fixture request containing parameters.
-        admin_client (DynamicClient): Kubernetes dynamic client.
-        model_namespace (Namespace): Kubernetes namespace for model deployment.
-        mlserver_runtime_image (str): The container image for the MLServer runtime.
+        request: Pytest fixture request containing deployment parameters.
+        admin_client: Kubernetes dynamic client.
+        model_namespace: Kubernetes namespace for model deployment.
+        supported_accelerator_type: Accelerator type from CLI (e.g., 'nvidia').
+        mlserver_runtime_image: Optional container image override for the runtime.
 
     Yields:
-        ServingRuntime: An instance of the MLServer ServingRuntime configured as per parameters.
+        ServingRuntime: An instance of the MLServer ServingRuntime (CPU or CUDA).
     """
+    params = request.param if hasattr(request, "param") and isinstance(request.param, dict) else {}
+    use_gpu = params.get("gpu", False)
+
+    if use_gpu:
+        accelerator = (supported_accelerator_type or "").lower()
+        template_name = MLSERVER_TEMPLATE_MAP.get(accelerator, RuntimeTemplates.MLSERVER_CUDA)
+        runtime_name = MLSERVER_RUNTIME_NAME_MAP.get(accelerator, ModelInferenceRuntime.MLSERVER_CUDA_RUNTIME)
+    else:
+        template_name = RuntimeTemplates.MLSERVER
+        runtime_name = ModelInferenceRuntime.MLSERVER_RUNTIME
+
     with ServingRuntimeFromTemplate(
         client=admin_client,
-        name=ModelInferenceRuntime.MLSERVER_RUNTIME,
+        name=runtime_name,
         namespace=model_namespace.name,
-        template_name=RuntimeTemplates.MLSERVER,
+        template_name=template_name,
         deployment_type=request.param["deployment_mode"],
         runtime_image=mlserver_runtime_image,
     ) as model_runtime:
@@ -72,19 +90,24 @@ def mlserver_inference_service(
     admin_client: DynamicClient,
     model_namespace: Namespace,
     mlserver_serving_runtime: ServingRuntime,
+    supported_accelerator_type: str | None,
     s3_models_storage_uri: str,
     mlserver_model_service_account: ServiceAccount,
 ) -> Generator[InferenceService]:
-    """
-    Creates and yields a configured InferenceService instance for MLServer testing.
+    """Creates a configured InferenceService with dynamic accelerator-aware GPU resources.
+
+    Supports optional GPU allocation via gpu_count param. When gpu_count > 0 and
+    supported_accelerator_type is set, the appropriate GPU identifier and resource
+    limits are applied to the ISVC container spec.
 
     Args:
-        request (pytest.FixtureRequest): Pytest fixture request containing test parameters.
-        admin_client (DynamicClient): Kubernetes dynamic client.
-        model_namespace (Namespace): Kubernetes namespace for model deployment.
-        mlserver_serving_runtime (ServingRuntime): The MLServer ServingRuntime instance.
-        s3_models_storage_uri (str): URI for the S3 storage location of models.
-        mlserver_model_service_account (ServiceAccount): Service account for the model.
+        request: Pytest fixture request containing test parameters.
+        admin_client: Kubernetes dynamic client.
+        model_namespace: Kubernetes namespace for model deployment.
+        mlserver_serving_runtime: The MLServer ServingRuntime instance.
+        supported_accelerator_type: Accelerator type from CLI (e.g., 'nvidia').
+        s3_models_storage_uri: URI for the S3 storage location of models.
+        mlserver_model_service_account: Service account for the model.
 
     Yields:
         InferenceService: A configured InferenceService resource.
@@ -96,10 +119,13 @@ def mlserver_inference_service(
         "namespace": model_namespace.name,
         "runtime": mlserver_serving_runtime.name,
         "storage_uri": s3_models_storage_uri,
-        "model_format": mlserver_serving_runtime.instance.spec.supportedModelFormats[0].name,
+        "model_format": params.get(
+            "model_format", mlserver_serving_runtime.instance.spec.supportedModelFormats[0].name
+        ),
         "model_service_account": mlserver_model_service_account.name,
         "deployment_mode": params.get("deployment_mode", KServeDeploymentType.STANDARD),
         "external_route": params.get("enable_external_route", False),
+        "enable_auth": params.get("enable_auth", False),
     }
 
     gpu_count = params.get("gpu_count", 0)
@@ -108,18 +134,23 @@ def mlserver_inference_service(
 
     resources = copy.deepcopy(cast(dict[str, dict[str, str]], PREDICT_RESOURCES["resources"]))
     if gpu_count > 0:
-        identifier = Labels.Nvidia.NVIDIA_COM_GPU
+        accelerator = (supported_accelerator_type or "").lower()
+        identifier = MLSERVER_ACCELERATOR_IDENTIFIER.get(accelerator, Labels.Nvidia.NVIDIA_COM_GPU)
         resources["requests"][identifier] = gpu_count
         resources["limits"][identifier] = gpu_count
-        service_config["volumes"] = copy.deepcopy(PREDICT_RESOURCES["volumes"])
-        service_config["volumes_mounts"] = copy.deepcopy(PREDICT_RESOURCES["volume_mounts"])
+
     service_config["resources"] = resources
+    service_config["volumes"] = copy.deepcopy(PREDICT_RESOURCES["volumes"])
+    service_config["volumes_mounts"] = copy.deepcopy(PREDICT_RESOURCES["volume_mounts"])
 
     if timeout:
         service_config["timeout"] = timeout
 
-    if min_replicas:
+    if min_replicas is not None:
         service_config["min_replicas"] = min_replicas
+
+    service_config["wait"] = params.get("wait", True)
+    service_config["wait_for_predictor_pods"] = params.get("wait_for_predictor_pods", True)
 
     with create_isvc(**service_config) as isvc:
         yield isvc
@@ -183,7 +214,7 @@ def mlserver_model_car_inference_service(
         storage_uri=storage_uri,
         model_format=model_format,
         deployment_mode=deployment_mode,
-        external_route=params.get("enable_external_route"),
+        external_route=params.get("enable_external_route", False),
         wait_for_predictor_pods=params.get("wait_for_predictor_pods", False),
         model_env_variables=params.get("model_env_variables"),
     ) as isvc:
--- a/tests/model_serving/model_runtime/mlserver/constant.py
+++ b/tests/model_serving/model_runtime/mlserver/constant.py
@@ -7,7 +7,14 @@ and input queries used across MLServer runtime tests for different frameworks.
 
 from typing import Any
 
-from utilities.constants import KServeDeploymentType, ModelFormat
+from utilities.constants import (
+    AcceleratorType,
+    KServeDeploymentType,
+    Labels,
+    ModelFormat,
+    ModelInferenceRuntime,
+    RuntimeTemplates,
+)
 
 
 class OutputType:
@@ -37,7 +44,7 @@ PREDICT_RESOURCES: dict[str, list[dict[str, str | dict[str, str]]] | dict[str, d
 BASE_RAW_DEPLOYMENT_CONFIG: dict[str, Any] = {
     "deployment_mode": KServeDeploymentType.STANDARD,
     "min-replicas": 1,
-    "enable_external_route": False,
+    "enable_external_route": True,
 }
 
 SKLEARN_REST_INPUT_QUERY: dict[str, Any] = {
@@ -88,6 +95,22 @@ ONNX_REST_INPUT_QUERY = {
     ],
 }
 
+
+_RESNET50_INPUT_SIZE: int = 3 * 224 * 224
+
+ONNX_RESNET50_REST_INPUT_QUERY: dict[str, Any] = {
+    "id": "resnet50",
+    "inputs": [
+        {
+            "name": "pixel_values",
+            "shape": [1, 3, 224, 224],
+            "datatype": "FP32",
+            "data": [0.5] * _RESNET50_INPUT_SIZE,
+        }
+    ],
+}
+
+
 MODEL_CONFIGS: dict[str, dict[str, Any]] = {
     ModelFormat.LIGHTGBM: {
         "model_name": ModelFormat.LIGHTGBM,
@@ -97,9 +120,10 @@ MODEL_CONFIGS: dict[str, dict[str, Any]] = {
     },
     ModelFormat.ONNX: {
         "model_name": ModelFormat.ONNX,
-        "model_version": "v1.0.0",
-        "rest_query": ONNX_REST_INPUT_QUERY,
+        "model_version": "",
+        "rest_query": ONNX_RESNET50_REST_INPUT_QUERY,
         "output_type": OutputType.DETERMINISTIC,
+        "s3_model_dir": "resnet-50-onnx",
     },
     ModelFormat.SKLEARN: {
         "model_name": ModelFormat.SKLEARN,
@@ -114,3 +138,20 @@ MODEL_CONFIGS: dict[str, dict[str, Any]] = {
         "output_type": OutputType.DETERMINISTIC,
     },
 }
+
+
+MLSERVER_TEMPLATE_MAP: dict[str, str] = {
+    AcceleratorType.NVIDIA: RuntimeTemplates.MLSERVER_CUDA,
+}
+
+MLSERVER_ACCELERATOR_IDENTIFIER: dict[str, str] = {
+    AcceleratorType.NVIDIA: Labels.Nvidia.NVIDIA_COM_GPU,
+}
+
+MLSERVER_RUNTIME_NAME_MAP: dict[str, str] = {
+    AcceleratorType.NVIDIA: ModelInferenceRuntime.MLSERVER_CUDA_RUNTIME,
+}
+
+
+RESNET50_INFERENCE_REQUEST_COUNT: int = 10
+GPU_SPEEDUP_THRESHOLD_RATIO: float = 1.5
--- a/tests/model_serving/model_runtime/mlserver/gpu/conftest.py
+++ b/tests/model_serving/model_runtime/mlserver/gpu/conftest.py
@@ -1,0 +1,184 @@
+"""GPU-specific fixtures for MLServer CUDA runtime tests.
+
+Provides fixtures for:
+- GPU accelerator availability skip logic
+- GPU worker node listing
+- RBAC fixtures for inference endpoint authorization
+- Performance test CPU baseline fixtures
+"""
+
+import copy
+from collections.abc import Generator
+from typing import Any, cast
+
+import pytest
+from kubernetes.dynamic import DynamicClient
+from kubernetes.dynamic.exceptions import ResourceNotFoundError
+from ocp_resources.inference_service import InferenceService
+from ocp_resources.namespace import Namespace
+from ocp_resources.node import Node
+from ocp_resources.pod import Pod
+from ocp_resources.role import Role
+from ocp_resources.role_binding import RoleBinding
+from ocp_resources.service_account import ServiceAccount
+from timeout_sampler import TimeoutSampler
+
+from tests.model_serving.model_runtime.mlserver.constant import PREDICT_RESOURCES
+from utilities.constants import (
+    KServeDeploymentType,
+    ModelFormat,
+    ModelInferenceRuntime,
+    RuntimeTemplates,
+    Timeout,
+)
+from utilities.inference_utils import create_isvc
+from utilities.infra import (
+    create_inference_token,
+    create_isvc_view_role,
+    create_ns,
+    get_pods_by_isvc_label,
+    s3_endpoint_secret,
+)
+from utilities.serving_runtime import ServingRuntimeFromTemplate
+
+
+@pytest.fixture(scope="session")
+def skip_if_no_gpu_for_mlserver(supported_accelerator_type: str | None) -> None:
+    """Fail if no GPU accelerator configured for MLServer GPU tests."""
+    if not supported_accelerator_type or supported_accelerator_type.lower() not in {"nvidia"}:
+        pytest.fail(f"MLServer GPU tests require nvidia accelerator, got: {supported_accelerator_type}")
+
+
+@pytest.fixture(scope="session")
+def gpu_worker_nodes(admin_client: DynamicClient) -> list[Node]:
+    """All cluster nodes with nvidia.com/gpu.present label."""
+    return list(Node.get(client=admin_client, label_selector="nvidia.com/gpu.present=true"))
+
+
+@pytest.fixture(scope="class")
+def mlserver_gpu_inference_view_role(
+    admin_client: DynamicClient,
+    mlserver_inference_service: InferenceService,
+) -> Generator[Role]:
+    """Creates a Role granting view access to the GPU InferenceService."""
+    with create_isvc_view_role(
+        client=admin_client,
+        isvc=mlserver_inference_service,
+        name=f"{mlserver_inference_service.name}-view",
+        resource_names=[mlserver_inference_service.name],
+    ) as role:
+        yield role
+
+
+@pytest.fixture(scope="class")
+def authorized_inference_role_binding(
+    admin_client: DynamicClient,
+    model_namespace: Namespace,
+    mlserver_model_service_account: ServiceAccount,
+    mlserver_gpu_inference_view_role: Role,
+) -> Generator[RoleBinding]:
+    """Creates a RoleBinding granting the SA inference access."""
+    with RoleBinding(
+        client=admin_client,
+        namespace=model_namespace.name,
+        name="mlserver-gpu-inference-rb",
+        role_ref_name=mlserver_gpu_inference_view_role.name,
+        role_ref_kind=mlserver_gpu_inference_view_role.kind,
+        subjects_kind=mlserver_model_service_account.kind,
+        subjects_name=mlserver_model_service_account.name,
+    ) as rb:
+        yield rb
+
+
+@pytest.fixture(scope="class")
+def authorized_inference_token(
+    mlserver_model_service_account: ServiceAccount,
+    authorized_inference_role_binding: RoleBinding,
+) -> str:
+    """Returns bearer token for the authorized service account."""
+    return create_inference_token(model_service_account=mlserver_model_service_account)
+
+
+@pytest.fixture(scope="class")
+def mlserver_pending_predictor_pods(
+    admin_client: DynamicClient,
+    mlserver_inference_service: InferenceService,
+) -> list[Pod]:
+    """Waits up to 2 minutes for predictor pods to appear (may remain Pending)."""
+    predictor_pods: list[Pod] = []
+    for sample in TimeoutSampler(
+        wait_timeout=Timeout.TIMEOUT_2MIN,
+        sleep=10,
+        exceptions_dict={ResourceNotFoundError: []},
+        func=get_pods_by_isvc_label,
+        client=admin_client,
+        isvc=mlserver_inference_service,
+    ):
+        if sample:
+            predictor_pods = sample
+            break
+    return predictor_pods
+
+
+@pytest.fixture(scope="class")
+def mlserver_cpu_perf_namespace(admin_client: DynamicClient) -> Generator[Namespace]:
+    """Dedicated namespace for CPU baseline ISVC in performance tests."""
+    with create_ns(admin_client=admin_client, name="mlserver-perf-cpu") as namespace:
+        yield namespace
+
+
+@pytest.fixture(scope="class")
+def mlserver_cpu_perf_isvc(
+    admin_client: DynamicClient,
+    mlserver_cpu_perf_namespace: Namespace,
+    mlserver_runtime_image: str | None,
+    s3_models_storage_uri: str,
+    aws_access_key_id: str,
+    aws_secret_access_key: str,
+    models_s3_bucket_name: str,
+    models_s3_bucket_region: str,
+    models_s3_bucket_endpoint: str,
+) -> Generator[InferenceService]:
+    """CPU MLServer InferenceService in its own namespace for perf baseline."""
+    with (
+        s3_endpoint_secret(
+            client=admin_client,
+            name="mlserver-perf-cpu-s3",
+            namespace=mlserver_cpu_perf_namespace.name,
+            aws_access_key=aws_access_key_id,
+            aws_secret_access_key=aws_secret_access_key,
+            aws_s3_bucket=models_s3_bucket_name,
+            aws_s3_region=models_s3_bucket_region,
+            aws_s3_endpoint=models_s3_bucket_endpoint,
+        ) as cpu_s3_secret,
+        ServiceAccount(
+            client=admin_client,
+            namespace=mlserver_cpu_perf_namespace.name,
+            name="mlserver-perf-cpu-sa",
+            secrets=[{"name": cpu_s3_secret.name}],
+        ) as service_account,
+        ServingRuntimeFromTemplate(
+            client=admin_client,
+            name=ModelInferenceRuntime.MLSERVER_RUNTIME,
+            namespace=mlserver_cpu_perf_namespace.name,
+            template_name=RuntimeTemplates.MLSERVER,
+            deployment_type=KServeDeploymentType.STANDARD,
+            runtime_image=mlserver_runtime_image,
+        ) as cpu_runtime,
+        create_isvc(
+            client=admin_client,
+            name="resnet-50-onnx",
+            namespace=mlserver_cpu_perf_namespace.name,
+            runtime=cpu_runtime.name,
+            storage_uri=s3_models_storage_uri,
+            model_format=ModelFormat.ONNX,
+            model_service_account=service_account.name,
+            deployment_mode=KServeDeploymentType.STANDARD,
+            external_route=True,
+            resources=copy.deepcopy(cast(dict[str, Any], PREDICT_RESOURCES["resources"])),
+            volumes=copy.deepcopy(PREDICT_RESOURCES["volumes"]),
+            volumes_mounts=copy.deepcopy(PREDICT_RESOURCES["volume_mounts"]),
+            timeout=Timeout.TIMEOUT_10MIN,
+        ) as isvc,
+    ):
+        yield isvc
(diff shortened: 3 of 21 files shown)


tests/model_serving/model_runtime/mlserver/conftest.py
```diff
@@ -41,25 +44,40 @@ def mlserver_serving_runtime(
...
     with ServingRuntimeFromTemplate(
         client=admin_client,
-        name=ModelInferenceRuntime.MLSERVER_RUNTIME,
+        name=runtime_name,
         namespace=model_namespace.name,
-        template_name=RuntimeTemplates.MLSERVER,
+        template_name=template_name,
         deployment_type=request.param["deployment_mode"],
         runtime_image=mlserver_runtime_image,
     ) as model_runtime:
```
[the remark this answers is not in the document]
Snomaan6846: The defensive guard on line 65 exists because "gpu" is an optional key that CPU tests don't pass — it's not protecting against request.param being absent. This fixture is always invoked via indirect=True parametrize, so request.param is guaranteed to be a dict.

deployment_mode is a required field — every caller must provide it. A KeyError here is intentional; it surfaces misconfigured parametrize blocks immediately at collection time rather than silently propagating None into ServingRuntimeFromTemplate where it would fail with a less obvious error downstream.

Will simplify the guard to use_gpu = request.param.get("gpu", False) to remove the misleading hasattr check, but keeping request.param["deployment_mode"] as a hard access.

tests/model_serving/model_runtime/mlserver/constant.py
```diff
@@ -37,7 +44,7 @@ class OutputType:
 BASE_RAW_DEPLOYMENT_CONFIG: dict[str, Any] = {
     "deployment_mode": KServeDeploymentType.STANDARD,
     "min-replicas": 1,
-    "enable_external_route": False,
+    "enable_external_route": True,
 }
```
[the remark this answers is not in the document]
Snomaan6846: This is intentional. run_mlserver_inference was refactored to prefer external routes (with port-forward as fallback), and S3 tests need external routes for reliable E2E validation. The enable_auth flag controls authentication independently — external route ≠ unauthenticated access.

These are ephemeral test namespaces on dedicated QE clusters, not production deployments. The CWE-284 classification doesn't apply to a test fixture that creates and destroys resources within a single pytest class scope.

tests/model_serving/model_runtime/mlserver/probes/conftest.py
```diff
@@ -64,7 +64,7 @@ def mlserver_probes_inference_service(
         "model_format": supported_formats[0].name,
         "model_service_account": mlserver_model_service_account.name,
         "deployment_mode": request.param.get("deployment_mode", KServeDeploymentType.STANDARD),
-        "external_route": True,
+        "external_route": request.param.get("enable_external_route", False),
```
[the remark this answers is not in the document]
Snomaan6846: The False default is never reached. The probes test parametrizes with **BASE_RAW_DEPLOYMENT_CONFIG which contains "enable_external_route": True — so request.param.get("enable_external_route", False) evaluates to True. Behavior is unchanged; the fixture now reads from params consistently rather than hardcoding.

tests/model_serving/model_runtime/mlserver/pvc/test_mlserver_pvc_inference.py
```diff
@@ -32,7 +31,7 @@
             {"pvc-size": "5Gi"},
             {"model-dir": "mlserver/model_repository/sklearn"},
             {"name": ModelFormat.SKLEARN, "deployment_mode": KServeDeploymentType.STANDARD},
-            {"name": "sklearn", "deployment_mode": KServeDeploymentType.STANDARD},
+            {"name": "sklearn", "deployment_mode": KServeDeploymentType.STANDARD, "enable_external_route": True},
```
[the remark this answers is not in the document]
Snomaan6846: This is an OpenShift-only test suite — all target clusters have a Router, so external routes are always available. The flag was added intentionally to use the reliable external-route inference path instead of port-forward.

Also, run_mlserver_inference does not raise ValueError when a route is unavailable — it checks Inference.visibility_exposed and falls back to port-forward with a warning log. The ValueError only fires for unsupported protocols, not missing routes.

tests/model_serving/model_runtime/mlserver/upgrade/conftest.py
```diff
@@ -0,0 +1,249 @@
...
+        protocol=Protocols.REST,
+    )
+    validate_deterministic_snapshot(response=baseline_response)
+
+    restart_counts = get_restart_counts(pod=predictor_pods[0])
+    cm = ConfigMap(
+        client=admin_client,
+        name=UPGRADE_BASELINE_CM,
+        namespace=mlserver_upgrade_namespace.name,
+        data={UPGRADE_RESTART_KEY: json.dumps(restart_counts)},
```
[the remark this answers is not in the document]
Snomaan6846: The predictor pod is owned by the ServingRuntime instance bound to this ISVC. During RHOAI upgrade, the operator updates its own components (controller, dashboard, etc.) — it does NOT modify or roll user-created ServingRuntimes or their predictor pods unless the runtime image/spec is explicitly changed.

So the pod UID will be the same pre and post upgrade. The restart count comparison is valid as-is: same pod, same UID, any additional restarts indicate instability caused by the upgrade (e.g., operator reconciliation thrashing, webhook failures, or sidecar restarts).

Pod replacement only happens if someone updates the ServingRuntime image or ISVC spec, which this test intentionally avoids. No UID tracking needed.

utilities/constants.py
```diff
@@ -319,7 +321,7 @@ class ModelCarImage:
     MLSERVER_XGBOOST: str = "oci://quay.io/jooholee/mlserver-xgboost@sha256:5b6982bdc939b53a7a1210f56aa52bf7de0f0cbc693668db3fd1f496571bff29"  # noqa: E501
     MLSERVER_LIGHTGBM: str = "oci://quay.io/jooholee/mlserver-lightgbm@sha256:77eb15a2eccefa3756faaf2ee4bc1e63990b746427d323957c461f33a4f1a6a3"  # noqa: E501
     MLSERVER_ONNX: str = (
-        "oci://quay.io/jooholee/mlserver-onnx@sha256:d0ad00fb6f2caa8f02a0250fc44a576771d0846b2ac8d164ec203b10ec5d604b"  # noqa: E501
+        "oci://quay.io/syedali/mlserver-onnx@sha256:1724ae50e1178a11c3b8dd3c65c03e85d3f416e5994c80c63bcc556c71189e9d"  # noqa: E501
     )
```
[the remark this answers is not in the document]
Snomaan6846: Intentional — this is a temporary image reference for development. A follow-up activity is planned to move all MLServer model-car images to an org-level quay repo (quay.io/opendatahub/) for shared access rather than individual accounts. Will address the image_constants.py placement at that time.

tests/model_serving/model_runtime/mlserver/upgrade/conftest.py
```diff
@@ -0,0 +1,249 @@
...
+    pytestconfig: pytest.Config,
+    admin_client: DynamicClient,
+    teardown_resources: bool,
+) -> Generator[Namespace]:
+    """Namespace for MLServer upgrade tests — persists through upgrade."""
+    ns = Namespace(client=admin_client, name=UPGRADE_NAMESPACE)
+
+    if pytestconfig.option.post_upgrade:
+        yield ns
+        ns.clean_up()
```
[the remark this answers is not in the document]
Snomaan6846: Good catch — accepted. Will guard with if teardown_resources: for consistency with the pre-upgrade branch.

tests/model_serving/model_runtime/mlserver/pvc/test_mlserver_pvc_inference.py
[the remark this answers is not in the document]
Snomaan6846: Acknowledged — get_exposed_isvc_url does raise ValueError if status.url is missing, so the earlier claim about "no ValueError" was imprecise.

However, this is unreachable in practice: the ISVC fixture has wait=True (default), which blocks until the ISVC reaches Ready state. KServe populates status.url as part of the Ready condition when external_route=True. The fixture would timeout during creation before any test body executes if the route failed to provision.

Since this is an OpenShift-only test suite (Router always available), enable_external_route: True is intentional to use the reliable inference path. No change needed.

utilities/constants.py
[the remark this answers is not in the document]
Snomaan6846: ok create it

tests/model_serving/model_runtime/mlserver/gpu/test_mlserver_gpu_rbac.py
```diff
@@ -0,0 +1,103 @@
...
+            f"got {no_token_response.status_code}: {no_token_response.text[:200]}"
+        )
+
+        invalid_token_response = requests.post(
+            url=infer_url,
+            json=_minimal_payload,
+            headers={"Authorization": "Bearer invalid-token-rbac-check"},
+            verify=False,
+            timeout=60,
+        )
```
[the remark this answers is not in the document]
Snomaan6846: verify=False is the standard pattern across this entire test repo — every requests call to cluster-internal routes uses it. OpenShift test clusters use self-signed certificates on Routes, and the cluster CA is not distributed to test runners.

The token is a short-lived ServiceAccount token scoped to a single ephemeral test namespace that is torn down after the test class completes. There is no credential leakage risk in this context.

This is consistent with every other test in opendatahub-tests (grep for verify=False — hundreds of occurrences). Changing this one test to verify=True would require injecting the cluster CA bundle, which is out of scope and would be a repo-wide architectural decision, not a per-test fix.

tests/model_serving/model_runtime/mlserver/gpu/test_mlserver_gpu_serving.py
```diff
@@ -0,0 +1,174 @@
...
+        """
+        base_url = get_exposed_isvc_url(isvc=mlserver_inference_service)
+        model_name = mlserver_inference_service.name
+
+        readiness_url = f"{base_url}/v2/models/{model_name}/ready"
+        readiness_response = requests.get(url=readiness_url, verify=False, timeout=60)
+        readiness_response.raise_for_status()
+
+        health_url = f"{base_url}/v2/health/ready"
+        health_response = requests.get(url=health_url, verify=False, timeout=60)
```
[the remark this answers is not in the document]
Snomaan6846: Same answer as the RBAC comment — verify=False is the repo-wide standard for all requests to cluster-internal OpenShift Routes with self-signed certs. This is not a per-test decision; it's architectural. Hundreds of occurrences across the repo.

RH-steve-grubb (approved):
/lgtm

tests/model_serving/model_runtime/mlserver/gpu/test_mlserver_gpu_build.py
```diff
@@ -0,0 +1,129 @@
...
+from ocp_resources.template import Template
+from pytest_testconfig import config as py_config
+
+from utilities.constants import RHOAI_OPERATOR_NAMESPACE, Containers, RuntimeTemplates
+from utilities.operator_utils import get_cluster_service_version
+from utilities.serving_runtime import get_runtime_image_from_template
+
+pytestmark = [
+    pytest.mark.usefixtures("skip_if_no_gpu_for_mlserver"),
+    pytest.mark.gpu,
```
kpunwatk, line 21: Hi @Snomaan6846, Could you check with the DevTestOps team about this marker? We noticed that the AI Safety tests weren't being picked up in the GPU runs, so we updated our tests to use the following markers instead:

* `@pytest.mark.vllm_nvidia_single_gpu`
* `@pytest.mark.vllm_amd_gpu`

It might be worth confirming whether pytest.mark.gpu is the expected marker for the GPU test pipelines.

tests/model_serving/model_runtime/mlserver/gpu/test_mlserver_gpu_fallback.py
```diff
@@ -0,0 +1,209 @@
...
+            },
+            id="test_mlserver_cuda_silent_cpu_fallback",
+            marks=[pytest.mark.tier1],
+        ),
+    ],
+    indirect=True,
+)
+@pytest.mark.usefixtures("skip_if_no_gpu_for_mlserver")
+@pytest.mark.gpu
+@pytest.mark.model_server_gpu
```
kpunwatk: here as above you can use the module-level pytestmark list

tests/model_serving/model_runtime/mlserver/gpu/test_mlserver_gpu_fallback.py
```diff
@@ -0,0 +1,209 @@
...
+_ONNX_MODEL_CONFIG: dict[str, Any] = MODEL_CONFIGS[ModelFormat.ONNX]
+
+
+def _collect_predictor_pods(client: DynamicClient, isvc: InferenceService) -> list[Pod]:
+    """Return predictor pods without raising on absence."""
+    try:
+        return get_pods_by_isvc_label(client=client, isvc=isvc)
+    except ResourceNotFoundError:
+        return []
+
```
kpunwatk: I suggest moving this into a fixture in `conftest.py` or extracting it into a helper function under `utils`. That would keep the test file focused on the test logic and test data, while improving reusability and maintainability.

kpunwatk (changes requested):

tests/model_serving/model_runtime/mlserver/gpu/test_mlserver_gpu_build.py
Snomaan6846: Hi @kpunwatk ,

Thanks for the comment, I was thinking to check with DevTestOps team for adding below QG mapping for Model Runtimes component.
- mlserver-nvidia-gpu: "--supported-accelerator-type Nvidia -m 'mlserver_nvidia_gpu'"
- pre-upgrade: "--junit-xml=results/xunit_pre_upgrade_report.xml --pre-upgrade"
- post-upgrade: "--junit-xml=results/xunit_post_upgrade_report.xml --post-upgrade"

tests/model_serving/model_runtime/mlserver/gpu/test_mlserver_gpu_fallback.py
Snomaan6846: Can't use module-level pytestmark here — this file has two classes with opposing requirements:

- TestSilentCPUFallback needs skip_if_no_gpu_for_mlserver (requires GPU cluster)
- TestGPUSchedulingFailure must NOT have it — it runs on CPU-only clusters and has its own _skip_if_gpu_nodes_available that skips when GPUs are present

Moving to module-level would incorrectly skip TestGPUSchedulingFailure on non-GPU clusters.

tests/model_serving/model_runtime/mlserver/gpu/test_mlserver_gpu_fallback.py
Snomaan6846: Agreed — removed the helper entirely. Replaced with TimeoutSampler's built-in exceptions_dict={ResourceNotFoundError: []} parameter, which is the standard repo pattern (same as utilities/infra.py:766-773 and utilities/infra.py:994). No wrapper function needed.

tests/model_serving/model_runtime/mlserver/gpu/test_mlserver_gpu_perf.py
```diff
@@ -0,0 +1,93 @@
...
+            meets_threshold=speedup_ratio >= GPU_SPEEDUP_THRESHOLD_RATIO,
+        )
+
+        if speedup_ratio < GPU_SPEEDUP_THRESHOLD_RATIO:
+            warnings.warn(
+                f"GPU speedup ({speedup_ratio:.2f}x) is below the target {GPU_SPEEDUP_THRESHOLD_RATIO}x. "
+                f"CPU: {cpu_avg_latency:.3f}s, GPU: {gpu_avg_latency:.3f}s. "
+                "This may be due to network latency dominating compute time.",
+                stacklevel=1,
+            )
```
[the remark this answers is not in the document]
Snomaan6846: Intentional design decision. Performance tests running from remote workstations or CI runners introduce variable network latency that dominates the actual GPU compute speedup (we observed 1.15x on some runs due to network overhead vs 3-5x on local cluster access).

A hard assertion would make this test flaky in CI — the test validates that the full GPU inference pipeline is functional (deploy, route, inference), and the performance measurement is observational data logged for analysis. The warnings.warn ensures it's surfaced in test output without causing false failures.

If we later add dedicated bare-metal GPU CI nodes with local cluster access, we can promote this to a hard assertion. Until then, a flaky perf gate provides negative value.

tests/model_serving/model_runtime/mlserver/upgrade/conftest.py
```diff
@@ -0,0 +1,250 @@
...
+            ns.clean_up()
+    else:
+        with create_ns(
+            admin_client=admin_client,
+            name=UPGRADE_NAMESPACE,
+            model_mesh_enabled=False,
+            add_dashboard_label=True,
+            teardown=teardown_resources,
+        ) as ns:
+            yield ns
```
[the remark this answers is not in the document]
Snomaan6846: teardown_resources is a session-scoped fixture that returns False during upgrade CI pipeline runs — resources persist between pre and post phases as intended. This is the standard repo pattern used across all upgrade test suites (see tests/model_serving/model_server/upgrade/conftest.py).

Hardcoding teardown=False would prevent cleanup in non-upgrade development runs where someone tests fixtures independently.

tests/model_serving/model_runtime/mlserver/utils.py
```diff
@@ -44,40 +49,50 @@ def send_rest_request(url: str, input_data: dict[str, Any], verify: bool = False
...
-        raise ValueError(f"Unsupported deployment mode: {deployment_mode}. Supported modes: {supported_modes}")
+    structlog.get_logger(name=__name__).warning(
+        event="ISVC has no external route, falling back to port-forward",
+        model_name=model_name,
+    )
+
+    pods = get_pods_by_isvc_label(client=isvc.client, isvc=isvc)
+    if not pods:
+        raise RuntimeError(f"No predictor pods found for InferenceService '{model_name}'")
+    pod_name = pods[0].name
```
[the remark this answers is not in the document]
Snomaan6846: Addressed.

kpunwatk:
pre-commit check seems to be failing

tests/model_serving/model_runtime/mlserver/conftest.py
```diff
@@ -108,19 +134,24 @@ def mlserver_inference_service(
...
+
     service_config["resources"] = resources
+    service_config["volumes"] = copy.deepcopy(PREDICT_RESOURCES["volumes"])
+    service_config["volumes_mounts"] = copy.deepcopy(PREDICT_RESOURCES["volume_mounts"])
 
     if timeout:
         service_config["timeout"] = timeout
 
     if min_replicas:
         service_config["min_replicas"] = min_replicas
```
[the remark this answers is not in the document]
Snomaan6846: Good catch — accepted. min_replicas=0 is valid for scale-to-zero and would be dropped by truthiness check.

tests/model_serving/model_runtime/mlserver/gpu/test_mlserver_gpu_serving.py
```diff
@@ -0,0 +1,174 @@
...
+
+        pod_containers = mlserver_pod_resource.instance.spec.containers
+        assert pod_containers, "Predictor pod has no containers in spec"
+        kserve_container = next(
+            (container for container in pod_containers if "kserve" in (container.name or "")),
+            pod_containers[0],
+        )
+        pod_gpu_limit = (kserve_container.resources.limits or {}).get(Labels.Nvidia.NVIDIA_COM_GPU)
+        assert pod_gpu_limit is not None, f"Container '{kserve_container.name}' has no nvidia.com/gpu limit"
+        assert str(pod_gpu_limit) == "1", f"Expected nvidia.com/gpu limit of '1', got: {pod_gpu_limit!r}"
```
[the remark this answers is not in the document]
Snomaan6846: Good catch — accepted. Silent fallback to the first container could assert GPU limits against a sidecar. Will fail explicitly if kserve container is not found.

tests/model_serving/model_runtime/mlserver/utils.py
```diff
@@ -44,40 +50,51 @@ def send_rest_request(url: str, input_data: dict[str, Any], verify: bool = False
...
+
     model_name = isvc.instance.metadata.name
     version_suffix = f"/versions/{model_version}" if model_version else ""
     rest_endpoint = f"/v2/models/{model_name}{version_suffix}/infer"
 
-    if protocol != Protocols.REST:
-        raise ValueError(f"Unsupported protocol: {protocol}. Only REST is supported.")
+    if Inference(inference_service=isvc).visibility_exposed:
+        base_url = get_exposed_isvc_url(isvc=isvc)
+        return send_rest_request(url=f"{base_url}{rest_endpoint}", input_data=input_data, verify=False)
```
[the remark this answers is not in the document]
Snomaan6846: Same as previous verify=False comments — repo-wide standard for OpenShift cluster-internal routes with self-signed certs. See earlier discussion thread.

utilities/constants.py
```diff
@@ -319,7 +321,7 @@ class ModelCarImage:
     MLSERVER_XGBOOST: str = "oci://quay.io/jooholee/mlserver-xgboost@sha256:5b6982bdc939b53a7a1210f56aa52bf7de0f0cbc693668db3fd1f496571bff29"  # noqa: E501
     MLSERVER_LIGHTGBM: str = "oci://quay.io/jooholee/mlserver-lightgbm@sha256:77eb15a2eccefa3756faaf2ee4bc1e63990b746427d323957c461f33a4f1a6a3"  # noqa: E501
     MLSERVER_ONNX: str = (
-        "oci://quay.io/jooholee/mlserver-onnx@sha256:d0ad00fb6f2caa8f02a0250fc44a576771d0846b2ac8d164ec203b10ec5d604b"  # noqa: E501
+        "oci://quay.io/syedali/mlserver-onnx@sha256:1724ae50e1178a11c3b8dd3c65c03e85d3f416e5994c80c63bcc556c71189e9d"  # noqa: E501
```
sheltoncyril: @Snomaan6846 if this image is used in disconnected tests, could you please move it to utilities/image_constants.py?

tests/model_serving/model_runtime/mlserver/gpu/conftest.py
```diff
@@ -0,0 +1,157 @@
...
+from utilities.inference_utils import create_isvc
+from utilities.infra import create_inference_token, create_isvc_view_role, create_ns, s3_endpoint_secret
+from utilities.serving_runtime import ServingRuntimeFromTemplate
+
+
+@pytest.fixture(scope="session")
+def skip_if_no_gpu_for_mlserver(supported_accelerator_type: str | None) -> None:
+    """Skip if no GPU accelerator configured for MLServer GPU tests."""
+    if not supported_accelerator_type or supported_accelerator_type.lower() not in {"nvidia"}:
+        pytest.skip(f"MLServer GPU tests require nvidia accelerator, got: {supported_accelerator_type}")
```
dbasunag: Please fail here

tests/model_serving/model_runtime/mlserver/gpu/conftest.py
```diff
@@ -0,0 +1,157 @@
...
+def authorized_inference_token(
+    mlserver_model_service_account: ServiceAccount,
+    authorized_inference_role_binding: RoleBinding,
+) -> str:
+    """Returns bearer token for the authorized service account."""
+    return create_inference_token(model_service_account=mlserver_model_service_account)
+
+
+# --- Performance test fixtures ---
+
```
dbasunag, line 121: Please remove these comments.

tests/model_serving/model_runtime/mlserver/gpu/test_mlserver_gpu_build.py
```diff
@@ -0,0 +1,129 @@
...
+        And the MLServer and MLServer-CUDA template images match the respective CSV entries.
+        """
+        csv = get_cluster_service_version(
+            client=admin_client,
+            prefix="rhods-operator",
+            namespace=RHOAI_OPERATOR_NAMESPACE,
+        )
+
+        csv_dict = csv.instance.to_dict()
+        related_images = csv_dict.get("spec", {}).get("relatedImages", []) or []
```
dbasunag: csv.instance.spec.relatedImages, can you you direct access.

tests/model_serving/model_runtime/mlserver/gpu/test_mlserver_gpu_build.py
```diff
@@ -0,0 +1,129 @@
...
+
+        mlserver_template_image: str = get_runtime_image_from_template(
+            client=admin_client,
+            template_name=RuntimeTemplates.MLSERVER,
+            namespace=applications_namespace,
+        )
+        assert mlserver_template_image == cpu_image, (
+            f"MLServer template image '{mlserver_template_image}' does not match "
+            f"CSV relatedImages '{_ODH_MLSERVER_IMAGE_NAME}' entry: '{cpu_image}'"
+        )
```
dbasunag: Please simplify

tests/model_serving/model_runtime/mlserver/gpu/test_mlserver_gpu_fallback.py
```diff
@@ -0,0 +1,202 @@
...
+    """
+
+    @pytest.fixture(autouse=True)
+    def _skip_if_gpu_nodes_available(self, gpu_count_on_cluster: int) -> None:
+        """Skip this test if the cluster has GPUs — can't test scheduling failure with GPUs present."""
+        if gpu_count_on_cluster > 0:
+            pytest.skip(
+                f"Cluster has {gpu_count_on_cluster} GPU(s) — cannot test GPU scheduling failure. "
+                "Run on a CPU-only cluster."
+            )
```
dbasunag, line 137: We should fail, if the test is not marked for gpu and is still running there.

tests/model_serving/model_runtime/mlserver/gpu/test_mlserver_gpu_fallback.py
```diff
@@ -0,0 +1,202 @@
...
+                exceptions_dict={ResourceNotFoundError: []},
+                func=get_pods_by_isvc_label,
+                client=admin_client,
+                isvc=mlserver_inference_service,
+            ):
+                if sample:
+                    predictor_pods = sample
+                    break
+        except TimeoutExpiredError:
+            pass
```
dbasunag: Please create a fixture

tests/model_serving/model_runtime/mlserver/gpu/test_mlserver_gpu_fallback.py
```diff
@@ -0,0 +1,202 @@
...
+            f"Expected predictor pod Pending due to GPU scheduling failure but got {pod_phase!r}"
+        )
+
+        scheduling_events = Event.list(
+            client=admin_client,
+            namespace=mlserver_inference_service.namespace,
+            field_selector=f"involvedObject.name={predictor_pod.name}",
+        )
+        if scheduling_events:
+            structlog.get_logger(name=__name__).info(
```
dbasunag: For logging please use existing patterns.

tests/model_serving/model_runtime/mlserver/gpu/test_mlserver_gpu_perf.py
```diff
@@ -0,0 +1,93 @@
...
+        )
+        gpu_avg_latency = measure_average_latency(
+            isvc=mlserver_inference_service,
+            input_data=ONNX_RESNET50_REST_INPUT_QUERY,
+            request_count=RESNET50_INFERENCE_REQUEST_COUNT,
+        )
+
+        speedup_ratio = cpu_avg_latency / gpu_avg_latency
+
+        structlog.get_logger(name=__name__).info(
```
dbasunag: Same comment. Please follow existing coding patterns

tests/model_serving/model_runtime/mlserver/gpu/test_mlserver_gpu_rbac.py
```diff
@@ -0,0 +1,103 @@
...
+        no_token_response = requests.post(
+            url=infer_url,
+            json=_minimal_payload,
+            verify=False,
+            timeout=60,
+        )
+        assert no_token_response.status_code in (HTTPStatus.UNAUTHORIZED, HTTPStatus.FORBIDDEN), (
+            f"Expected HTTP 401 or 403 for unauthenticated request, "
+            f"got {no_token_response.status_code}: {no_token_response.text[:200]}"
+        )
```
dbasunag: Seems like a separate test.

tests/model_serving/model_runtime/mlserver/upgrade/conftest.py
```diff
@@ -0,0 +1,250 @@
...
+) -> Generator[ServiceAccount]:
+    """Service account for MLServer upgrade tests."""
+    sa = ServiceAccount(
+        client=admin_client,
+        name=UPGRADE_SA_NAME,
+        namespace=mlserver_upgrade_namespace.name,
+    )
+
+    if pytestconfig.option.post_upgrade:
+        yield sa
```
dbasunag, line 125: Should it not get cleaned up after post upgrade?

tests/model_serving/model_runtime/mlserver/upgrade/conftest.py
```diff
@@ -0,0 +1,250 @@
...
+    teardown_resources: bool,
+) -> Generator[ServingRuntime]:
+    """CPU MLServer ServingRuntime for upgrade tests."""
+    runtime = ServingRuntime(
+        client=admin_client,
+        name=ModelInferenceRuntime.MLSERVER_RUNTIME,
+        namespace=mlserver_upgrade_namespace.name,
+    )
+
+    if pytestconfig.option.post_upgrade:
```
dbasunag, line 153: ditto

tests/model_serving/model_runtime/mlserver/upgrade/conftest.py
```diff
@@ -0,0 +1,250 @@
...
+    """CPU MLServer InferenceService for upgrade tests."""
+    isvc_kwargs = {
+        "client": admin_client,
+        "name": UPGRADE_ISVC_NAME,
+        "namespace": mlserver_upgrade_namespace.name,
+    }
+
+    isvc = InferenceService(**isvc_kwargs)
+
+    if pytestconfig.option.post_upgrade:
```
dbasunag, line 188: ditto

dbasunag (changes requested):

tests/model_serving/model_runtime/mlserver/gpu/conftest.py
Snomaan6846: pytest.skip() is correct here — this is a hardware precondition gate, not a failure condition. The established repo pattern uses pytest.skip() for missing accelerators (see vllm/conftest.py:45, vllm/cpu/ibm_power_z/conftest.py:38). pytest.fail() is reserved for cases where something should exist on the cluster but doesn't (e.g., vllm/fast/conftest.py:28 when a template is missing).

Practically: the tier1 quality gate runs -m tier1, which collects these GPU tests on non-GPU clusters. They must be gracefully skipped — pytest.fail() would break every tier1 run on CPU-only clusters.

tests/model_serving/model_runtime/mlserver/gpu/conftest.py
Snomaan6846: Addressed.

tests/model_serving/model_runtime/mlserver/gpu/test_mlserver_gpu_build.py
Snomaan6846: Addressed.

tests/model_serving/model_runtime/mlserver/gpu/test_mlserver_gpu_build.py
Snomaan6846: Addressed.

tests/model_serving/model_runtime/mlserver/gpu/test_mlserver_gpu_fallback.py
Snomaan6846: This test is a negative scenario — it validates Pending behavior when GPUs are unavailable. It's intentionally NOT marked with @pytest.mark.mlserver_nvidia_gpu or @pytest.mark.gpu (unlike TestSilentCPUFallback above it in the same file).

It gets collected via -m tier1 and gracefully skips on GPU clusters because the test premise (no GPUs available) doesn't hold. This is a hardware precondition, not a misconfiguration — same reasoning as skip_if_no_gpu_for_mlserver but inverted.

tests/model_serving/model_runtime/mlserver/gpu/test_mlserver_gpu_fallback.py
Snomaan6846: Addressed.

tests/model_serving/model_runtime/mlserver/gpu/test_mlserver_gpu_fallback.py
Snomaan6846: Addressed.

tests/model_serving/model_runtime/mlserver/gpu/test_mlserver_gpu_perf.py
Snomaan6846: Addressed.

tests/model_serving/model_runtime/mlserver/gpu/test_mlserver_gpu_rbac.py
Snomaan6846: Done. Split into 3 separate test methods:

- test_authorized_inference — valid token, expects 200
- test_unauthenticated_rejected — no token, expects 401/403
- test_invalid_token_rejected — bad token, expects 401/403

All share the same class-scoped ISVC deployment (one deployment, three independent assertions).

tests/model_serving/model_runtime/mlserver/upgrade/conftest.py
Snomaan6846: The SA lives inside the upgrade namespace — it's automatically deleted when the namespace fixture runs ns.clean_up() in the post-upgrade teardown. Explicit SA cleanup would be redundant and could race with the namespace deletion. Same applies to the secret, runtime, and ISVC — they're all namespace-scoped resources cleaned up by the namespace teardown.

tests/model_serving/model_runtime/mlserver/upgrade/conftest.py
Snomaan6846: Same as above — namespace-scoped resource, cleaned up by the namespace teardown. No explicit per-resource cleanup needed.

tests/model_serving/model_runtime/mlserver/upgrade/conftest.py
Snomaan6846: Same — namespace-scoped, cleaned up by namespace teardown.

utilities/constants.py
Snomaan6846: Addressed.

tests/model_serving/model_runtime/mlserver/gpu/conftest.py
dbasunag: For 3.5 any skip is supposed to be treated like a blocker. So please don't introduce a new skip. If we are running tests against not matching hardware, that should be deemed as a testops bug and addressed, instead of silently skipping.

tests/model_serving/model_runtime/mlserver/gpu/test_mlserver_gpu_fallback.py
```diff
@@ -0,0 +1,187 @@
...
+            marks=[pytest.mark.tier1],
+        ),
+    ],
+    indirect=True,
+)
+class TestGPUSchedulingFailure:
+    """TC-FALLBACK-002: GPU resource request stays Pending when no GPU nodes are available.
+
+    This test is only meaningful on clusters WITHOUT GPU nodes. It is automatically
+    skipped when GPU nodes are detected — use a CPU-only cluster to exercise this path.
```
dbasunag: This is problematic. We should not add silent skips. Instead if this test is running on gpu, even though we are not marking it with gpu marker, we need to understand why that is happening.

tests/model_serving/model_runtime/mlserver/gpu/test_mlserver_gpu_fallback.py
dbasunag: Even then we should fail instead of silently skipping. Also unless you are using gpu marker, the tests won't be running against gpu clusters - if they do, then pytest marking is broken or test is wrongly marked.

tests/model_serving/model_runtime/mlserver/upgrade/conftest.py
dbasunag: This is not how we handle test resource clean up. Please add explicit clean up for resources your test is creating.

dbasunag (changes requested):

tests/model_serving/model_runtime/mlserver/upgrade/conftest.py
dbasunag: Same response. Since your pre-upgrade is creating the resource, we would need explicit resource clean up in teardown of post-upgrade. Please check existing upgrade tests, to see existing code styles and expectations.

tests/model_serving/model_runtime/mlserver/upgrade/conftest.py
dbasunag: Same response. Please add explicit clean up.

dbasunag (changes requested):

tests/model_serving/model_runtime/mlserver/gpu/conftest.py
```diff
@@ -0,0 +1,187 @@
...
+            exceptions_dict={ResourceNotFoundError: []},
+            func=get_pods_by_isvc_label,
+            client=admin_client,
+            isvc=mlserver_inference_service,
+        ):
+            if sample:
+                predictor_pods = sample
+                break
+    except TimeoutExpiredError:
+        pass
```
dbasunag: Should it not fail if we get to timeoutExpired?

Snomaan6846: Agree, this was missed; it should fail if no pods came up within the specified timeout. Addressed.

kpunwatk (dismissed):

tests/model_serving/model_runtime/mlserver/gpu/conftest.py
Snomaan6846: Addressed.

tests/model_serving/model_runtime/mlserver/gpu/test_mlserver_gpu_fallback.py
Snomaan6846: Addressed.

tests/model_serving/model_runtime/mlserver/upgrade/conftest.py
Snomaan6846: Addressed, and explicit cleanup added.

tests/model_serving/model_runtime/mlserver/upgrade/conftest.py
Snomaan6846: Addressed.

dbasunag (approved):

PR #2128, opened 2026-07-31
fix: PVC storage tests — PSA compliance and write-access test rewrite

threcc:
## Problem 1: PodSecurity Admission rejection

KServe PVC storage tests (`test_kserve_pvc_rwx.py`, `test_kserve_pvc_write_access.py`) fail with `ForbiddenError: 403` on clusters with global PodSecurity Admission `enforce: restricted`.

The `download-model-data` pod created by `download_model_data()` in `utilities/general.py` lacks container-level `securityContext` fields required by the `restricted` PSA profile (`allowPrivilegeEscalation: false`, `capabilities.drop: ["ALL"]`, `runAsNonRoot: true`, `seccompProfile.type: RuntimeDefault`).

### Fix

1. **`utilities/general.py`**: When `restricted_scc_init=True`, apply `securityContext` with all required restricted PSA fields to both `init-container` and `model-downloader` containers. Also removed the `wait_for_status(RUNNING)` call before `wait_for_status(SUCCEEDED)` — the download pod can go `Pending → Succeeded` without passing through `Running`, which caused a hang.
2. **PVC test `conftest.py`**: Pass `restricted_scc_init=True` to `download_model_data()`.

## Problem 2: Write-access tests were testing the wrong thing

The original write-access tests used `touch /mnt/models/test` to verify the `storage.kserve.io/readonly` annotation. This conflated two independent mechanisms:

1. **Mount-level access** (`ro`/`rw`) — controlled by the annotation via KServe's admission webhook, which sets `volumeMount.readOnly` on the pod spec
2. **Filesystem-level permissions** (uid/gid/mode) — controlled by the storage backend, SCC, and `fsGroup`

The annotation only controls [1]. When `readonly=false`, the mount flag is `rw`, but `touch` can still fail due to filesystem ownership mismatch (the download pod writes as one UID, the serving pod runs as a different UID assigned by OpenShift's SCC). The tests were asserting on [2] while thinking they were testing [1].

### Fix

Rewrote the tests with two verification methods that test exactly what the annotation controls:

- **Pod spec check** (`get_volume_mount_readonly`): Reads `volumeMount.readOnly` from the pod spec — verifies the webhook set the correct value
- **`/proc/mounts` check** (`get_mount_mode`): Reads the kernel's mount options for `/mnt/models` — verifies the runtime honored the flag

### Test architecture

6 tests in `TestKservePVCWriteAccess`, organized as 4 default-state tests + 2 patch-and-toggle tests:

| Test | What it verifies |
|------|-----------------|
| `test_pod_containers_not_restarted` | No container restarts after deployment |
| `test_isvc_read_only_annotation_not_set_by_default` | Annotation absent by default |
| `test_isvc_read_only_pod_spec_default` | Webhook sets `readOnly=true` by default |
| `test_isvc_read_only_mount_default` | `/mnt/models` mounted `ro` by default |
| `test_isvc_read_only_false` | None→false (verify rw) → toggle to true (verify ro) |
| `test_isvc_read_only_true` | None→true (verify ro) → toggle to false (verify rw) |

Each patch/toggle test verifies all three layers: ISVC annotation value, pod spec `volumeMount.readOnly`, and `/proc/mounts` mount options.

### Key implementation details

- **`patched_read_only_isvc` fixture**: Uses `ResourceEditor` to patch the annotation, waits for deployment rollout to complete. On teardown, waits for the restore rollout too, so the next test sees a clean state.
- **`wait_for_rollout_complete`**: Polls the Deployment until `replicas == readyReplicas == updatedReplicas == expected`. Replaces the old `wait_deleted` approach that had a 4-minute timeout race.
- **`get_running_predictor_pod`**: Uses `Pod.get()` with `field_selector="status.phase=Running"` and `raw=True` to avoid the 404 race where old pods are fully deleted between list and `.instance` access. Picks the newest pod by `creationTimestamp`.
- **Inline toggle**: After the fixture sets the initial annotation value, the test body calls `isvc.update()` to toggle to the opposite value, waits for rollout, and verifies all three layers again.
- **Helpers in `utils.py`**: `wait_for_rollout_complete`, `get_running_predictor_pod`, `get_volume_mount_readonly`, `get_mount_mode` — avoids NIC001 conftest import violation.

## Test matrix (8 tests total)

**`TestKservePVCReadWriteManyAccess`** (2 tests):
- Verify both predictor pods in a 2-replica deployment can read from a shared RWX NFS PVC

**`TestKservePVCWriteAccess`** (6 tests):
- 4 default state tests + 2 patch-and-toggle tests covering None→false→true and None→true→false transitions
- All 8 tests pass in ~2 minutes

Screenshot From 2026-08-03 18-21-26

## Summary by CodeRabbit

* **Tests**
  * Expanded PVC read-only validation to cover default settings, configuration changes, mount modes, permissions, and actual write attempts.
  * Added rollout and predictor-service checks for more reliable model-serving updates.
  * Added diagnostics for mount permissions and model-directory access.

* **Bug Fixes**
  * Improved model download handling with restricted security settings and completion-based waiting.
  * Increased reliability when validating PVC access behavior during rollouts.

### DIFF
--- a/tests/model_serving/model_server/kserve/storage/pvc/conftest.py
+++ b/tests/model_serving/model_server/kserve/storage/pvc/conftest.py
@@ -14,7 +14,7 @@ from pytest import FixtureRequest
 from utilities.constants import KServeDeploymentType
 from utilities.general import download_model_data
 from utilities.inference_utils import create_isvc
-from utilities.infra import get_pods_by_isvc_label
+from utilities.infra import get_pods_by_isvc_label, verify_no_failed_pods, wait_for_inference_deployment_replicas
 
 
 @pytest.fixture(scope="class")
@@ -67,8 +67,20 @@ def predictor_pods_scope_class(
 
 @pytest.fixture()
 def patched_read_only_isvc(
-    request: FixtureRequest, pvc_inference_service: InferenceService, first_predictor_pod: Pod
+    request: FixtureRequest,
+    unprivileged_client: DynamicClient,
+    pvc_inference_service: InferenceService,
+    first_predictor_pod: Pod,
 ) -> Generator[InferenceService, Any, Any]:
+    """Patch the storage.kserve.io/readonly annotation on the ISVC.
+
+    Expects request.param with:
+        readonly (str): The annotation value to set ("true" or "false").
+        rollout (bool): Whether the patch changes the effective readOnly mount mode. When True,
+            waits for the old predictor pod to be deleted before proceeding (a deployment rollout
+            replaces it). When False, skips the wait — the pod spec is unchanged so no rollout
+            occurs (e.g. patching absent → "true", both resolve to readOnly=true).
+    """
     with ResourceEditor(
         patches={
             pvc_inference_service: {
@@ -78,7 +90,11 @@ def patched_read_only_isvc(
             }
         }
     ):
-        first_predictor_pod.wait_deleted()
+        if request.param.get("rollout", True):
+            first_predictor_pod.wait_deleted()
+
+        verify_no_failed_pods(client=unprivileged_client, isvc=pvc_inference_service)
+        wait_for_inference_deployment_replicas(client=unprivileged_client, isvc=pvc_inference_service)
         yield pvc_inference_service
 
 
--- a/tests/model_serving/model_server/kserve/storage/pvc/conftest.py
+++ b/tests/model_serving/model_server/kserve/storage/pvc/conftest.py
@@ -1,3 +1,4 @@
+import logging
 from collections.abc import Generator
 from typing import Any
 
@@ -11,10 +12,13 @@ from ocp_resources.resource import ResourceEditor
 from ocp_resources.serving_runtime import ServingRuntime
 from pytest import FixtureRequest
 
+from tests.model_serving.model_server.kserve.storage.pvc.utils import wait_for_rollout_complete
 from utilities.constants import KServeDeploymentType
 from utilities.general import download_model_data
 from utilities.inference_utils import create_isvc
-from utilities.infra import get_pods_by_isvc_label, verify_no_failed_pods, wait_for_inference_deployment_replicas
+from utilities.infra import get_pods_by_isvc_label
+
+LOGGER = logging.getLogger(__name__)
 
 
 @pytest.fixture(scope="class")
@@ -70,32 +74,31 @@ def patched_read_only_isvc(
     request: FixtureRequest,
     unprivileged_client: DynamicClient,
     pvc_inference_service: InferenceService,
-    first_predictor_pod: Pod,
 ) -> Generator[InferenceService, Any, Any]:
-    """Patch the storage.kserve.io/readonly annotation on the ISVC.
+    """Patch the storage.kserve.io/readonly annotation and wait for rollout.
+
+    On teardown, ResourceEditor restores the original annotation and the
+    fixture waits for that rollout too, so the next test sees a clean state.
 
     Expects request.param with:
         readonly (str): The annotation value to set ("true" or "false").
-        rollout (bool): Whether the patch changes the effective readOnly mount mode. When True,
-            waits for the old predictor pod to be deleted before proceeding (a deployment rollout
-            replaces it). When False, skips the wait — the pod spec is unchanged so no rollout
-            occurs (e.g. patching absent → "true", both resolve to readOnly=true).
     """
+    readonly_value = request.param["readonly"]
+    LOGGER.info(f"patched_read_only_isvc: patching readonly={readonly_value}")
     with ResourceEditor(
         patches={
             pvc_inference_service: {
                 "metadata": {
-                    "annotations": {"storage.kserve.io/readonly": request.param["readonly"]},
+                    "annotations": {"storage.kserve.io/readonly": readonly_value},
                 }
             }
         }
     ):
-        if request.param.get("rollout", True):
-            first_predictor_pod.wait_deleted()
-
-        verify_no_failed_pods(client=unprivileged_client, isvc=pvc_inference_service)
-        wait_for_inference_deployment_replicas(client=unprivileged_client, isvc=pvc_inference_service)
+        wait_for_rollout_complete(client=unprivileged_client, isvc=pvc_inference_service)
         yield pvc_inference_service
+    # ResourceEditor restores the annotation to None, triggering a rollout.
+    # Wait for it to settle so the next test sees a clean state.
+    wait_for_rollout_complete(client=unprivileged_client, isvc=pvc_inference_service)
 
 
 @pytest.fixture(scope="class")
--- a/tests/model_serving/model_server/kserve/storage/pvc/conftest.py
+++ b/tests/model_serving/model_server/kserve/storage/pvc/conftest.py
@@ -1,8 +1,8 @@
-import logging
 from collections.abc import Generator
 from typing import Any
 
 import pytest
+import structlog
 from kubernetes.dynamic import DynamicClient
 from ocp_resources.inference_service import InferenceService
 from ocp_resources.namespace import Namespace
@@ -18,7 +18,7 @@ from utilities.general import download_model_data
 from utilities.inference_utils import create_isvc
 from utilities.infra import get_pods_by_isvc_label
 
-LOGGER = logging.getLogger(__name__)
+LOGGER = structlog.get_logger(name=__name__)
 
 
 @pytest.fixture(scope="class")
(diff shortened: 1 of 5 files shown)


threcc:
**Update: additional test failure on cluster with `enforce: restricted`**

After the PSA fix, the `test_kserve_pvc_rwx` tests pass (2/2), and 5 of the 7 `test_kserve_pvc_write_access` tests pass. 2 write-access tests fail:

- `test_isvc_read_only_annotation_false`
- `test_isvc_readonly_annotation_toggle_lifecycle[test_readonly_annotation_false]`

**Root cause:** file ownership mismatch between the download pod and the serving pod.

The `download-model-data` pod runs in namespace `pvc-write-access` with `fsGroup: 1000420000` (from the namespace's `sa.scc.uid-range`). Files on the PVC are created with owner `1000:1000420000` and permissions `755` (owner-writable only).

The serving pod (`kserve-container`) runs as `uid 1001870000` with `fsGroup: 1001870000` — a different UID range. It cannot write to the PVC because it's neither the file owner nor in the owning group.

```
# PVC files ownership (from download pod)
drwxr-sr-x. 3 1000 1000420000  /mnt/models/

# Serving pod identity
uid=1001870000  gid=0(root)  groups=0(root),1001870000
```

Previously this was masked by the non-restricted code path which ran `chmod -R 777` on the PVC directory, making it world-writable. With `restricted_scc_init=True` that chmod is gone (it requires privilege escalation), exposing the underlying ownership mismatch.

**This is not a test-only issue** — on any cluster with `enforce: restricted`, PVC write access tests will fail because the download pod and serving pod run under different UID ranges. The `storage.kserve.io/readonly: "false"` annotation doesn't affect filesystem permissions.

**Error log:**
```
Execute ['touch', '/mnt/models/test'] on onnx-predictor-868cfd6f5-jsh4p (rhoai35-gcp-1-pool-hj-s8mp7-worker-a-s5gmb)

ocp_resources.exceptions.ExecOnPodError: Command execution failure: ['touch', '/mnt/models/test'],
  RC: -1, OUT: , ERR: {'metadata': {}, 'status': 'Failure',
  'message': 'command terminated with non-zero exit code: exit status 1',
  'reason': 'NonZeroExitCode', 'details': {'causes': [{'reason': 'ExitCode', 'message': '1'}]}}
```

@mwaykole  any idea on this?

mwaykole:
No.idea can you start thread in the  channel

threcc:
Updated PR description after fixing the tests

tests/model_serving/model_server/kserve/storage/pvc/test_kserve_pvc_write_access.py
```diff
@@ -76,66 +128,89 @@ def test_isvc_read_only_annotation_default_value(self, first_predictor_pod):
...
-        "patched_read_only_isvc, expected_raises",
+        "patched_read_only_isvc, expected_mode",
         [
-            pytest.param({"readonly": "false"}, False, id="test_readonly_annotation_false"),
-            pytest.param({"readonly": "true"}, True, id="test_readonly_annotation_true"),
+            pytest.param({"readonly": "false"}, "rw", id="test_readonly_annotation_false"),
+            pytest.param({"readonly": "true"}, "ro", id="test_readonly_annotation_true"),
         ],
         indirect=["patched_read_only_isvc"],
     )
```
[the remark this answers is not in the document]
threcc: Addressed by refactoring: the separate toggle tests and the `rollout` parameter are gone.

The `patched_read_only_isvc` fixture now always waits for rollout (no conditional logic). Toggle verification is merged inline into `test_isvc_read_only_false` (None→false→true) and `test_isvc_read_only_true` (None→true→false). Each test:
1. Fixture patches to the initial value, waits for rollout
2. Verifies annotation, pod spec, and `/proc/mounts`
3. Inline `update()` toggles to the opposite value, waits for rollout
4. Verifies all three again

Additionally, `get_active_predictor_pod` was replaced with `get_running_predictor_pod` which uses `Pod.get()` with `field_selector="status.phase=Running"` and `raw=True` to pick the newest Running pod by `creationTimestamp` — avoiding the 404 race where old pods were fully deleted between list and `.instance` access.

The fixture teardown also waits for rollout after `ResourceEditor` restores the annotation, preventing inter-test race conditions.

6 tests total, all passing.

tests/model_serving/model_server/kserve/storage/pvc/conftest.py
```diff
@@ -1,3 +1,4 @@
+import logging
```
dbasunag: Please use `import structlog`

tests/model_serving/model_server/kserve/storage/pvc/test_kserve_pvc_write_access.py
```diff
@@ -1,21 +1,22 @@
-import shlex
+import logging
```
dbasunag: Ditto. We switched to structlog a while back. Please make sure any recent pr did not start using logging.

dbasunag (dismissed):

tests/model_serving/model_server/kserve/storage/pvc/conftest.py
threcc: done

tests/model_serving/model_server/kserve/storage/pvc/test_kserve_pvc_write_access.py
threcc: done

utilities/general.py
```diff
@@ -156,35 +156,45 @@ def download_model_data(
...
+    restricted_security_context = {
+        "allowPrivilegeEscalation": False,
+        "capabilities": {"drop": ["ALL"]},
+        "runAsNonRoot": True,
+        "seccompProfile": {"type": "RuntimeDefault"},
+    }
+
+    init_container: dict[str, Any] = {
+        "name": "init-container",
+        "image": "quay.io/quay/busybox@sha256:92f3298bf80a1ba949140d77987f5de081f010337880cd771f7e7fc928f8c74d",
```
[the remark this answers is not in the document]
threcc: Already addressed: line 170 now uses `SharedImages.BUSYBOX` (added in `utilities/image_constants.py`). The raw reference was from an earlier revision.

utilities/general.py
```diff
@@ -156,35 +156,45 @@ def download_model_data(
...
+            {"name": "AWS_DEFAULT_REGION", "value": aws_default_region},
+            {"name": "S3_VERIFY_SSL", "value": "false"},
+            {"name": "awsAnonymousCredential", "value": "false"},
+        ],
+        "volumeMounts": [volume_mount],
+    }
+
+    if restricted_scc_init:
+        init_container["securityContext"] = restricted_security_context
+        downloader_container["securityContext"] = restricted_security_context
```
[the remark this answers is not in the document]
threcc: By design. The download pod creates 755 dirs owned by UID 1000 — the serving pod can't write to them regardless of mount mode. The test uses `chmod_model_directory()` right before asserting write access, simulating the customer workflow of preparing a PVC for read-write serving. This keeps the downloader generic and the chmod step explicit in the test.

threcc:
## Update: write assertions added + review fixes

### What changed since the last review

**Write assertions** — the two toggle tests now also verify actual write access via `try_write_file()`, not just mount mode and pod spec. Before asserting write on an rw mount, `chmod_model_directory()` sets the model subdirectory to 777 — simulating a customer preparing a PVC for read-write serving. This is needed because the download pod (busybox, UID 1000) creates 755 dirs, while the serving pod runs under the namespace SCC UID range.

The test now verifies 4 layers per state: annotation → pod spec `readOnly` → `/proc/mounts` → actual `touch` success/failure.

**Updated test matrix (7 tests):**

| Test | Assertion |
|------|-----------|
| `test_pod_containers_not_restarted` | No container restarts |
| `test_isvc_read_only_annotation_not_set_by_default` | Annotation absent |
| `test_isvc_read_only_pod_spec_default` | `readOnly=true` in pod spec |
| `test_isvc_read_only_mount_default` | `/mnt/models` mounted ro |
| `test_isvc_write_blocked_by_default` | `touch` fails on ro mount (755 dir) |
| `test_isvc_read_only_false` | None→false: chmod 777, write succeeds; toggle to true: write blocked by kernel ro mount even on 777 dir |
| `test_isvc_read_only_true` | None→true: write blocked; toggle to false: chmod 777, write succeeds |

### Other fixes
- Switched from `import logging` to `import structlog` (per reviewer feedback)
- Replaced blind `except Exception` with `except ExecOnPodError` in debug helpers
- `try_write_file` uses uuid-based filenames instead of a global counter (xdist-safe)
- Trimmed `log_write_debug_info` to essentials: `id`, `ls -la`, volumeMount spec
- `SharedImages.BUSYBOX` used everywhere (no raw image references)

All 7 tests pass in ~2m30s.

mwaykole (approved):
merging as we need  it

PR #2149, opened 2026-08-04
Enabled Power support

npanpaliya:
# Pull Request

## Related Issues

- Fixes: 
- JIRA: 

- [ ] Locally
- [ ] Jenkins

## Additional Requirements

## Summary by CodeRabbit

- **New Features**
  - Added support for building and running the application on x86_64 and PPC64LE platforms.
  - Added architecture-specific installation for required tools, libraries, and binary components.
  - Improved container setup for consistent dependency installation and command availability.

- **Bug Fixes**
  - Prevented unsupported evaluation tests from running on PPC64LE systems.
  - Improved platform-specific dependency configuration for more reliable application setup.


npanpaliya:
@coderabbitai resume

sheltoncyril:
Re `.github/workflows/build-push-container-on-merge.yml:45`:

An AI-assisted review flagged this and I wanted to check before treating it as actionable.

This looks like it changes the published image from `quay.io/opendatahub/opendatahub-tests` to `quay.io/opendatahub/odh-tests`, which would break existing consumers and diverges from the Makefile `IMAGE_NAME`. Also the numpy/scipy/grpcio caps appear dropped for x86 (gated to ppc64le only). Are the rename and the x86 dep-cap changes intentional? Thanks!

.github/workflows/build-push-container-on-merge.yml
```diff
@@ -60,27 +60,31 @@ jobs:
...
 
+      - name: Build and Push Multi-Arch Image
+        id: build-and-push
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          file: ./Dockerfile
+          platforms: linux/amd64,linux/ppc64le
+          push: true
+          tags: quay.io/opendatahub/odh-tests:${{ env.TAG }}
```
sheltoncyril: This line would break existing consumers, our image is published to opendatahub-tests on odh-tests.

sheltoncyril (changes requested):
need a further review from @dbasunag or @lugi0

.github/workflows/build-push-container-on-merge.yml
npanpaliya: Thanks. I corrected the image name.

pyproject.toml
```diff
@@ -149,11 +189,16 @@ dependencies = [
...
     "openai>=2.36",
-    "ragas>=0.4",
-    "numpy>=2.1.0,<2.5.0",
-    "scipy>=1.13.0,<1.18.0",
+    "tiktoken==0.12.0; platform_machine == 'ppc64le'",
+    "pyzmq==27.1.0",
+    "psutil==7.2.2",
+    "rfc3161-client==1.0.6; platform_machine == 'ppc64le'",
+    "ragas>=0.4; platform_machine != 'ppc64le'",
+    "numpy>=2.1.0,<2.5.0;platform_machine == 'ppc64le'",
```
npanpaliya, line 197: I've pinned numpy only for ppc because on non-ppc architecture, the latest version versions is fetched even without being specified here. So, on x86, the latest numpy would still be fetched as  a transitive dependency of some package. For ppc, I'd to put caps as the latest version doesn;t work.

npanpaliya: I checked, all the dependencies where I've put `platform == 'ppc64le' is just to restrict their version on ppc, and these are all transitive dependencies. On x86 they would just install whatever is mentioned in their parent  package.

npanpaliya:
@sheltoncyril - Could you please review it once?

npanpaliya:
> Re `.github/workflows/build-push-container-on-merge.yml:45`:
> 
> An AI-assisted review flagged this and I wanted to check before treating it as actionable.
> 
> This looks like it changes the published image from `quay.io/opendatahub/opendatahub-tests` to `quay.io/opendatahub/odh-tests`, which would break existing consumers and diverges from the Makefile `IMAGE_NAME`. Also the numpy/scipy/grpcio caps appear dropped for x86 (gated to ppc64le only). Are the rename and the x86 dep-cap changes intentional? Thanks!

@sheltoncyril  - This is addressed. I have corrected the image name to be published to quay.io. Apologies for that miss.
Regarding dependencies' pin, I've already replied.

npanpaliya:
@Raghul-M - Request you to review this PR.

dbasunag (changes requested):
Why do we need this? Jenkins uses amd containers and as far as I know it worked fine against power clusters. We never claim to support running tests from all sort of containers. We are testing the rhoai clusters not the containers that runs the tests.

PR #2154, opened 2026-08-04
fix(workbenches): add pod diagnostics to notebook readiness timeout errors

jstourac:
When wait_for_notebook_pod_ready times out, the error message now includes structured diagnostics: pod phase, scheduling gates, all conditions with reasons/messages, and not-ready container states.

This makes it possible to determine the root cause (e.g. Kueue scheduling gate, image pull failure, sidecar crash) directly from test logs without needing live cluster access.

We should backport to older branches all the way where it's easily manageable.

# Pull Request

## Related Issues

- Fixes: 
- JIRA: 

- [ ] Locally
- [ ] Jenkins

## Additional Requirements

## Summary by CodeRabbit

* **Bug Fixes**
  * Notebook readiness failures now include detailed pod diagnostics, such as scheduling status, conditions, and container or initialization errors.
  * Diagnostic collection errors are reported without obscuring the original readiness failure.

### DIFF
--- a/tests/workbenches/notebooks_server/controller/utils.py
+++ b/tests/workbenches/notebooks_server/controller/utils.py
@@ -246,11 +246,58 @@ def wait_for_notebook_pod_ready(notebook_pod: Pod, *, context: str, timeout: int
     except (TimeoutError, TimeoutExpiredError) as e:
         if notebook_pod.exists:
             collect_pod_information(notebook_pod)
+            diagnostics = _collect_pod_diagnostics(notebook_pod)
             raise AssertionError(
                 f"{context} pod '{notebook_pod.name}' failed to reach Ready state "
-                f"within {timeout} seconds.\nOriginal error: {e}"
+                f"within {timeout} seconds.\n{diagnostics}\nOriginal error: {e}"
             ) from e
 
         raise AssertionError(
             f"{context} pod '{notebook_pod.name}' was not created. Check notebook controller logs."
         ) from e
+
+
+def _collect_pod_diagnostics(pod: Pod) -> str:
+    """Collect pod state diagnostics for inclusion in assertion error messages."""
+    lines: list[str] = []
+    try:
+        instance = pod.instance
+        status = instance.status
+
+        lines.append(f"Phase: {status.phase}")
+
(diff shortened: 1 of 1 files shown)


tests/workbenches/notebooks_server/controller/utils.py
```diff
@@ -246,11 +246,58 @@ def wait_for_notebook_pod_ready(notebook_pod: Pod, *, context: str, timeout: int
...
+    lines: list[str] = []
+    try:
+        instance = pod.instance
+        status = instance.status
+
+        lines.append(f"Phase: {status.phase}")
+
+        scheduling_gates = instance.spec.get("schedulingGates") or []
+        if scheduling_gates:
+            gate_names = [g.get("name", "unknown") for g in scheduling_gates]
```
dbasunag: @jstourac can we refrain from using letters for variable names, please?

tests/workbenches/notebooks_server/controller/utils.py
```diff
@@ -246,11 +246,58 @@ def wait_for_notebook_pod_ready(notebook_pod: Pod, *, context: str, timeout: int
...
-                f"within {timeout} seconds.\nOriginal error: {e}"
+                f"within {timeout} seconds.\n{diagnostics}\nOriginal error: {e}"
             ) from e
 
         raise AssertionError(
             f"{context} pod '{notebook_pod.name}' was not created. Check notebook controller logs."
         ) from e
+
+
+def _collect_pod_diagnostics(pod: Pod) -> str:
```
dbasunag, line 260: Is collect_pod_information() not enough for triage?

jstourac: @dbasunag thank you for pointing this out. I originally thought it would be more convenient to have some basic important information directly in the log so one doesn't have to go and download extra testing files; though you are right, all the information is already part of that file collected by the `collect_pod_information`

jstourac:
Thanks for the review @dbasunag . I've pushed update to fix your concerns.

Regarding whether we truly need that - I replied directly, but I guess you are right. If I'm the only one missing this information directly in the log, let's not complicate this. I'll close the PR now.

PR #2165, opened 2026-08-06
fix(vllm fast): use pytest.skip instead of pytest.fail for absent fast templates

edwardquarm:
## Summary
- Fixes PR #1727 which intended to gracefully skip tests when fast templates are absent (RHOAIENG-68181), but the tests are **failing instead of skipping** because `pytest.fail()` was used instead of `pytest.skip()`
- Changes `pytest.fail()` to `pytest.skip()` so absent fast templates result in a skip, not a failure

## Test plan
- [x] Run vLLM fast tests on a cluster without fast templates — tests should show as SKIPPED
- [x] Run vLLM fast tests on a cluster with fast templates — tests should run normally

## Summary by CodeRabbit

* **Tests**
  * Tests now skip gracefully when the fast ServingRuntime template is unavailable, rather than failing.


brettmthompson (approved):
/lgtm

tests/model_serving/model_runtime/vllm/fast/conftest.py
```diff
@@ -17,15 +17,15 @@ def serving_runtime(  # noqa: UFN001
...
+    """ServingRuntime from a fast template; skips if the template is absent."""
     template_name: str = request.param["template_name"]
     template = Template(
         client=admin_client,
         name=template_name,
         namespace=py_config["applications_namespace"],
     )
     if not template.exists:
-        pytest.fail(
+        pytest.skip(
```
sheltoncyril, line 28: Thanks for the PR, I think from DevTestOps POV, skipped tests have the same value as failed tests, so this test might lead to your team getting blocker tickets created in case this gets skipped. I would advise that you make it pass instead if it is non-problematic instead of skipping it. @dbasunag am I right in my theory here?

sheltoncyril (changes requested):

tests/model_serving/model_runtime/vllm/fast/conftest.py
dbasunag: I don't think this PR is useful. If you expect a template to exist in a rhoai deployment, then it being missing is a problem, that should not be hidden behind a skip. Whether you fail or skip, it would get highlighted going forward. Passing is not an option either - in that case there is no point even testing. This PR should be closed instead.

dbasunag (changes requested):

tests/model_serving/model_runtime/vllm/fast/conftest.py
edwardquarm: the fast templates won't exists in some RHOAI versions and that's expected. @dbasunag these changes are needed as vLLM fast images is a feature coming up in RHAIIS 3.6-fast-1 (3.6 EA1). I don't know why this got added to the main branch but its currently blocking us from running validations  on vLLM CUDA images for 3.5.

pierDipi: this is one case where skip = fail is going to create noise for teams because we have tests that are only applicable to certain versions/conditions (in this case patch releases) like this, we'd welcome real alternatives that don't block progress.

edwardquarm:
Closing — the fail behavior is correct for main (3.6ea1). The fix belongs on the 3.5 branch where fast templates don't exist yet. Opening a new PR to remove these tests from 3.5.

pierDipi:
@edwardquarm we're going to have the same problem testing 3.6ea1 as well, ea releases don't have fast variants, let's keep it here too

PR #2168, opened 2026-08-06
fix(vllm): exclude fast template tests from collection on 3.5

edwardquarm:
## Summary
- Fixes failing vLLM fast template tests on the 3.5 branch
- Fast runtime templates (`vllm-fast-1-cuda`, `vllm-fast-2-cuda`) are targeted for 3.5.z patches (via RHAIIS 3.6ea1) and do not exist on 3.5 GA clusters
- The tests were failing on setup with `pytest.fail` because the templates are absent, causing false failures in CI
- Uses `pytest_collection_modifyitems` to deselect `fast/` tests from collection instead of removing them, so they can be re-enabled when fast images land on 3.5.z


dbasunag (approved):

brettmthompson (approved):
/lgtm

PR #2170, opened 2026-08-07
fix(vllm): exclude fast template tests from collection

edwardquarm:
## Summary
- Same fix as #2168 (3.5 branch) but for main
- Fast runtime templates are not yet available on all test clusters, causing false failures in CI when tests run on clusters without the templates
- Uses `pytest_collection_modifyitems` to deselect `fast/` tests from collection instead of removing them, so they can be re-enabled when fast images are available
- See discussion: https://github.com/opendatahub-io/opendatahub-tests/pull/2165#issuecomment-5208450955

## Test plan
- [ ] Verify vLLM CUDA test suite no longer reports false failures from absent fast templates
- [ ] Verify fast tests are deselected (not collected) during test runs

## Summary by CodeRabbit

* **Tests**
  * Updated test selection to exclude tests in the fast-test path during collection.
  * Retained all other tests and clearly reported excluded tests as deselected.

### DIFF
--- a/tests/model_serving/model_runtime/vllm/conftest.py
+++ b/tests/model_serving/model_runtime/vllm/conftest.py
@@ -38,6 +38,21 @@ LOGGER = structlog.get_logger(name=__name__)
 SUPPORTED_CPU_X86_ACCELERATORS: set[str] = {AcceleratorType.CPU_x86}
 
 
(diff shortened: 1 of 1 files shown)


dbasunag (approved):

brettmthompson (approved):
/lgtm

PR #1861, opened 2026-06-25
refactor: session-scoped LLM emulator fixtures to reduce test setup time

sheltoncyril:
## Summary

- Moves stateless LLM emulators (LLM-d sim, vLLM emulator) from class-scoped to session-scoped fixtures
- Deploys emulators once per test session to a shared `ai-safety-emulators` namespace
- Tests reference emulators via FQDN for cross-namespace access
- Saves ~76 min per full ai_safety run by eliminating redundant deploy/teardown cycles

## Components refactored

- **guardrails:** 4 → 1 LLM-d sim cycle (tier1), + HF detectors
- **nemo_guardrails:** 5 → 1 LLM-d sim cycle (tier1+tier2)
- **evalhub:** 9 → 1 vLLM emulator cycle (multitenancy+mcp+garak)
- **lm_eval:** function-scoped → session-scoped vLLM emulator

## What's NOT changed

- GPU tests (`test_guardrails_gpu.py`, `test_lmeval_gpu`) — keep class-scoped real GPU models
- Upgrade tests — keep class-scoped for pre/post-upgrade lifecycle
- TrustyAI tests — use real ML models, not emulators

## Key design decisions

- Session-scoped `emulator_namespace` fixture creates `ai-safety-emulators` ns once
- `EMULATOR_NAMESPACE` constant + `get_llm_d_chat_generation_config()` function for FQDN hostnames
- vLLM emulator fixtures consolidated in `tests/ai_safety/conftest.py` (shared by evalhub + lm_eval)
- LLM-d sim session fixtures in `tests/fixtures/inference.py`

## Jira

Fixes: RHOAIENG-71674, RHOAIENG-71675, RHOAIENG-71676, RHOAIENG-71677, RHOAIENG-71678
Parent: RHOAIENG-66046

## Test plan

- [ ] Run guardrails tier1 tests — verify session ISVC shared across classes
- [ ] Run nemo_guardrails tier1 tests — verify ConfigMap URLs use FQDN
- [ ] Run evalhub multitenancy tests — verify session vLLM emulator shared
- [ ] Run lm_eval vllm_emulator test — verify FQDN in LMEvalJob base_url
- [ ] Run guardrails GPU tests — verify unchanged, still working
- [ ] Run smoke tests — verify no LLM deployed (CRD checks only)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Tests**
[...]

PR #1962, opened 2026-07-08
test: replace Timeout constants with explicit integers

tareqalayan:
Replace all Timeout.TIMEOUT_* constants with explicit integer values (in seconds) across the entire tests/ directory for improved clarity and centralized timeout management.

Replacements:
- Timeout.TIMEOUT_1MIN (60s) → 60
- Timeout.TIMEOUT_2MIN (120s) → 120
- Timeout.TIMEOUT_4MIN (240s) → 240
- Timeout.TIMEOUT_5MIN (300s) → 300
- Timeout.TIMEOUT_10MIN (600s) → 600
- Timeout.TIMEOUT_15MIN (900s) → 900
- Timeout.TIMEOUT_20MIN (1200s) → 1200
- Timeout.TIMEOUT_30MIN (1800s) → 1800

Changes:
- 98 timeout constant replacements across 44 files
- Removed unused Timeout imports from 35 files
- Preserved tts() function calls (unchanged)
- Preserved OpenshiftRouteTimeout constants (different class)

Impact: Explicit timeout values are now self-documenting and easier to adjust for different test environments.

Fixes: RHOAIENG-68266

# Pull Request

## Related Issues

- Fixes: 
- JIRA: 

## Please review and indicate how it has been tested

- [ ] Locally
- [ ] Jenkins

## Additional Requirements

- [ ] If this PR introduces a new test image, did you create a PR to mirror it in disconnected environment?
- [ ] If this PR introduces new marker(s)/adds a new component, was relevant ticket created to update relevant Jenkins job?

## Summary by CodeRabbit

* **Tests**
  * Standardized many test and fixture timeouts to explicit seconds instead of shared constants.
  * Updated readiness, deployment, and polling waits across AI safety, model serving, notebooks, pipelines, and workbench test coverage.
  * Preserved existing behavior while making timeout values easier to read and maintain.

* **Chores**
  * Removed unused timeout imports where they were no longer needed.

PR #2201, opened 2026-08-12
refactor(ai_safety): extract shared ISVC deployment helper for trustyai/guardrails

sheltoncyril:
## Summary

Relates to RHOAIENG-66047 ("Consolidate model deployment fixtures across AI Safety sub-components").

- Adds `get_or_create_isvc()` to `tests/fixtures/inference.py` — a shared helper that consolidates the "create a new InferenceService, or reference the one from pre-upgrade and clean it up" pattern that was duplicated across ai_safety sub-component `conftest.py` files.
- Refactors **trustyai_service**'s `gaussian_credit_model` fixture and **guardrails**' `prompt_injection_detector_isvc` / `hap_detector_isvc` fixtures to use the new helper instead of duplicating the create/wait/teardown logic inline.
- Behavior is preserved exactly, including each component's quirks (see "Design notes" below).

`tests/fixtures/inference.py` is already the established location for cross-component inference fixtures/helpers (it's registered as a `pytest_plugins` entry in `tests/conftest.py` and already hosts `qwen_isvc`, `qwen_gpu_isvc`, etc. used across components), so this follows existing convention rather than introducing a new pattern. Per repo convention, importing it from `tests/ai_safety/*/conftest.py` requires `# noqa: NIT001` (same pattern already used by `tests/pipelines_components/autorag/conftest.py` importing `tests.fixtures.vector_io`).

### Design notes / preserved per-component behavior

The two consumers weren't quite identical, so the helper takes a few flags rather than forcing false uniformity:
- `wait_for_ready_post_upgrade` / `ready_timeout`: Guardrails' detector fixtures wait for the existing ISVC to report `Ready` (timeout 600s) in post-upgrade mode; TrustyAI's `gaussian_credit_model` historically did not. Preserved via a flag (default `False`, matching TrustyAI).
- `gate_post_upgrade_cleanup_by_teardown`: Guardrails only calls `isvc.clean_up()` post-upgrade `if teardown_resources`; TrustyAI cleans up unconditionally. Preserved via a flag (default `False` = unconditional, matching TrustyAI; Guardrails passes `True`).
- `post_create_hook`: TrustyAI's `gaussian_credit_model` needs `wait_for_isvc_deployment_registered_by_trustyai_service()` called right after creation, before yielding. Exposed as an optional callback so only TrustyAI needs to pass it.

## Acceptance criteria status (RHOAIENG-66047)

- [x] **Common ISVC deployment helper extracted and reused by at least 2 sub-components** — `get_or_create_isvc()` is used by trustyai_service and guardrails.
- [ ] **At least one model fixture promoted from class-scope to session-scope** — investigated, **not done**, see below.
- [ ] **Test execution time reduced for multi-class test suites** — not achieved by this PR (no fixture scope changed); the helper itself reduces duplicated code/maintenance surface but not runtime.
- [x] **No regression in test isolation** — no fixture scope was changed; behavior is byte-for-byte equivalent to before (verified via `pytest --collect-only` across the full `tests/ai_safety/` suite — 461/501 tests collect identically — and a line-by-line diff review of both refactored fixtures against their originals).

### Why the session-scope promotion (2nd criterion) was not done

The ticket suggested `gaussian-credit-model-modelcar` (trustyai_service) and the Guardrails HuggingFace detector models (`prompt-injection-detector`, `hap-detector`) as candidates, with the check being "does any test mutate the ISVC." Investigation turned up a different, more fundamental blocker specific to these two components: **the model must be colocated in the same namespace as the operator watching it**, so it can't simply move to a shared/global namespace the way the (separate, in-review) stateless-emulator work in PR #1861 did for `llm_d_inference_sim_isvc`:

- **TrustyAIService**: `wait_for_isvc_deployment_registered_by_trustyai_service()` in `tests/ai_safety/trustyai_service/trustyai_service_utils.py` derives the expected KServe logger sink as `f"{trustyai_service.name}.{isvc.namespace}.svc.cluster.local"` — i.e. TrustyAIService only ever watches deployments in its own namespace. Since each test class already gets its own `model_namespace` (and a `trustyai_service` instance configured for that class's specific storage variant — PVC vs DB), moving `gaussian_credit_model` to a separate session-shared namespace would decouple it from the TrustyAIService instance meant to monitor it.
- **GuardrailsOrchestrator autoConfig**: `TestGuardrailsOrchestratorAutoConfig`'s own docstring in `tests/ai_safety/guardrails/test_guardrails.py` already states: *"AutoConfig requires the ISVC in the same namespace as the orchestrator, so these tests use class-scoped llm_d_inference_sim_isvc (not session-scoped)."* The same constraint applies to `hap_detector_isvc` / `prompt_injection_detector_isvc`, which both `TestGuardrailsOrchestratorAutoConfig` and `TestGuardrailsOrchestratorAutoConfigWithGateway` consume via `detectorServiceLabelToMatch` label-based, namespace-scoped discovery.

The one Guardrails class that *doesn't* use autoConfig (`TestGuardrailsOrchestratorWithHuggingFaceDetectors`) references detectors via an explicit hostname built in `tests/fixtures/guardrails.py`'s `orchestrator_config_gpu`-style fixtures (`f"{PROMPT_INJECTION_DETECTOR}-predictor.{model_namespace.name}.svc.cluster.local"`), which is in principle cross-namespace-capable — but it's the only consumer of these two fixtures via that path, so there's no actual multi-class redundancy to eliminate, and forcing it through a new shared namespace would require also updating that shared `orchestrator_config` fixture's hostname construction (used by several other parametrizations) — a bigger, harder-to-validate change I didn't want to make without live-cluster testing.

I'd rather report this honestly than force a promotion that either breaks correctness or can't be verified. Recommended follow-up once PR #1861's `shared_models_namespace`/`AI_SAFETY_SHARED_MODELS_NAMESPACE` infra lands: extend it to also host a session-scoped `hap_detector_isvc`/`prompt_injection_detector_isvc` pair specifically for `TestGuardrailsOrchestratorWithHuggingFaceDetectors`, since that's the only verified-safe candidate found.

## Testing

- `python -m py_compile` on all changed files: passed.
- `pre-commit run --files <changed files>` (ruff, ruff format, flake8 incl. the repo's custom NIT001 import-isolation rule, detect-secrets, etc.): passed, run twice.
- `pytest --collect-only -q tests/ai_safety/`: 461/501 tests collected, identical to pre-refactor collection (no import or fixture-resolution errors).
- **No live-cluster/integration test run was performed** — this repo's tests require a live OpenShift cluster with AI Safety components installed, which wasn't available in this environment. The refactor was validated via static analysis (compile, lint, type-check config) and manual line-by-line review confirming the extracted helper reproduces each fixture's exact prior behavior (including the per-component differences noted above).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when setting up inference services during AI safety and TrustyAI test scenarios.
  * Preserved readiness checks, cleanup behavior, authentication, resource configuration, and registration waits during upgrades.

* **Refactor**
  * Standardized inference service creation and reuse through shared test infrastructure.
  * Reduced duplicated setup, upgrade handling, and cleanup logic across test workflows.

### DIFF
--- a/tests/ai_safety/guardrails/conftest.py
+++ b/tests/ai_safety/guardrails/conftest.py
@@ -32,11 +32,12 @@ from tests.ai_safety.guardrails.constants import (
     TEST_TLS_PRIVATE_KEY,
 )
 from tests.ai_safety.image_constants import AiSafetyImages
+from tests.fixtures.inference import get_or_create_isvc  # noqa: NIT001
 from utilities.constants import (
     KServeDeploymentType,
     RuntimeTemplates,
 )
-from utilities.inference_utils import LOGGER, create_isvc
+from utilities.inference_utils import LOGGER
 from utilities.operator_utils import get_cluster_service_version
 from utilities.serving_runtime import ServingRuntimeFromTemplate
 
@@ -69,46 +70,33 @@ def prompt_injection_detector_isvc(
     pytestconfig: pytest.Config,
     teardown_resources: bool,
 ) -> Generator[InferenceService, Any, Any]:
-    if pytestconfig.option.post_upgrade:
-        isvc = InferenceService(
-            client=admin_client,
-            name="prompt-injection-detector",
-            namespace=model_namespace.name,
-        )
-        isvc.wait_for_condition(
-            condition=isvc.Condition.READY,
-            status=isvc.Condition.Status.TRUE,
-            timeout=600,
-        )
-        yield isvc
-        if teardown_resources:
-            isvc.clean_up()
-    else:
-        # During pre-upgrade or normal tests, create new InferenceService
-        with create_isvc(
-            client=admin_client,
-            name="prompt-injection-detector",
-            namespace=model_namespace.name,
-            deployment_mode=KServeDeploymentType.RAW_DEPLOYMENT,
-            model_format="guardrails-detector-huggingface",
-            runtime=huggingface_sr.name,
-            storage_key=minio_data_connection.name,
-            storage_path="deberta-v3-base-prompt-injection-v2",
-            wait_for_predictor_pods=False,
-            enable_auth=False,
-            resources={
-                "requests": {"cpu": "1", "memory": "2Gi", "nvidia.com/gpu": "0"},
-                "limits": {"cpu": "1", "memory": "2Gi", "nvidia.com/gpu": "0"},
-            },
-            max_replicas=1,
-            min_replicas=1,
-            labels={
-                "opendatahub.io/dashboard": "true",
-                AUTOCONFIG_DETECTOR_LABEL: "true",
-            },
-            teardown=teardown_resources,
-        ) as isvc:
-            yield isvc
+    yield from get_or_create_isvc(
+        admin_client=admin_client,
+        name="prompt-injection-detector",
+        namespace=model_namespace.name,
+        pytestconfig=pytestconfig,
+        teardown=teardown_resources,
+        wait_for_ready_post_upgrade=True,
+        ready_timeout=600,
+        gate_post_upgrade_cleanup_by_teardown=True,
+        deployment_mode=KServeDeploymentType.RAW_DEPLOYMENT,
+        model_format="guardrails-detector-huggingface",
+        runtime=huggingface_sr.name,
+        storage_key=minio_data_connection.name,
+        storage_path="deberta-v3-base-prompt-injection-v2",
+        wait_for_predictor_pods=False,
+        enable_auth=False,
+        resources={
+            "requests": {"cpu": "1", "memory": "2Gi", "nvidia.com/gpu": "0"},
+            "limits": {"cpu": "1", "memory": "2Gi", "nvidia.com/gpu": "0"},
+        },
+        max_replicas=1,
+        min_replicas=1,
+        labels={
+            "opendatahub.io/dashboard": "true",
+            AUTOCONFIG_DETECTOR_LABEL: "true",
+        },
+    )
 
 
 @pytest.fixture(scope="class")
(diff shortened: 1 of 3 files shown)


tests/ai_safety/guardrails/conftest.py
```diff
@@ -69,46 +70,33 @@ def prompt_injection_detector_isvc(
...
-            max_replicas=1,
-            min_replicas=1,
-            labels={
-                "opendatahub.io/dashboard": "true",
-                AUTOCONFIG_DETECTOR_LABEL: "true",
-            },
-            teardown=teardown_resources,
-        ) as isvc:
-            yield isvc
+    yield from get_or_create_isvc(
```
kpunwatk, line 73: @sheltoncyril I don't see an if statement why 'or' ?

tests/ai_safety/guardrails/conftest.py
```diff
@@ -69,46 +70,33 @@ def prompt_injection_detector_isvc(
...
+        gate_post_upgrade_cleanup_by_teardown=True,
+        deployment_mode=KServeDeploymentType.RAW_DEPLOYMENT,
+        model_format="guardrails-detector-huggingface",
+        runtime=huggingface_sr.name,
+        storage_key=minio_data_connection.name,
+        storage_path="deberta-v3-base-prompt-injection-v2",
+        wait_for_predictor_pods=False,
+        enable_auth=False,
+        resources={
+            "requests": {"cpu": "1", "memory": "2Gi", "nvidia.com/gpu": "0"},
```
kpunwatk, line 90: if the gpu requirement is null do we still need to add this "nvidia.com/gpu": "0" as a parameter?

tests/fixtures/inference.py
```diff
@@ -32,6 +33,77 @@
...
 
 
+def get_or_create_isvc(
+    admin_client: DynamicClient,
+    name: str,
+    namespace: str,
+    pytestconfig: pytest.Config,
+    teardown: bool,
+    wait_for_ready_post_upgrade: bool = False,
+    ready_timeout: int = Timeout.TIMEOUT_10MIN,
```
kpunwatk: I believe we are getting away from these constants and replace with explicit integer values [PR ](https://github.com/opendatahub-io/opendatahub-tests/pull/1962)

tests/fixtures/inference.py
```diff
@@ -32,6 +33,77 @@
...
+        post_create_hook: Optional callback invoked with the newly created InferenceService right
+            after creation (only outside post_upgrade mode), before it's yielded. Used for
+            component-specific post-creation waits, e.g. TrustyAI's
+            `wait_for_isvc_deployment_registered_by_trustyai_service`.
+        **create_isvc_kwargs: Forwarded to `utilities.inference_utils.create_isvc` when not in
+            post_upgrade mode (e.g. model_format, runtime, storage_uri/storage_key/storage_path,
+            resources, labels, wait_for_predictor_pods, enable_auth, min_replicas, max_replicas).
+
+    Yields:
+        InferenceService
```
kpunwatk: We don't need long description within the fixture with the args and it's functionality as they are already been defined.

kpunwatk (changes requested):

tests/ai_safety/guardrails/conftest.py
sheltoncyril: The if/else lives inside the shared `get_or_create_isvc` helper now (tests/fixtures/inference.py) — that's the point of extracting it, this call site no longer needs its own if/else. "get_or_create" is just naming it after that pattern (create new, or get the pre-upgrade one).

tests/ai_safety/guardrails/conftest.py
sheltoncyril: Pre-existing behavior, unchanged by this refactor — verified against origin/main before this PR, the `resources` dict (including the explicit `nvidia.com/gpu: "0"`) is passed through as-is via `**create_isvc_kwargs`. Out of scope for this ISVC-helper-consolidation PR; happy to file a follow-up if you'd like it cleaned up.

tests/fixtures/inference.py
sheltoncyril: Fixed — replaced with explicit `600` in 4e9477d, matching the #1962 convention.

tests/fixtures/inference.py
sheltoncyril: Fixed — trimmed to a one-line summary in 4e9477d.

tests/ai_safety/guardrails/conftest.py
```diff
@@ -32,11 +32,12 @@
     TEST_TLS_PRIVATE_KEY,
 )
 from tests.ai_safety.image_constants import AiSafetyImages
+from tests.fixtures.inference import get_or_create_isvc  # noqa: NIT001
```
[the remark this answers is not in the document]
sheltoncyril: Kept as-is: NIT001 is a real custom flake8 rule in this repo ("Import from tests is not allowed"), not registered in ruff's lint.external (which only tracks ruff-legacy-alias codes). Removing the noqa breaks the actual flake8 pre-commit hook — confirmed locally. Same suppression pattern is already used for tests.fixtures imports elsewhere (tests/pipelines_components/autorag/conftest.py).

sheltoncyril:
Closing this PR.

The consolidation in RHOAIENG-66047 depended on the shared `get_or_create_isvc()` helper having at least two real consumers — `trustyai_service` and `guardrails`. With the FMS Guardrails ecosystem being removed from the operator (trustyai-explainability/trustyai-service-operator#902) and the corresponding test suite deleted (#2328), the guardrails consumer no longer exists.

That leaves only `trustyai_service` using the helper, which reduces the change to indirection rather than genuine cross-component deduplication — the ticket's core goal ("helper reused by at least 2 sub-components") can no longer be met without forcing an artificial second consumer.

Rather than force a false fit, closing this as no-longer-applicable. If a second suite with the same create-or-reference-on-upgrade pattern emerges later, the helper can be revisited then.

PR #2224, opened 2026-08-19
test(notebooks extensions): do not require pipeline server for elyra test right now

hmtosi:
The current Elyra test verifies that the extension and runtime config are preserved across the migration. The test does not depend on the pipeline server right now, so we will allow Elyra to start without the pipelines platform. [Future work](https://redhat.atlassian.net/browse/RHOAIENG-85728) will update the test to utilize the platform, and at that time we will remove this error from the allowed list.


sheltoncyril (approved):
/lgtm

PR #2238, opened 2026-08-20
test: add observability smoke tests

SB159:
# Pull Request

## Summary

add observability smoke tests

## Related Issues

- Fixes: 
- JIRA: 

- [x] Locally
- [ ] Jenkins

## Additional Requirements

### DIFF
--- a/tests/ai_gateway/models_as_a_service/observability/conftest.py
+++ b/tests/ai_gateway/models_as_a_service/observability/conftest.py
@@ -29,9 +29,7 @@ def maas_monitoring_namespace(
     """Return the monitoring namespace configured for maas-controller observability."""
     monitoring_namespace = resolve_maas_monitoring_namespace(admin_client=admin_client)
     if not monitoring_namespace:
-        pytest.fail(
-            "maas-controller MONITORING_NAMESPACE is not configured; observability reconcile cannot run"
-        )
+        pytest.fail("maas-controller MONITORING_NAMESPACE is not configured; observability reconcile cannot run")
     if not monitoring_namespace_exists(admin_client=admin_client, namespace_name=monitoring_namespace):
         pytest.fail(
             f"Monitoring namespace '{monitoring_namespace}' does not exist; "
@@ -49,8 +47,7 @@ def servicemonitor_crd_available(admin_client: DynamicClient) -> None:
     )
     if not service_monitor_crd.exists:
         pytest.fail(
-            f"ServiceMonitor CRD '{SERVICE_MONITOR_CRD_NAME}' not installed; "
-            "Limitador metrics scrape requires it"
+            f"ServiceMonitor CRD '{SERVICE_MONITOR_CRD_NAME}' not installed; Limitador metrics scrape requires it"
         )
 
 
--- a/tests/ai_gateway/models_as_a_service/observability/conftest.py
+++ b/tests/ai_gateway/models_as_a_service/observability/conftest.py
@@ -6,6 +6,7 @@ from ocp_resources.service_monitor import ServiceMonitor
 from ocp_utilities.monitoring import Prometheus
 
 from tests.ai_gateway.models_as_a_service.observability.constants import (
+    LIMITADOR_SERVICE_MONITOR_NAME,
     RHOAI_THANOS_QUERIER_ROUTE_NAME,
     SERVICE_MONITOR_CRD_NAME,
 )
@@ -67,10 +68,16 @@ def limitador_service_monitor(
     servicemonitor_crd_available: None,
 ) -> ServiceMonitor:
     """Wait for maas-controller to create the Limitador ServiceMonitor."""
-    return wait_for_limitador_service_monitor(
+    limitador_service_monitor = wait_for_limitador_service_monitor(
         admin_client=admin_client,
         monitoring_namespace=maas_monitoring_namespace,
     )
+    if limitador_service_monitor is None:
+        pytest.fail(
+            f"ServiceMonitor '{LIMITADOR_SERVICE_MONITOR_NAME}' not found in "
+            f"'{maas_monitoring_namespace}' after maas-controller reconcile"
+        )
+    return limitador_service_monitor
 
 
 @pytest.fixture(scope="session")
--- a/tests/ai_gateway/models_as_a_service/observability/conftest.py
+++ b/tests/ai_gateway/models_as_a_service/observability/conftest.py
@@ -29,23 +29,29 @@ def maas_monitoring_namespace(
     """Return the monitoring namespace configured for maas-controller observability."""
     monitoring_namespace = resolve_maas_monitoring_namespace(admin_client=admin_client)
     if not monitoring_namespace:
-        pytest.skip("maas-controller MONITORING_NAMESPACE is not configured; observability reconcile is skipped")
+        pytest.fail(
+            "maas-controller MONITORING_NAMESPACE is not configured; observability reconcile cannot run"
+        )
     if not monitoring_namespace_exists(admin_client=admin_client, namespace_name=monitoring_namespace):
-        pytest.skip(
-            f"Monitoring namespace '{monitoring_namespace}' does not exist; maas-controller skips observability setup"
+        pytest.fail(
+            f"Monitoring namespace '{monitoring_namespace}' does not exist; "
+            "maas-controller observability setup requires it"
         )
     return monitoring_namespace
 
 
 @pytest.fixture(scope="session")
 def servicemonitor_crd_available(admin_client: DynamicClient) -> None:
-    """Skip when the Prometheus Operator ServiceMonitor CRD is not installed."""
+    """Fail when the Prometheus Operator ServiceMonitor CRD is not installed."""
     service_monitor_crd = CustomResourceDefinition(
         client=admin_client,
         name=SERVICE_MONITOR_CRD_NAME,
     )
     if not service_monitor_crd.exists:
-        pytest.skip("ServiceMonitor CRD not installed; Limitador metrics scrape is optional on this cluster")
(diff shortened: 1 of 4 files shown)


tests/ai_gateway/models_as_a_service/observability/utils.py
```diff
@@ -0,0 +1,260 @@
...
+    endpoint = spec.endpoints[0]
+    assert endpoint.path == LIMITADOR_METRICS_PATH
+    assert endpoint.port == LIMITADOR_METRICS_PORT
+    assert str(endpoint.interval) == expected_limitador_scrape_interval(maas_config=maas_config)
+
+    selector_labels = dict(spec.selector.matchLabels or {})
+    assert selector_labels.get("app") == LIMITADOR_APP_LABEL
+
+    namespace_selector = spec.namespaceSelector
+    assert namespace_selector.any is True, "Expected namespaceSelector.any=true for cross-namespace Limitador scrape"
```
ahadas: one of the issues in **my** current implementation is that the service monitor looks for a pod with the app=limitador label in all namespaces, so theoretically someone can start such a pod in their own namespace and we'll get their metrics...

I'm saying this because if/when I fix that, the namespace selector would change from "any" to the namespace we find the Kuadrant CR on - so how about preparing this test for that in advance and check that the namespace in which Kuadrant CR exists is matched by the namespace selector?

ahadas (commented):
looks good overall, one comment inside

jland-redhat:
So looks ok overall but I do think one big thing that is missing is that we want to validate that we are at least seeing a basic population of `authorized_hits` and that those include like model/user/group information (if that is turned on in the telemetry policy)

ahadas:
> So looks ok overall but I do think one big thing that is missing is that we want to validate that we are at least seeing a basic population of `authorized_hits` and that those include like model/user/group information (if that is turned on in the telemetry policy)

+1
IIUC, the plan is to add this at a different layer, not in smoke tests - @SB159 did I get that right?

tests/ai_gateway/models_as_a_service/observability/utils.py
SB159: Good point - I’ll update the ServiceMonitor assertion to resolve the Kuadrant CR namespace and verify the namespaceSelector targets that namespace, instead of asserting any: true. That should stay valid when you tighten the selector.

SB159:
> > So looks ok overall but I do think one big thing that is missing is that we want to validate that we are at least seeing a basic population of `authorized_hits` and that those include like model/user/group information (if that is turned on in the telemetry policy)
> 
> +1 IIUC, the plan is to add this ay a different layer, not in smoke tests - @SB159 did I get that right?

@jland-redhat @ahadas Yes, that’s the plan. This smoke PR covers maas-controller reconcile (ServiceMonitor, limitador_up, usageLogging off). authorized_hits after an inference call with model/user/group labels is planned as a follow-up test once smoke lands — not in this PR.

tests/ai_gateway/models_as_a_service/observability/utils.py
SB159: Addressed in the latest push — ServiceMonitor check now verifies namespaceSelector against the Kuadrant CR namespace instead of asserting any: true.

ahadas (approved):

sheltoncyril (dismissed):

tests/ai_gateway/models_as_a_service/observability/conftest.py
```diff
@@ -0,0 +1,107 @@
...
+    maas_subscription_controller_enabled_latest: None,
+) -> str:
+    """Return the monitoring namespace configured for maas-controller observability."""
+    monitoring_namespace = resolve_maas_monitoring_namespace(admin_client=admin_client)
+    if not monitoring_namespace:
+        pytest.skip("maas-controller MONITORING_NAMESPACE is not configured; observability reconcile is skipped")
+    if not monitoring_namespace_exists(admin_client=admin_client, namespace_name=monitoring_namespace):
+        pytest.skip(
+            f"Monitoring namespace '{monitoring_namespace}' does not exist; maas-controller skips observability setup"
+        )
```
dbasunag, line 38: Please no skips.

tests/ai_gateway/models_as_a_service/observability/conftest.py
```diff
@@ -0,0 +1,107 @@
...
+
+@pytest.fixture(scope="session")
+def servicemonitor_crd_available(admin_client: DynamicClient) -> None:
+    """Skip when the Prometheus Operator ServiceMonitor CRD is not installed."""
+    service_monitor_crd = CustomResourceDefinition(
+        client=admin_client,
+        name=SERVICE_MONITOR_CRD_NAME,
+    )
+    if not service_monitor_crd.exists:
+        pytest.skip("ServiceMonitor CRD not installed; Limitador metrics scrape is optional on this cluster")
```
dbasunag: Same comment. if something is needed by MaaS tests and it is not present, then it should fail, instead of silent skips.

dbasunag (changes requested):
Please address all the comments in https://github.com/opendatahub-io/opendatahub-tests/pull/2241 in this one too. Also please refrain from creating duplicate prs. Let's wait for the base one to be reviewed, before putting the stacked ones.

tests/ai_gateway/models_as_a_service/observability/conftest.py
SB159: Thanks, Done

tests/ai_gateway/models_as_a_service/observability/conftest.py
SB159: Thanks, Done

tests/ai_gateway/models_as_a_service/observability/utils.py
```diff
@@ -0,0 +1,297 @@
...
+        namespace=monitoring_namespace,
+    )
+    try:
+        for _ in TimeoutSampler(
+            wait_timeout=timeout,
+            sleep=5,
+            func=lambda: service_monitor.exists,
+        ):
+            if service_monitor.exists:
+                return service_monitor
```
dbasunag: This one does not seem necessary. Instead of _ you should name it and check it instead.

tests/ai_gateway/models_as_a_service/observability/utils.py
```diff
@@ -0,0 +1,297 @@
...
+    except TimeoutExpiredError as exc:
+        raise AssertionError(
+            f"ServiceMonitor '{LIMITADOR_SERVICE_MONITOR_NAME}' not found in "
+            f"'{monitoring_namespace}' after maas-controller reconcile"
+        ) from exc
```
dbasunag: This is not necessary. Please raise the TimeoutExpiredError

tests/ai_gateway/models_as_a_service/observability/utils.py
```diff
@@ -0,0 +1,297 @@
...
+    )
+
+
+def validate_limitador_metrics_in_prometheus(
+    prometheus: Prometheus,
+    limitador_service_monitor: ServiceMonitor,
+    timeout: int = METRICS_POLL_TIMEOUT,
+) -> None:
+    """Verify Limitador metrics are visible in RHOAI observability Thanos via limitador_up."""
+    _ = limitador_service_monitor
```
dbasunag: Why this is needed?

tests/ai_gateway/models_as_a_service/observability/utils.py
```diff
@@ -0,0 +1,297 @@
...
+    assert limitador_namespace_selector_matches_kuadrant_namespace(
+        namespace_selector=namespace_selector,
+        kuadrant_namespace=kuadrant_namespace,
+    ), (
+        f"Expected ServiceMonitor namespaceSelector to match Kuadrant namespace "
+        f"'{kuadrant_namespace}', got namespaceSelector={namespace_selector}"
+    )
+
+
+def validate_limitador_metrics_in_prometheus(
```
dbasunag: This function is only calling validate_metrics_field. We don't need a wrapper here.

tests/ai_gateway/models_as_a_service/observability/utils.py
SB159: Updated.Thanks

tests/ai_gateway/models_as_a_service/observability/utils.py
SB159: Yes, Updated.

tests/ai_gateway/models_as_a_service/observability/utils.py
SB159: Not needed, removed.

tests/ai_gateway/models_as_a_service/observability/utils.py
SB159: Agreed, removed the wrapper; test calls validate_metrics_field directly.

tests/ai_gateway/models_as_a_service/observability/utils.py
```diff
@@ -0,0 +1,272 @@
...
+        namespace=monitoring_namespace,
+    )
+    for sampler in TimeoutSampler(
+        wait_timeout=timeout,
+        sleep=5,
+        func=lambda: service_monitor.exists,
+    ):
+        if sampler:
+            return service_monitor
+    return service_monitor
```
dbasunag: We would never reach this line.

SB159: Agreed — removed the unreachable return; function now returns ServiceMonitor | None.

dbasunag (approved):

fege (approved):
/lgtm

PR #2240, opened 2026-08-20
test(ai-hub): add MCP registry integration test

jonburdo:
Add backend integration coverage for registering an MCP Catalog entry in MLflow MCP Registry, deploying it through the dashboard BFF APIs, and verifying the resulting MCP endpoint is reachable on-cluster.

# Pull Request

## Summary

This adds a test to ensure we can register from MCP Catalog to MCP Registry and then deploy the mcp server.

Tested by running locally:
```
uv run --frozen pytest tests/ai_hub/mcp_servers/integration/test_mcp_registry_happy_path.py -vv -s --cluster-sanity-skip-check
```
after `oc login` on https://rh-ai.apps.rosa.jburdo2.53ms.p3.openshiftapps.com/ with RHOAI nightly 3.6 ea1 installed:
`quay.io/rhoai/rhoai-fbc-fragment:rhoai-3.6-ea.1@sha256:90ed438388cbed021f78f2634b35ee68f6019b64e7cb91bed92d5aba49ee6b8f`

## Related Issues

- Fixes: 
- JIRA: 

- [x] Locally
- [ ] Jenkins

## Additional Requirements

### DIFF
--- a/tests/ai_hub/mcp_servers/integration/conftest.py
+++ b/tests/ai_hub/mcp_servers/integration/conftest.py
@@ -1,0 +1,579 @@
+from __future__ import annotations
+
+from collections.abc import Generator
+from typing import Any
+from urllib.parse import quote
+from uuid import uuid4
+
+import portforward
+import pytest
+import structlog
+import yaml
+from kubernetes.dynamic import DynamicClient
+from ocp_resources.namespace import Namespace
+from ocp_resources.resource import ResourceEditor
+from requests import RequestException
+
+from tests.ai_hub.image_constants import AiHubImages
+from tests.ai_hub.mcp_servers.config.utils import get_mcp_catalog_sources
+from tests.ai_hub.mcp_servers.integration.constants import (
+    DASHBOARD_LOCAL_PORT,
+    DASHBOARD_REMOTE_PORT,
+    DEPLOYABLE_MCP_EXPECTED_TOOL_NAMES,
+    DEPLOYABLE_MCP_LICENSE,
+    DEPLOYABLE_MCP_PATH,
+    DEPLOYABLE_MCP_PORT,
+    DEPLOYABLE_MCP_PROVIDER,
+    DEPLOYABLE_MCP_REGISTRY_NAMESPACE,
+    DEPLOYABLE_MCP_SERVER_DESCRIPTION,
+    DEPLOYABLE_MCP_SERVER_TITLE,
+    DEPLOYABLE_MCP_SOURCE_NAME_PREFIX,
+    DEPLOYABLE_MCP_TAGS,
+    DEPLOYABLE_MCP_TIMESTAMP_MS,
+    DEPLOYABLE_MCP_VERSION,
+    MLFLOW_BFF_PATH,
+    MLFLOW_BFF_REGISTER_PATH,
+    MODEL_REGISTRY_BFF_PATH,
+)
+from tests.ai_hub.mcp_servers.integration.utils import (
+    deploy_yaml_from_spec,
+    extract_data_envelope,
+    extract_deploy_spec,
+    mcp_server_name_path_segment,
+    normalize_registry_server_json,
+    normalize_registry_tools,
+    request_json,
+    wait_for_ready_deployment,
+)
+from tests.ai_hub.utils import execute_get_command_with_retry, wait_for_catalog_api
+from utilities.resources.pod import Pod as UtilPod
+
+LOGGER = structlog.get_logger(name=__name__)
+
+
+def _run_cleanup_request(step: str, **request_kwargs: Any) -> None:
+    """Run a cleanup HTTP request, log failures, and continue teardown."""
+    try:
+        request_json(expect_json=False, **request_kwargs)
+    except (AssertionError, RequestException) as expt:
+        LOGGER.warning("Cleanup request failed", step=step, error=str(expt))
+
+
+def _delete_mcp_access_endpoint(
+    dashboard_api_base_url: str,
+    encoded_name: str,
+    endpoint_id: str | None,
+    endpoint_url: str,
+    headers: dict[str, str],
+    workspace_name: str,
+) -> None:
+    """Delete a created MLflow access endpoint by id or best-effort lookup."""
+    if endpoint_id:
+        _run_cleanup_request(
+            step="delete mlflow access endpoint by id",
+            method="DELETE",
+            url=(
+                f"{dashboard_api_base_url}{MLFLOW_BFF_PATH}/"
+                f"{encoded_name}/endpoints/{quote(endpoint_id, safe='')}"
+                f"?workspace={quote(workspace_name, safe='')}"
+            ),
+            headers=headers,
+            verify=False,
+            expected_status_codes={200, 204, 404},
+        )
+        return
+
+    endpoint_list_payload = request_json(
+        method="GET",
+        url=(
+            f"{dashboard_api_base_url}{MLFLOW_BFF_PATH}/{encoded_name}/endpoints"
+            f"?workspace={quote(workspace_name, safe='')}"
+        ),
+        headers=headers,
+        verify=False,
+        expected_status_codes={200, 404},
+        expect_json=False,
+    )
+    endpoint_list_data = endpoint_list_payload.get("data", endpoint_list_payload)
+    endpoint_candidates: list[dict[str, Any]] = []
+    if isinstance(endpoint_list_data, list):
+        endpoint_candidates = [item for item in endpoint_list_data if isinstance(item, dict)]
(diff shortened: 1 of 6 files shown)


tests/ai_hub/mcp_servers/integration/test_mcp_registry_happy_path.py
```diff
@@ -0,0 +1,286 @@
...
+                    "displayName": "QE MCP registry deployment",
+                    "registryServer": server_name,
+                    "registryVersion": DEPLOYABLE_MCP_VERSION,
+                    "image": deploy_spec["source"]["containerImage"]["ref"],
+                    "yaml": deploy_yaml_from_spec(deploy_spec=deploy_spec),
+                }
+            },
+        )
+        deployment_data = extract_data_envelope(payload=deployment_payload, url=deployment_url)
+        deployment_name = deployment_data["name"]
```
dkuc: **Resource leak: cleanup flags/names captured only after `extract_data_envelope`**

The register POST (line 145) and deployment POST (line 165) can return `201` — meaning the resource is genuinely created — but if `extract_data_envelope` then raises `AssertionError` (envelope lacks a dict `data`), `created_registry_version` stays `False` / `deployment_name` stays `None`. The `finally` block then skips deletion, orphaning the registry version and MCP deployment on the cluster.

Set the cleanup flag/name *before* parsing the response envelope, so teardown runs whenever the POST returned 201.

tests/ai_hub/mcp_servers/integration/utils.py
```diff
@@ -0,0 +1,155 @@
...
+        expected_status_codes={200},
+    )
+
+    last_seen: dict[str, Any] = {}
+    for sample in sampler:
+        last_seen = extract_data_envelope(payload=sample, url=deployment_url)
+        if deployment_ready(deployment=last_seen):
+            return last_seen
+
+    pytest.fail(f"MCP deployment {deployment_name} did not become Ready: {last_seen}")
```
dkuc, line 166: **Unreachable `pytest.fail` on deployment timeout**

`TimeoutSampler` raises `TimeoutExpiredError` at timeout rather than ending the `for sample in sampler` loop via `StopIteration`. As a result, this `pytest.fail(...)` with `last_seen` never runs — a timeout surfaces as a generic `TimeoutExpiredError` with no captured deployment state.

Wrap the sampler loop in `try/except TimeoutExpiredError` (or restructure) so the diagnostic with `last_seen` actually fires on timeout.

dkuc (approved):

dkuc:
Posted a couple edge case findings - I'm fine skipping them - up to you.  Otherwise, Looks good! I was able to successfully run the test.  Approved.

utilities/resources/mcp_server.py
```diff
@@ -0,0 +1,43 @@
+# Generated using https://github.com/RedHatQE/openshift-python-wrapper/blob/main/scripts/resource/README.md
```
lugi0, line 1: this would need to go to the python wrapper

tests/ai_hub/mcp_servers/integration/conftest.py
```diff
@@ -0,0 +1,191 @@
...
+    DEPLOYABLE_MCP_VERSION,
+)
+from tests.ai_hub.utils import execute_get_command_with_retry, wait_for_catalog_api
+from utilities.resources.pod import Pod as UtilPod
+
+LOGGER = structlog.get_logger(name=__name__)
+
+
+@pytest.fixture(scope="class")
+def mcp_registry_test_metadata(model_namespace: Namespace) -> dict[str, Any]:
```
lugi0, line 142: I think most of these variables could be defined as constants in the relevant file and then you'd only append the suffix as needed here

tests/ai_hub/mcp_servers/integration/conftest.py
```diff
@@ -0,0 +1,191 @@
...
+        "server_name": server_name,
+        "source_id": source_id,
+        "source_name": source_name,
+        "yaml_catalog_path": yaml_catalog_path,
+        "catalog_yaml": catalog_yaml,
+    }
+
+
+@pytest.fixture(scope="class")
+def dashboard_api_base_url(dashboard_pod: UtilPod) -> Generator[str, Any]:
```
lugi0, line 225: why do you need to go through the BFF?

tests/ai_hub/mcp_servers/integration/conftest.py
```diff
@@ -0,0 +1,191 @@
...
+    ).get("size", 0)
+
+    with ResourceEditor(patches={catalog_config_map: patches}):
+        wait_for_catalog_api(
+            endpoint="mcp_servers",
+            item_name="MCP servers",
+            url=mcp_catalog_rest_urls[0],
+            headers=model_registry_rest_headers,
+            previous_size=pre_patch_size,
+            expected_size=pre_patch_size + 1,
```
lugi0, line 291: do you only want to check that it has increased in size or also that the entry is correct?

tests/ai_hub/mcp_servers/integration/test_mcp_registry_happy_path.py
```diff
@@ -0,0 +1,286 @@
...
+    mcp_catalog_rest_urls: list[str],
+    mcp_registry_test_metadata: dict[str, Any],
+    model_namespace: Namespace,
+    model_registry_rest_headers: dict[str, str],
+    unused_tcp_port_factory: Callable[[], int],
+) -> None:
+    """
+    Given: A deployable MCP server exists in the live AI Hub MCP catalog
+    When: Its catalog metadata is registered in MLflow and the registered version is deployed through the
+          dashboard BFF APIs
```
lugi0, line 41: same as above, do you need to go through the BFF? It's unusual for our tests

tests/ai_hub/mcp_servers/integration/test_mcp_registry_happy_path.py
```diff
@@ -0,0 +1,286 @@
...
+    [
+        pytest.param(
+            {"name": "test-mcp-registry-happy-path"},
+            marks=[pytest.mark.install, pytest.mark.tier2],
+            id="test_mcp_registry_backend_happy_path",
+        )
+    ],
+    indirect=True,
+)
+def test_register_and_deploy_mcp_server_from_catalog_backend_happy_path(
```
lugi0, line 32: This test is very long and has a lot of asserts compared to our usual style. 
Would it be possible to split this into multiple tests, or alternatively to abstract away some of the logic in fixtures? 

If I understand correctly everything before the try is setup and sanity checks on the running resources, and if so it would be ideal to have these steps in fixtures rather than in the test. 

The `finally` block is also a bit unusual, we'd expect teardown to be performed by whatever fixture first created the resource

tests/ai_hub/mcp_servers/integration/test_mcp_registry_happy_path.py
jonburdo: Good catch, fixed

tests/ai_hub/mcp_servers/integration/utils.py
jonburdo: Fixed

tests/ai_hub/mcp_servers/integration/conftest.py
jonburdo: The goal of the test is to cover the same dashboard-driven register/deploy flow the product uses, so it goes through the dashboard BFF rather than bypassing it with lower-level direct calls

tests/ai_hub/mcp_servers/integration/test_mcp_registry_happy_path.py
jonburdo: I agreed with the fixture/teardown part and refactored most of the orchestration into fixtures so the test body is much smaller and resource cleanup is owned closer to creation. I kept it as one test, though, because the catalog->registry->deploy->endpoint path is the single end-to-end behavior this PR is trying to validate.

tests/ai_hub/mcp_servers/integration/conftest.py
jonburdo: Good point. This size check tells us the catalog API has picked up the source update. I’ve also added a dedicated `catalog_server_details` fixture that resolves the exact injected entry by `name` and `source_id` and validates its detail/tools payload before the test uses it

dkuc (approved):

lugi0 (approved):

utilities/resources/mcp_server.py
jonburdo: Created a PR here: https://github.com/RedHatQE/openshift-python-wrapper/pull/2796

PR #2247, opened 2026-08-21
fix: add s390x support

bchakraborty1:
# Pull Request

## Summary

Include s390x support for odh-tests

## Related Issues

- Fixes: 
- JIRA: 

- [ ] Locally
- [ ] Jenkins

## Additional Requirements

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -86,12 +86,17 @@ project-excludes = [
 # Start permissive to match mypy behavior
 missing-import = false  # Covered by ignore-missing-imports
 
+[tool.pytest.ini_options]
+markers = [
+    "skip_on_arch_s390x: deselect tests that are not supported on s390x",
+]
+
 [dependency-groups]
 dev = [
     "ipdb>=0.13.13",
     "ipython>=8.12.3",
     "pre-commit>=4.3.0",
-    "pyrefly>=1.0",
+    "pyrefly>=1.0; platform_machine != 's390x'",
 ]
 
 [project]
@@ -118,7 +123,7 @@ dependencies = [
     "pytest-progress",
     "structlog>=24.1.0",
     "pyyaml",
-    "pyarrow>=23.0.0",
+    "pyarrow>=23.0.0; platform_machine != 's390x'",
     "tenacity",
     "types-requests>=2.32.0.20241016",
     "requests",
@@ -149,11 +154,11 @@ dependencies = [
     "pytest>=9.0.0",
     "huggingface-hub>=1.2.3",
     "openai>=2.36",
-    "ragas>=0.4",
-    "numpy>=2.1.0,<2.5.0",
-    "scipy>=1.13.0,<1.18.0",
+    "ragas>=0.4; platform_machine != 's390x'",
+    "numpy>=2.1.0,<2.5.0; platform_machine != 's390x'",
+    "scipy>=1.13.0,<1.18.0; platform_machine != 's390x'",
     "langchain-community<0.4.2",
-    "ipykernel>=7.3.0",
+    "ipykernel>=7.3.0; platform_machine != 's390x'",
     "boto3>=1.43.33",
 ]
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -64,6 +64,7 @@ from utilities.infra import (
     download_helm_console_cli,
     download_oc_console_cli,
     get_cluster_authentication,
+    get_machine_platform,
     get_openshift_token,
     login_with_user_password,
     update_configmap_data,
@@ -91,6 +92,33 @@ pytest_plugins = [
 @pytest.fixture(scope="session")
 def admin_client() -> DynamicClient:
     return get_client()
+
+
+def pytest_collection_modifyitems(config: pytest.Config, items: list[pytest.Item]) -> None:
+    """Deselect tests that are not supported on the current architecture."""
+    client = get_client()
+    nodes = list(Node.get(dyn_client=client))
+
+    if not nodes:
+        return
+
+    cluster_architecture = nodes[0].instance.status.nodeInfo.architecture
+
+    if cluster_architecture != "s390x":
+        return
+
+    deselected = []
+    remaining = []
+
+    for item in items:
+        if "skip_on_arch_s390x" in item.keywords:
+            deselected.append(item)
+        else:
+            remaining.append(item)
+
+    if deselected:
+        config.hook.pytest_deselected(items=deselected)
+        items[:] = remaining
 
 
 @pytest.fixture(scope="session")
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -64,7 +64,6 @@ from utilities.infra import (
     download_helm_console_cli,
     download_oc_console_cli,
     get_cluster_authentication,
-    get_machine_platform,
     get_openshift_token,
     login_with_user_password,
     update_configmap_data,
--- a/tests/ogx/vector_io/conftest.py
+++ b/tests/ogx/vector_io/conftest.py
@@ -7,7 +7,6 @@ import pytest
 import structlog
 from ogx_client import OgxClient
 from ogx_client.types.vector_store import VectorStore
-from ragas import SingleTurnSample
 
 from tests.ogx.constants import (
     RAGAS_EVAL_MAX_TOKENS,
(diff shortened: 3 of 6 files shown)


modassarrana89-new (commented):
LGTM

Makefile
```diff
@@ -18,13 +18,14 @@ build:
 		CHECKSUM=$$(printf '%s' "$$MANIFEST" | shasum -a 256 | cut -d' ' -f1); \
 		echo "Image manifest generated (sha256:$$CHECKSUM)"; \
 		$(IMAGE_BUILD_CMD) build \
```
sheltoncyril, line 20: An AI-assisted review flagged this and I wanted to check before treating it as actionable.

`$(IMAGE_BUILD_CMD) build --platform linux/amd64,linux/ppc64le,linux/s390x -t ... .` looks like it fails on both backends — podman needs `--manifest` (not `-t`) for a multi-platform list, and docker's default driver rejects multi-platform without buildx — which would break `make build` for everyone including x86. Is a manifest/buildx path intended here? Thanks!

modassarrana89-new: Yes i agree to this. It would create a problem 
I also wanna know the intention to have multi-arch option in both if else .

bchakraborty1: Hi @sheltoncyril @modassarrana89-new 
I did check this code and I agree, it does cause an issue because we don't want build breaking for all archs. I am seeing that this change is not needed. Will test and upload another patch.

tests/ogx/vector_io/conftest.py
```diff
@@ -7,6 +8,13 @@
 import structlog
 from ogx_client import OgxClient
 from ogx_client.types.vector_store import VectorStore
+
+if platform.machine() in ("ppc64le", "s390x"):
+    pytest.skip(
+        "ragas is not supported on this architecture",
+        allow_module_level=True,
+    )
```
dbasunag: This would skip if the container your tests are running from, are `"ppc64le", "s390x"`, instead of skipping on clusters with node `"ppc64le", "s390x"`, no?

Dockerfile
```diff
@@ -1,3 +1,24 @@
...
+     cd / && rm -rf /tmp/must-gather-clean && \
+     dnf clean all && rm -rf /var/cache/dnf; \
+   elif [[ $(uname -m) == "x86_64" ]]; then \
+     curl -sSL "https://github.com/openshift/must-gather-clean/releases/download/v0.0.4/must-gather-clean-linux-amd64.tar.gz" \
+          --output /tmp/must-gather-clean.tar.gz && \
+     echo "29e90cc94b593c6bf992923790a975ecf54bc1579c1e999b4acce90167b41235  /tmp/must-gather-clean.tar.gz" | sha256sum --check --strict - && \
+     tar xzf /tmp/must-gather-clean.tar.gz -C /tmp && \
+     rm -f /tmp/must-gather-clean.tar.gz && \
+     mkdir -p /out && mv /tmp/must-gather-clean /out/must-gather-clean; \
+   fi
```
dbasunag: Why this change is needed?

dbasunag (changes requested):

Dockerfile
bchakraborty1: The original Dockerfile hardcoded must-gather-clean-linux-amd64.tar.gz which fails on s390x since no pre-built binary exists for that architecture. This multi-stage builder compiles it from source on s390x while keeping the existing pre-built binary path for x86_64.

tests/ogx/vector_io/conftest.py
bchakraborty1: You are right, if the tests are triggered from a different arch container, the skip would not work. 
Do you have a recommended pattern for skipping based on cluster node architecture?

dbasunag: Yes, please check the tests/ai_hub/conftest.py, specially pytest_sessionstart

Dockerfile
dbasunag: We can't take this change. We would need to explore other alternatives here. For Z and power do you need to run from z/power container? I want to understand why.

tests/ogx/vector_io/conftest.py
bchakraborty1: Hi @dbasunag I have uploaded a new patch, that has the corrected cluster node arch code.

Dockerfile
bchakraborty1: We do not need to have a specific Z container, we can compile and run from the original Dockerfile itself. This change is not needed and is removed in the latest patch.

bchakraborty1:
@dbasunag @sheltoncyril @modassarrana89-new I have updated the changes, tested on s390x. Requesting your review.

pyproject.toml
```diff
@@ -86,12 +86,17 @@ project-excludes = [
 # Start permissive to match mypy behavior
 missing-import = false  # Covered by ignore-missing-imports
 
+[tool.pytest.ini_options]
+markers = [
+    "skip_on_arch_s390x: deselect tests that are not supported on s390x",
+]
+
```
dbasunag, line 93: Not needed.

tests/conftest.py
```diff
@@ -93,6 +93,33 @@ def admin_client() -> DynamicClient:
     return get_client()
 
 
+def pytest_collection_modifyitems(config: pytest.Config, items: list[pytest.Item]) -> None:
```
dbasunag, line 96: Should be in tests/ogx/vector_io/conftest.py

tests/ogx/vector_io/test_vector_stores_evaluation.py
```diff
@@ -2,8 +2,18 @@
...
+    from ragas import EvaluationDataset, SingleTurnSample, evaluate
+    from ragas.metrics import AnswerRelevancy, ContextPrecision, ContextRecall, Faithfulness
+except ImportError:
+    EvaluationDataset = None
+    SingleTurnSample = Any
+    evaluate = None
+    AnswerRelevancy = None
+    ContextPrecision = None
+    ContextRecall = None
+    Faithfulness = None
```
dbasunag, line 16: I don't understand why we still need this.

tests/ogx/vector_io/conftest.py
```diff
@@ -109,13 +108,15 @@ def ragas_samples(
...
-) -> list[SingleTurnSample]:
+) -> list[Any]:
     """Build RAGAS evaluation samples by querying the RAG pipeline for each ground-truth QA pair.
 
     Uses the Responses API with the file_search tool against the vector store,
     mirroring a real-world RAG scenario.  The number of questions sent to the
     LLM is capped by ``RAGAS_MAX_SAMPLES`` (env var, default 5).
     """
+    from ragas import SingleTurnSample
+
```
dbasunag, line 119: Same question.

tests/conftest.py
```diff
@@ -93,6 +93,33 @@ def admin_client() -> DynamicClient:
...
+    if not nodes:
+        return
+
+    cluster_architecture = nodes[0].instance.status.nodeInfo.architecture
+
+    if cluster_architecture != "s390x":
+        return
+
+    deselected = []
+    remaining = []
```
dbasunag, line 110: These should be skipped on collect only and setup plan.

dbasunag (changes requested):

tests/ogx/vector_io/test_vector_stores_evaluation.py
bchakraborty1: I am seeing an issue where Python is trying to import ragas even before `pytest_collection_modifyitems` can deselect it.

This is causing the setup/collect to fail with the error of "ragas not found".

tests/conftest.py
bchakraborty1: acknowledged.

tests/conftest.py
bchakraborty1: ack

pyproject.toml
bchakraborty1: Ack.

tests/ogx/vector_io/conftest.py
bchakraborty1: Added a try catch in tests/ogx/vector_io/test_vector_stores_evaluation.py while importing ragas.

When we try to setup/test locally on s390x machine, ragas fails to import since it is not supported for this arch. The issue with the imports in tests/ogx/vector_io/test_vector_stores_evaluation.py is that the file loads before we get to the pytest skip marker and the module imports get collected. This causes the setup to fail with error "module 'ragas' not found". To avoid this, I added a try catch that accepts lazy imports so that the module does not break while I am running uv run.

bchakraborty1: I added the imports in the files inside the test cases so that it does not affect us as we are planning on skipping / not collecting these tests.

dbasunag (changes requested):
If jenkins works without this, then we don't need this pr just to accommodate local runs from various architecture machines. We never claim to support running from different architecture machines, our goal is only to support testing against different supported clusters and currently as far as I know that works. Please check and close it if jenkins works.

dbasunag:
We already saw that jenkins does not require s390x containers. Can you please close this pr?

PR #2250, opened 2026-08-21
fix(evalhub): make evalhub_mcp_mt_cr fixture set spec.mcp and wait for readiness

kpunwatk:
The EvalHub wrapper's to_dict() rebuilds res["spec"] from its known constructor attributes (collections/database/env/otel/providers/replicas) on every create() call, so mutating evalhub.res["spec"]["mcp"] after construction was silently discarded before the object was ever posted. Build the resource via kind_dict instead so spec.mcp actually reaches the server.

Also replace evalhub.wait(), which only checks object existence, with a poll of status.ready so the fixture doesn't yield before the operator has finished provisioning MCP resources.

Fixes test_evalhub_mcp_configmap_exists intermittently failing with ResourceNotFoundError for the operator-provisioned MCP ConfigMap.

# Pull Request

## Related Issues

- Fixes: 
- JIRA: 

- [ ] Locally
- [ ] Jenkins

## Additional Requirements

## Summary by CodeRabbit

- **Tests**
  - Improved automated validation for MCP-enabled EvalHub resources.
  - Readiness checks now monitor reported resource status and handle timeouts more reliably.
  - Tests fail promptly when a resource reports an error, making failures easier to identify.
  - Resource configuration is created explicitly to better reflect the intended MCP-enabled setup.
  - These improvements provide more accurate and dependable validation of resource creation and readiness.


sheltoncyril (dismissed):

tests/ai_safety/evalhub/mcp/conftest.py
```diff
@@ -108,29 +108,39 @@ def evalhub_mcp_mt_cr(
...
-            }
-        ],
-    }
-
-    with evalhub:
-        evalhub.wait(timeout=300)
+    ) as evalhub:
+        # evalhub.wait() only checks that the object exists, not that the operator
+        # has finished reconciling it, so poll status.ready instead.
+        for sample in TimeoutSampler(wait_timeout=300, sleep=2, func=lambda: evalhub.instance.status.get("ready")):
```
ruivieira: @kpunwatk This only breaks on "True". If `status.ready` is "False" or "Unknown" permanently, it will wait entire 300 seconds before failing with a timeout. Should it actually break with other conditions? Something like

```python
for sample in TimeoutSampler(wait_timeout=300, sleep=2, func=lambda: evalhub.instance.status.get("ready")):
    if sample == "True":
        break
    if sample == "False":
        phase = evalhub.instance.status.get("phase", "unknown")
        pytest.fail(f"EvalHub {EVALHUB_MCP_CR_NAME} entered ready=False (phase={phase})")
```
wdyt?

ruivieira (dismissed):

tests/ai_safety/evalhub/mcp/conftest.py
kpunwatk: Thanks @ruivieira good point on checking sample == "False" would break the fixture on literally every run not just failures. I checked the CRD schema: phase enum: ['Pending', 'Ready', 'Error']

The actual terminal-failure signal is phase == "Error", not ready == "False" — Pending and Error are both non-ready but only one of them is a dead end. I'd write it like this instead:
 if sample.get("phase") == "Error":

  This keeps your intent (fail fast on genuine errors instead of burning the full 300s) but polls on phase for the fail-fast check rather than ready, so it won't trip during normal Pending startup.

kpunwatk: added 3d43d64

sheltoncyril (approved):

sheltoncyril:
Re `tests/ai_safety/evalhub/mcp/conftest.py`:

An AI-assisted review flagged this and I wanted to check before treating it as actionable.

The new polling reads `evalhub.instance.status` then `sample.get("ready")`, but on a freshly created CR `.status` is often None for the first cycles, which looks like it'd raise `AttributeError` and crash the fixture (the lifecycle test elsewhere guards with `(... .status or {})`). Also success keys on `status.ready == "True"` while failure keys on `status.phase` — is `ready` a confirmed field? Thanks!

PR #2259, opened 2026-08-24
fix(ai-gateway): poll for inference readiness in subscription enforcement test

liangwen12year:
## Summary
- `test_subscribed_user_gets_200` sent a single POST immediately after class-scoped fixture setup and asserted 200, but the Envoy data-plane may not have propagated the upstream cluster yet — resulting in sporadic 503 "no healthy upstream" failures in CI
- Replace the single `request_session_http.post()` with `poll_expected_status()` (retry loop, 240s timeout, 5s poll interval) — the same pattern already used by `test_maas_auth_enforcement.py` and `test_cascade_deletion.py`

## Root cause
When `TestSubscriptionEnforcementTinyLlama` runs after another test class (e.g. component_health) that deploys and then tears down the `llm` namespace, the subscription enforcement fixtures recreate the namespace and redeploy the model. The LLMInferenceService and MaaSSubscription report `Ready=True` from the control plane, but the Envoy upstream cluster needs additional time to become routable. In the CI log, the test ran ~1 second after the subscription became Ready and hit 503.

## Test plan
- [ ] Verify `test_subscribed_user_gets_200` passes in CI when run after component_health teardown
- [ ] Confirm no regressions in other subscription enforcement tests

### DIFF
--- a/tests/ai_gateway/models_as_a_service/maas_subscription/test_maas_sub_enforcement.py
+++ b/tests/ai_gateway/models_as_a_service/maas_subscription/test_maas_sub_enforcement.py
@@ -6,6 +6,7 @@ import structlog
 
 from tests.ai_gateway.models_as_a_service.maas_subscription.utils import (
     chat_payload_for_url,
+    poll_expected_status,
 )
 from tests.ai_gateway.models_as_a_service.utils import build_maas_headers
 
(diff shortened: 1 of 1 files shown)


SB159 (commented):
Looks good. Please fix the DCO and update the branch

SB159 (approved):
/lgtm

sheltoncyril (approved):

PR #2268, opened 2026-08-25
fix(ogx): treat file_citation annotations as best-effort in upgrade RAG test

Artemon-line:
In upgrade RAG test, treat file_citation annotations as best-effort logging a warning if empty rather than hard failing. Citation marker emission varies by LLM model compliance.


uv.lock
jgarciao, line 1: Are the modifications in this file needed?

jgarciao (approved):

PR #2293, opened 2026-08-27
fix(llama_stack): relax annotations assertion in 3.4 upgrade RAG test

Artemon-line:
When file_search_call completes with retrieved chunk results, smaller LLMs do not consistently emit inline citation markers in their output text. Relaxing the strict assertion on non-empty annotations prevents false-positive test failures in 3.4 upgrade runs.

### DIFF
--- a/tests/llama_stack/vector_io/upgrade/test_upgrade_vector_store_rag.py
+++ b/tests/llama_stack/vector_io/upgrade/test_upgrade_vector_store_rag.py
@@ -50,10 +50,10 @@ def _assert_minimal_rag_response(
             if item_annotations:
                 annotations.extend(item_annotations)
 
-    assert annotations, "Response should contain file_citation annotations when file_search returns results"
-    assert any(annotation.type == "file_citation" for annotation in annotations), (
-        "Expected at least one file_citation annotation in response output"
-    )
+    if annotations:
+        assert any(annotation.type == "file_citation" for annotation in annotations), (
+            "Expected at least one file_citation annotation in response output"
+        )
 
 
 @pytest.mark.parametrize(


tests/llama_stack/vector_io/upgrade/test_upgrade_vector_store_rag.py
```diff
@@ -50,10 +50,10 @@ def _assert_minimal_rag_response(
...
                 annotations.extend(item_annotations)
 
-    assert annotations, "Response should contain file_citation annotations when file_search returns results"
-    assert any(annotation.type == "file_citation" for annotation in annotations), (
-        "Expected at least one file_citation annotation in response output"
-    )
+    if annotations:
+        assert any(annotation.type == "file_citation" for annotation in annotations), (
+            "Expected at least one file_citation annotation in response output"
+        )
```
dbasunag, line 56: Would it not silently pass if the annotation is not present? Do we expect it to be present or not?

jgarciao: file citations are a little tricky and they depend a lot on the LLM being used. It is for @Artemon-line and his team to decide about this

jgarciao (approved):

tests/llama_stack/vector_io/upgrade/test_upgrade_vector_store_rag.py
Artemon-line: Yes, it will bypass if `annotations` is empty. 

For upgrade verification, confirming that vector store retrieval succeeded across the upgrade (`file_search_call.status == "completed"` and `file_search_call.results` is non-empty) is sufficient. `file_citation` annotations depend on LLM non-determinism, so strictly asserting them leads to false negatives.

PR #2295, opened 2026-08-27
fix(upgrade): use subset comparison for LLMISVC config ref pins

bartoszmajsak:
The upgrade test for LLMInferenceService configuration pins requires exact equality between pre-upgrade and post-upgrade annotation lists. RHOAI 3.5 adds three new well-known configuration pins during reconciliation, so the post-upgrade list is a strict superset of the pre-upgrade list and the test fails on every 3.4-to-3.5 upgrade.

Switches to a subset comparison - each pre-upgrade annotation key must keep its value after the upgrade, but additional post-upgrade keys are accepted. A missing or changed pre-upgrade pin still fails the test with a clear diagnostic showing which pins changed and what the expected values were.

Baselines captured by older test code (before this patch) fall back to name-set comparison so in-flight upgrade runs are not broken by a missing field.

Fixes [RHOAIENG-88337](https://redhat.atlassian.net/browse/RHOAIENG-88337) [RHOAIENG-88338](https://redhat.atlassian.net/browse/RHOAIENG-88338)

[RHOAIENG-88337]: https://redhat.atlassian.net/browse/RHOAIENG-88337?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RHOAIENG-88338]: https://redhat.atlassian.net/browse/RHOAIENG-88338?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved LLMInferenceService upgrade validation to accurately preserve and verify configuration references.
  * Upgrade checks now recognize newly added references while ensuring existing references remain unchanged.
  * Added compatibility handling for older upgrade baselines and configurations with empty references.

### DIFF
--- a/tests/model_serving/model_server/upgrade/utils.py
+++ b/tests/model_serving/model_server/upgrade/utils.py
@@ -1253,11 +1253,7 @@ def get_llmisvc_config_ref_pins(llmisvc: LLMInferenceService) -> dict[str, str]:
     """Get config ref annotation key-value pairs from LLMInferenceService status."""
     _CONFIG_REF_ANNOTATION_PREFIX = "serving.kserve.io/config-llm-"
     annotations = getattr(llmisvc.instance.status, "annotations", None) or {}
-    return {
-        key: value
-        for key, value in annotations.items()
-        if key.startswith(_CONFIG_REF_ANNOTATION_PREFIX) and value
-    }
+    return {key: value for key, value in annotations.items() if key.startswith(_CONFIG_REF_ANNOTATION_PREFIX) and value}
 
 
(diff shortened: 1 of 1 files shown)


pierDipi (approved):
/lgtm

dbasunag (approved):

mwaykole (approved):
/lgtm

PR #2358, opened 2026-09-10
feat(ci): add workflow to cut a release branch and build its image

sheltoncyril:
## Summary

Adds a manually triggered (`workflow_dispatch`) workflow that cuts a release branch and pushes a matching container image tag. During code freeze, this replaces creating the branch by hand and waiting for a merge to produce the image.

## Changes

- **`.github/workflows/cut-release-branch.yml`** (new) — creates the release branch, then builds and pushes `<registry>/<image>:<tag>`.

  | Input | Default | Description |
  | --- | --- | --- |
  | `branch_name` | (required) | Release branch to create, for example `3.6`. |
  | `source_ref` | `main` | Branch, tag or commit SHA to cut from. |
  | `image_tag` | `branch_name` | Image tag override. |
  | `image_name` | `opendatahub-tests` | Image name, without the registry prefix. |
  | `registry` | `quay.io/opendatahub` | Registry and namespace to push to. |
  | `build_image` | `true` | Build and push the image after the branch is created. |

  Branch name and image tag are validated, and `main` is rejected as a target. If the branch already exists, creation is skipped with a warning and the image is still rebuilt, so the workflow is safe to re-run.

- **`.github/workflows/build-push-image.yml`** (new) — reusable `workflow_call` workflow holding the build and push logic that previously lived only in the on-merge workflow, including the `io.opendatahub.tests.required-images` manifest labels. Without this, a duplicated build for release branches would have produced images missing those labels.

- **`.github/workflows/build-push-container-on-merge.yml`** — refactored to call the reusable workflow. Tag resolution (`main` to `latest`, otherwise the base ref) and the PR status comment are unchanged.

- **`docs/GITHUB_WORKFLOWS.md`** — documents the new workflow, its inputs, and how to call the reusable workflow.

## Notes for reviewers

- Branch creation uses `RHODS_CI_BOT_PAT` and falls back to `GITHUB_TOKEN`. If branch creation is restricted by branch protection, the PAT needs `contents: write` on this repository.
- Post-cut steps such as updating version constants or branch-specific configuration are not automated here. Happy to add them in a follow-up if that is wanted.
- A full release process document is still to come.

## Testing

`workflow_dispatch` workflows can only be run from the default branch, so this cannot be exercised end to end until it is merged. YAML syntax was validated locally, and the reusable workflow is byte-for-byte the same build and push steps that run today on merge.

## Summary by CodeRabbit

- **Documentation**
  - Added a release process guide covering release branches, container image tags, test-pipeline registration, naming conventions, verification, backports, and troubleshooting.
  - Added README guidance linking to the release process.
  - Expanded workflow documentation with input normalization and validation, image tag fallback rules, conditional publishing, reusable-workflow configuration, registry secrets, and fork-based testing instructions.

### DIFF
--- a/.github/workflows/build-push-container-on-merge.yml
+++ b/.github/workflows/build-push-container-on-merge.yml
@@ -10,86 +10,32 @@ permissions:
   issues: write
 
 jobs:
-  if_merged:
+  resolve-tag:
     if: github.event.pull_request.merged == true
     runs-on: ubuntu-latest
+    outputs:
+      tag: ${{ steps.tag.outputs.tag }}
     steps:
-      - name: Checkout Repository
-        uses: actions/checkout@v7
-        with:
-          ref: ${{ github.event.pull_request.base.ref }}
-      - name: Set env TAG
-        run: |
-          if [ ${{ github.event.pull_request.base.ref }} == "main" ]; then
-              echo "TAG=latest" >> $GITHUB_ENV
-          else
-              echo "TAG=${{ github.event.pull_request.base.ref }}" >> "$GITHUB_ENV"
-          fi
-      - name: Install uv
-        uses: astral-sh/setup-uv@v10.0.1
-
-      - name: Generate image manifest
-        id: manifest
-        continue-on-error: true
-        run: echo "json=$(uv run python scripts/generate_image_manifest.py --compact)" >> "$GITHUB_OUTPUT"
-
-      - name: Compute manifest checksum
-        id: checksum
-        if: steps.manifest.outcome == 'success'
+      - name: Set image tag
+        id: tag
         env:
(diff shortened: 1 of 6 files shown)


.github/workflows/build-push-container-on-merge.yml
```diff
@@ -10,86 +10,34 @@ permissions:
   issues: write
 
 jobs:
-  if_merged:
```
dbasunag, line 13: We should leave this one as it is. This automation works, and adding changes here is not necessary for the automation you are trying to do.

sheltoncyril:
Good point, reverted `build-push-container-on-merge.yml` to its original inline steps in b415cc3f. `build-push-image.yml` is now only used by `cut-release-branch.yml`, so the on-merge automation is untouched.

PR #2361, opened 2026-09-10
[RHAI-196] Use CPU LLMInferenceServiceConfig in llm-d tests

threcc:
# Pull Request

## Summary

- Update CpuConfig to discover the versioned CPU configuration and set it through baseRefs. 
- Remove the hardcoded vLLM CPU image and CPU KV-cache environment setting so llm-d tests inherit these values from the cluster configuration.

## Related Issues

- Upstream PR: https://github.com/opendatahub-io/kserve/pull/1869
- JIRA: https://redhat.atlassian.net/browse/RHAI-196

## Tests

- [x] Locally - manually created the LLMInferenceServiceConfig in local cluster. before mergin this PR full tests with  the real LLMInferenceServiceConfig should executed
- [] Jenkins

## Summary by CodeRabbit

- **Tests**
  - Improved CPU inference test configuration by dynamically discovering compatible, versioned service configurations.
  - Increased the default wait time for inference test operations.
  - Added support for matching service configurations when accelerator information is unavailable.
  - Improved diagnostics when accelerator filtering is not applied.
  - Updated disconnected-cluster handling for models stored using Hugging Face locations.
  - Refined configuration matching to require the specified workload topology, improving test selection accuracy.

## Summary by CodeRabbit

- **Tests**
  - Improved model-serving test configuration discovery by selecting compatible, versioned CPU inference service configurations based on deployment topology.
  - Added filtering for recommended accelerator settings and stricter matching of topology and configuration names.
  - Improved behavior in ARM64 and disconnected environments using Hugging Face storage.
  - Added clearer handling and diagnostics when no matching inference configuration is found.
  - Increased the default wait period for model-serving operations to better accommodate longer-running deployments.


threcc:
/hold

this should be merged only when the upstream PR: https://github.com/opendatahub-io/kserve/pull/1869  lands on rhoai 3.6ea2 downstream

tests/model_serving/model_server/llmd/llmd_configs/config_base.py
```diff
@@ -165,27 +164,51 @@ def readiness_probe(cls) -> dict:
...
+        2. Skip HuggingFace-backed tests on disconnected clusters.
+        3. Find the versioned CPU ``LLMInferenceServiceConfig`` for the
+           single-node workload topology.
+        4. Fail if no matching configuration is available.
+        5. Return a derived configuration class referencing the selected
+           ``LLMInferenceServiceConfig`` through ``base_refs``.
         """
-        Skip CPU inference on arm64 clusters.
+        supported_topology = "workload-single-node"
+        name_regex = r"^(?:.*-)?kserve-config-llm-template-cpu$"
```
pierDipi: if in `find_matching_llminferenceserviceconfig` we would change the meaning of `accelerator = None` to be "find the config with [empty `opendatahub.io/recommended-accelerators`](https://github.com/opendatahub-io/kserve/blob/master/config/overlays/odh/accelerators/cpu-config-llm-template.yaml#L10)"  we wouldn't need to rely on hardcoded config names

threcc: Done it, now the preset is selected based on 
- accelerator (None)  
- `topology` ("workload-single-node")
- ` accelerator_config_name_regex` not fast by default as we do for GPU, so that now can be overridden by fast variant

mwaykole (approved):

PR #2406, opened 2026-09-18
fix: resolve TrustyAI migration route directly

sheltoncyril:
## What
- resolve the TrustyAI service Route directly from the cluster
- avoid relying on unsupported TrustyAIService.external_route

## Validation
- `uv run pytest -q tests/ai_safety/trustyai_service/service/test_trustyai_service.py::test_trustyai_service_db_migration`
- Result: 1 passed

The image-registry validation behavior is intentionally unchanged.


sheltoncyril:
Emergency fix for trusyai tests breaking

